PR #313, opened 2022-03-23
feat(varnish receiver): Varnish add cache name

JonathanWamsley:
### Proposed Change
The config has been updated from working_dir to cache_dir. working_dir was confusing to work with. cache directory  can be left empty. When that happens the cache_name will be the default hostname. If multiple varnish instances are running, the specified cache_dir path can be used to select the varnish instance.

The scraper now adds a resource metric cache_name that comes from the cache_dir

##### Checklist
- [x] Changes are tested
- [x] CI has passed


receiver/varnishreceiver/config.go
```diff
@@ -27,15 +28,21 @@ import (
...
 
 // Validate validates the config.
 func (c *Config) Validate() error {
-	if c.WorkingDir != "" {
-		if _, err := os.Stat(c.WorkingDir); err != nil {
-			return fmt.Errorf(`"working_dir" does not exists: %w`, err)
+	if c.InstanceName != "" {
+		if _, err := os.Stat(c.InstanceName); err != nil {
+			return fmt.Errorf(`"instance_name" does not exists: %w`, err)
+		} else {
```
cpheps, line 40: I don't think this needs to be in an else statement as the if statement returns.

receiver/varnishreceiver/config.go
```diff
@@ -27,15 +28,21 @@ import (
...
+			lastIndex := strings.LastIndex(c.InstanceName, "/")
+			if lastIndex+1 == len(c.InstanceName) {
+				return fmt.Errorf(`"instance_name" has invalid trailing "/" %s`, c.InstanceName)
```
cpheps, line 43: Curious do you think we should just fix this or should it be an error? Also wondering if we can use the pathing stuff in stdlib to just give us a valid path.

receiver/varnishreceiver/scraper.go
```diff
@@ -59,6 +59,8 @@ func (v *varnishScraper) scrape(context.Context) (pdata.Metrics, error) {
 	now := pdata.NewTimestampFromTime(time.Now())
 	md := v.mb.NewMetricData()
 
+	md.ResourceMetrics().At(0).Resource().Attributes().UpsertString(metadata.A.CacheName, v.config.InstanceName)
```
cpheps, line 62: This is a direct index to index 0. Is that safe?

receiver/varnishreceiver/config.go
```diff
@@ -46,3 +53,16 @@ func (c *Config) Validate() error {
 
 	return nil
 }
+
+// SetDefaultHostname sets a default hostname if config InstanceName is empty.
+func (c *Config) SetDefaultHostname() error {
+	if c.InstanceName == "" {
+		hostname, err := os.Hostname()
```
cpheps, line 60: Does this happen before or after config validation and get passed into the `-n` argument? If it does is this always valid?

receiver/varnishreceiver/config.go
JonathanWamsley: fixed but now outdated

receiver/varnishreceiver/scraper.go
JonathanWamsley: The NewMetricData initializes the resource metric. If I use `md.ResourceMetrics().AppendEmpty().Resource().Attributes().UpsertString(metadata.A.CacheName, v.cacheName)` then I will have 2 resource metrics.

receiver/varnishreceiver/config.go
JonathanWamsley: this was in the wrong place. I need to preserve the was InstanceName, now CacheDir and so I moved this to the scraper since it is relevant to the scraper.

receiver/varnishreceiver/config.go
JonathanWamsley: outdated. checking for trailing / is not needed now because os.stat will send an error catching that

receiver/varnishreceiver/scraper.go
```diff
@@ -45,11 +43,36 @@ func newVarnishScraper(settings component.TelemetrySettings, config *Config) *va
 	}
 }
 
+func (v *varnishScraper) start(_ context.Context, host component.Host) error {
+	v.client = newVarnishClient(v.config, host, v.settings)
+	err := v.setCacheName()
```
cpheps, line 48: I think you solved the linter problem not quite the way it wanted. It was saying you can just `return v.setCacheName()`

receiver/varnishreceiver/scraper.go
```diff
@@ -45,11 +43,36 @@ func newVarnishScraper(settings component.TelemetrySettings, config *Config) *va
...
+	if v.config.CacheDir == "" {
+		hostname, err := os.Hostname()
+		if err != nil {
+			return err
+		}
+		v.cacheName = hostname
+		return nil
+	}
+	// cache dir is a valid named file dir and does not end in /.
+	lastIndex := strings.LastIndex(v.config.CacheDir, "/")
```
cpheps, line 66: I thought you said we don't need to do this checking any more as os.Stat would catch it.

receiver/varnishreceiver/scraper.go
JonathanWamsley: I see. fixed

cpheps (approved):

receiver/varnishreceiver/scraper.go
```diff
@@ -45,11 +43,32 @@ func newVarnishScraper(settings component.TelemetrySettings, config *Config) *va
...
+	v.cacheName = v.config.CacheDir[lastIndex+1:]
```
BinaryFissionGames, line 63: I see this has been discussed a little, but this comment here implies that paths like "/path/host-name/" are invalid (v.cacheName becomes ""), but we don't catch this in during config validation. 

Could we maybe use https://pkg.go.dev/path/filepath#Base instead to extract the cacheName, and not worry about the case where the path ends in `/`?

JonathanWamsley: filepath.Base() looks perfect! updated

BinaryFissionGames (approved):

PR #315, opened 2022-03-25
feat: Installation Packages Demo Feedback

BinaryFissionGames:
### Proposed Change
Some changes relating to internal feedback of current installations:

* Add LICENSE file to both Windows and Linux installations
* Add LICENSE to MSI installer
  * This shows a new prompt where the user must accept the license before installing
* Change Windows default install folder name to be "observIQ OpenTelemetry Collector"
  * This also changes the product display name, which I think is a positive change
  * Also changed the docs/screenshots to reflect this
* Add a link to our docs on configuring the collector to Linux installation, in addition to the OpenTelemetry docs.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


scripts/package/postinstall.sh
```diff
@@ -68,7 +68,10 @@ The "observiq-otel-collector" service has been configured!
 The collector's config file can be found here: 
   /opt/observiq-otel-collector/config.yaml
 
-For more information on configuring the collector, see the OpenTelemetry docs:
+For more information on configuring the collector, see the docs:
+  https://github.com/observIQ/observiq-otel-collector/blob/main/docs/installation-linux.md#configuring-the-collector
```
schmikei, line 72: Hmm I wonder if there is a way to catch whether we update the title instructions at all 🤔 either way should be good at the moment.

schmikei (commented):
Looks good for the most part, can't think of an easy way to validate but I assume everything has been tested.

scripts/package/postinstall.sh
BinaryFissionGames: Yeah, that's a good point. But even if the anchor ends up not existing, it'll still take them to that installation-linux docs, which is probably enough.

schmikei (approved):

windows/LICENSE.rtf
```diff
@@ -0,0 +1,206 @@
+{\rtf1\ansi\ansicpg1252\deff0\nouicompat\deflang1033{\fonttbl{\f0\fnil\fcharset0 Tahoma;}}
```
cpheps, line 1: What's the purpose of a rich text version of the license if we copy the normal version during the release flow? I'm a bit uneasy about having multiple copies of the license in the repo. If we ever change the license we'll need to know this is here.

cpheps (approved):

windows/LICENSE.rtf
BinaryFissionGames: We have a screen like this:
Screen Shot 2022-03-27 at 8 45 04 PM

If we don't format it, it looks really bad and janky. So we basically have 2 options: 

1. remove this license agreement step and just have the agreement be implicit by using our software. This is mostly what we were already doing before, and basically what we're doing on linux.
2. Add an rtf formatted license for this screen that makes the screen look correct. This is the option in this PR right now.

I figured no. 2 was preferred, since we're bundling the license and it makes sense to have the customer agree to the license before installing. It seems like the "right" way to handle it.

cpheps: ah gotcha. Ok I'm fine with that.

PR #316, opened 2022-03-27
feat: Add plugin receiver

jmwilliams89:
### Proposed Change
- Add a new plugin receiver for running templated pipelines

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (commented):
This looks pretty sweet, I only have a few nitpicks

receiver/pluginreceiver/README.md
```diff
@@ -0,0 +1,83 @@
...
+  {{if .enable_memory}}
+        memory:
+  {{end}}
+  service:
+    pipelines:
+      metrics:
+        receivers: [hostmetrics]
+```
+### Metadata
+Metadata fields are used to catalog and distinguish plugins. The following fields are used for metadata:
```
cpheps, line 57: Are all these fields required?

receiver/pluginreceiver/README.md
```diff
@@ -0,0 +1,83 @@
...
+- `description`
+- `version`
+
+### Parameters
+Parameters are the fields used to configure a plugin. The values of these fields are used when rendering the plugin's template, resulting in a dynamic pipeline. 
+
+The following keys are used when defining a parameter.
+| Key | Description |
+| --- | --- |
+| `name`      | The name of the parameter. This is the key used when configuring the parameter within the receiver. |
```
cpheps, line 68: Are all of these parameters required? If not are there default values?

receiver/pluginreceiver/plugin.go
```diff
@@ -0,0 +1,211 @@
...
+	if err := p.checkSupported(values); err != nil {
+		return fmt.Errorf("supported value failure: %w", err)
+	}
+
+	return nil
+}
+
+// checkDefined checks if any of the supplied values are defined by the plugin
+func (p *Plugin) checkDefined(values map[string]interface{}) error {
+	parameterMap := make(map[string]Parameter)
```
cpheps, line 112: Looks like we're only using this map as a quick lookup the value should probably be a `struct{}{}` to limit the memory impact since we don't care about the value associated with the key.

receiver/pluginreceiver/plugin.go
```diff
@@ -0,0 +1,211 @@
...
+
+// checkType checks if the values match their parameter type
+func (p *Plugin) checkType(values map[string]interface{}) error {
+	for _, parameter := range p.Parameters {
+		value, ok := values[parameter.Name]
+		if !ok {
+			continue
+		}
+
+		switch parameter.Type {
```
cpheps, line 146: Will go limit you to only specifying parameter types that were defined in your const list? I don't remember but since it's just a string alias I feel like you can probably put any valid string as a parameter.Type. If that's the case we should probably have a default clause here.

receiver/pluginreceiver/receiver.go
```diff
@@ -0,0 +1,91 @@
...
+
+	for {
+		select {
+		case err := <-errChan:
+			return err
+		case <-ticker.C:
+			if svc.GetState() == service.Running {
+				return nil
+			}
+		}
```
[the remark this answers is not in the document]
cpheps: We definitely should. As we've seen we can't trust the underlying collector structures to respect `ctx.Done`

receiver/pluginreceiver/factory.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, fmt.Errorf("invalid plugin parameter: %w", err)
+	}
+
+	components, err := plugin.RenderComponents(receiverConfig.Parameters)
+	if err != nil {
+		return nil, fmt.Errorf("failed to render plugin components: %w", err)
+	}
+
+	configProvider := createConfigProvider(components)
+	logger := set.Logger.Named(receiverConfig.ID().String())
```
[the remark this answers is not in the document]
jmwilliams89: Perfect. That should eliminate some unnecessary code.

receiver/pluginreceiver/provider.go
```diff
@@ -0,0 +1,227 @@
...
+
+		factories[extensionType] = extensionFactory
+	}
+
+	return factories, nil
+}
+
+// parseComponentType parses a component type from a string
+func parseComponentType(value string) config.Type {
+	typeValue := strings.Split(value, "/")[0]
```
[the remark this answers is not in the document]
jmwilliams89: No. I think we should make that change.

receiver/pluginreceiver/receiver.go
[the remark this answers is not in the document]
jmwilliams89: Agreed. Good catch.

receiver/pluginreceiver/README.md
jmwilliams89: They are not required at the moment. Do we want to enforce this with a validation step after we unmarshal to the plugin struct?

cpheps: I don't think we need to but I think it's worth calling out if they are required or not. I know if I'm setting one of these up to test it would be nice to know the minimum set of things I need in order to get going.

receiver/pluginreceiver/README.md
jmwilliams89: No. They are not required at the moment. The defaults would be the zero values of those fields when unmarshalling into the struct. As stated in the comment above, we could add additional validation if we think it's necessary.

cpheps: I think that's fine maybe just adding a `required` column to the table just so a user knows the minimum set.

receiver/pluginreceiver/README.md
jmwilliams89: Okay. I think that's a fair compromise. I'll update the documentation.

cpheps (approved):

PR #321, opened 2022-03-30
Example configuration proposal

jsirianni:
### Proposed Change

This PR represents a proposal for documenting usecases.

Added a directory structure for example configurations:
```bash
$ tree config
config
├── example.yaml
└── google_cloud_exporter
    ├── hostmetrics
    │   ├── config.yaml
    │   └── README.md
    ├── prerequisites.md
    └── prometheus
        ├── config.yaml
        └── README.md
```

Each exporter will have some sort of setup that is applicable for all examples, this is where `google_cloud_exporter/prerequisites.md` comes in. Each exporter directory will have sub directories for each usercase. This PR contains `hostmetrics` and `prometheus` example configurations.

Initially, we were not going to include a doc for each usecase, however, it may be too difficult to consume comments in the config.yaml. For example:
- hostmetrics process metrics require root
- postgres (any many  others) will require least privileged user (LPU) instructions
- mysql will require special configuration to enable query logging

An alternative would be to setup a dedicated repository for these configurations, and link to it from this repo.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


config/google_cloud_exporter/hostmetrics/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Prerequisites
+
+See the [prerequisites](../prerequisites.md) doc for Google Cloud prerequisites.
+
+## Configuration
+
+An example configuration is located [here](./config.yaml).
+
+1. Copy [config.yaml](./config.yaml) to `/opt/observiq-otel-collector/config.yaml`
```
cpheps, line 13: These instructions are all linux specific I think we either need to be generic by saying `copy to your config location` or have a line for each OS's default location.

cpheps (commented):
I did a glance review but this seems fine. Something we'll have to think through is should we put the base config in a directory or should we create an `examples` dir that configs that don't ship with the collector but are still valuable go into?

Also, if we want to ship with these wondering if we need something to call out what ones we want to ship with or do we just hardcode it into goreleaser/msi creation?

config/google_cloud_exporter/hostmetrics/README.md
jsirianni: My intention is to **not** ship these configs. Perhaps `docs/examples` is a more appropriate location. 

The current examples might work in any environment, but future examples may not. I do not want to give off the impression that they are "plug and play". Additionally, I can see the number of examples growing to 20-30, maybe 50.

config/google_cloud_exporter/prerequisites.md
```diff
@@ -0,0 +1,72 @@
...
+
+**Windows**
+
+In this example, the key is placed at `C:/observiq/collector/sa.json`.
+
+Set the `GOOGLE_APPLICATION_CREDENTIALS` with the command prompt `setx` command.
+
+Run the following command
+```
+setx GOOGLE_APPLICATION_CREDENTIALS "C:/observiq/collector/sa.json"
```
BinaryFissionGames, line 68: I'm curious, does this work without specifying `/m`? I haven't tried this way.

config/google_cloud_exporter/prerequisites.md
```diff
@@ -0,0 +1,72 @@
+# Google Cloud Exporter Prerequisites
```
BinaryFissionGames, line 1: Could this be rolled into a README.md, with a short desc of what ties these configs together (e.g. "This directory's subdirectories contain configs that export to Google Cloud Platform"). I think it'd be nicer for people looking through github, since it shows README files when you are looking at a particular directory.

config/google_cloud_exporter/prerequisites.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+**Windows**
+
+In this example, the key is placed at `C:/observiq/collector/sa.json`.
+
+Set the `GOOGLE_APPLICATION_CREDENTIALS` with the command prompt `setx` command.
+
+Run the following command
+```
```
BinaryFissionGames, line 67:
```suggestion
```cmd
```

config/google_cloud_exporter/hostmetrics/README.md
```diff
@@ -0,0 +1,44 @@
...
+2. Restart the collector: `sudo systemctl restart observiq-otel-collector`
+
+## Setup
+
+This example assumes that [node exporter](https://github.com/prometheus/node_exporter) is running on the host system, on port 9100. Generally, all prometheus style exporters are supported, such as the [aerospike exporter](https://github.com/aerospike/aerospike-prometheus-exporter).
+
+If you wish to scrape metrics from external systems, update `config.yaml`'s prometheus scrape config with a list of IP addresses:
+```
```
BinaryFissionGames, line 21:
```suggestion
```yaml
```

config/google_cloud_exporter/hostmetrics/README.md
cpheps: I think it's fine to keep these here for now. I think this is a cross that bridge when we come to it. If someone asks about windows stuff we address it then.

config/google_cloud_exporter/prerequisites.md
cpheps: I plan on linking to `Example` configs or explaining it more in the base readme once these are in.

cpheps: Honestly I think just rename this file to `README.md` and have this be a section. Allowing us to add more sections if need be

jsirianni: Good points, will handle it.

BinaryFissionGames (approved):

PR #324, opened 2022-03-30
feat(install): Cleaned up Linux Install Script

cpheps:
### Proposed Change
Cleaned up the linux install script to better match our standards

Here is a sample of install:
install_collector

Here is a sample of uninstall:
uninstall_collector

##### Checklist
- [X] Changes are tested
- [ ] CI has passed


scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+    url=$DOWNLOAD_BASE
+  fi
+
+  if [ -z "$arch" ] ; then 
+    os_arch=$arch
+  fi
+
+  if [ -z "$pkg_type" ] ; then
+    package_type=$pkg_type
+  fi
```
BinaryFissionGames, line 375: These don't make sense to me; If "$arch" is empty, "$os_arch" would be set to empty? I think these two if blocks probably shouldn't exist?

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
-        arch="arm"
-        ;;
+      os_arch="arm"
+      ;;
     *)
-        echo "Installation failed: Unsupported os arch: $os_arch"
-        exit 1
-        ;;
-    esac
+      error "Unsupported os arch: $os_arch"
```
BinaryFissionGames, line 338: I think this should be `error_exit`

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+
+# Set the package type before install
+set_package_type()
+{
+  if command -v dpkg > /dev/null 2>&1; then
+      package_type="deb"
+  elif command -v rpm > /dev/null 2>&1; then
+      package_type="rpm"
+  else
+      error "Could not find dpkg or rpm on the system"
```
BinaryFissionGames, line 351: Same here, should probably be `error_exit`

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+      return 0
+  elif command -v rpm > /dev/null 2>&1; then
+      succeeded
+      pkg_type="rpm"
+      return 0
+  else
+      failed
+      error_exit "Could not find dpkg or rpm on the system"
+  fi
 }
```
BinaryFissionGames, line 483: This check function sets a variable (`pkg_type`), but that's `set_package_type`'s job (and in fact, they both use different variables, for some reason)

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+      proxy_args="$proxy_args -U $proxy_user:$proxy_password"
     fi
+  fi
+
+  if [ -n "$proxy" ]; then
+    info "Downloading package using proxy..."
+  fi 
+
+  info "Downloading package..."
+  eval curl -L "$proxy_args" "$collector_download_url" -o "$out_file_path" --progress-bar --fail || error_exit "$LINENO" "Failed to download package"
```
BinaryFissionGames, line 505: I think this is correct, but the eval is confusing me so I just want to verify; "$proxy_args" DOES split here, right? Because of the eval?

Follow up question, would that cause "out_file_path" to split if it had spaces in it?

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+    rpm)
+      rpm -U "$out_file_path" > /dev/null
+      ;;
+    *)
+      error "Unrecognized package type"
+      return 1
+      ;;
+  esac
+  return 0
+}
```
BinaryFissionGames, line 535: This is a misnomer. This installs the package, not extract the package.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+        -l|--url)
+          url=$2 ; shift 2 ;;
+        -x|--proxy)
+          proxy=$2 ; shift 2 ;;
+        -U|--proxy-user)
+          proxy_user=$2 ; shift 2 ;;
+        -P|--proxy-password)
+          proxy_password=$2 ; shift 2 ;;
+        -r|--uninstall)
+          uninstall # TODO
```
BinaryFissionGames, line 625: This is implemented, and not a TODO from what I can tell?

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+    rpm)
+      rpm -e "observiq-otel-collector" > /dev/null 2>&1
+      ;;
+    *)
+      error "Unrecognized package type"
+      return 1
+      ;;
+  esac
+  return 0
+}
```
BinaryFissionGames, line 579: We don't check if anything is actually present to uninstall. Which might error out at this step.

ALSO, if nothing is installed, and we do not error out here, the user has no feedback that there was nothing to uninstall. Which might be fine, but I recall this was feedback that was given for the first version of the uninstall script, so I thought I might add it here.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+    info "Downloading package using proxy..."
+  fi 
+
+  info "Downloading package..."
+  eval curl -L "$proxy_args" "$collector_download_url" -o "$out_file_path" --progress-bar --fail || error_exit "$LINENO" "Failed to download package"
+  succeeded
+
+  info "Extracting package..."
+  extract_package || error_exit "$LINENO" "Failed to extract package"
+  succeeded
```
BinaryFissionGames, line 510: Verbage should reflect that this is actually installing the package.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+      fi
     fi
+  done
+
+  if [ -n "$FAILED_PREREQS" ]; then
+    failed
+    error_exit "The following dependencies are required by this script: [$FAILED_PREREQS]"
+  fi
+  succeeded
+  return 0
```
BinaryFissionGames, line 464: Why return 0 here? We don't ever check this, and we never return anything but 0.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+# This
+root_check()
+{
+  system_user_name=$(id -un)
+  if [ "${system_user_name}" != 'root' ]
+  then
+    failed
+    error_exit "Script needs to be run as root or with sudo permissions"
+  fi
+}
```
BinaryFissionGames, line 407: Comment here was left unfinished;

I'm wondering if this is the best way to handle checking for root. There seem to be pitfalls here that might not work everywhere. https://stackoverflow.com/questions/18215973/how-to-check-if-running-as-root-in-a-bash-script

Is it better to check if `id -u` is 0? I know this is super edgecase, but it's something I worry about.

Also, the error message indicates it should be run "with sudo permissions", but I would change this to be "with sudo", since you need more that just the permissions, the script does not invoke sudo for you.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+    package_file_name="${PACKAGE_NAME}_${arch}.${package_type}"
+    out_file_path="$TMP_DIR/$package_file_name"
+}
+
+set_proxy()
+{
+  if [ -n "$proxy" ]; then
+    info "Using proxy from arguments: $proxy"
+    if [ -n "$proxy_user" ]; then
+      while [ -z "$proxy_password" ] && [ ! "$accept_defaults" = "yes" ]; do
```
BinaryFissionGames, line 276: `accept_defaults` is not assigned anywhere, we should remove it from this if clause.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+      full_proxy="$protocol//$proxy_user:$proxy_password@$host"
+    fi
+
+  elif [ -n "$http_proxy" ]; then
+    info "Using proxy from profile: $http_proxy"
+    proxy="$http_proxy"
+  elif [ -n "$https_proxy" ]; then
+    info "Using proxy from profile: $https_proxy"
+    proxy="$https_proxy"
+  fi
```
BinaryFissionGames, line 299: Where do these variables come from? don't seem to be set anywhere, but are they supposed to be from the user's environment? If so, do we document that we use these?

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# shellcheck disable=SC2119
+# SC2119 -> Use foo "$@" if function's $1 should mean script's $1.
```
BinaryFissionGames, line 17: I dislike having this global disable; Is this violated all over, or can we disable it for just where this is relevant?

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+    set_file_name
+
+    success "Configuration complete!"
+    decrease_indent
+}
+
+set_file_name() {
+    package_file_name="${PACKAGE_NAME}_${arch}.${package_type}"
+    out_file_path="$TMP_DIR/$package_file_name"
+}
```
BinaryFissionGames, line 269: This variable doesn't seem to be used anywhere

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+{
+  info "Checking for package manager..."
+  if command -v dpkg > /dev/null 2>&1; then
+      succeeded
+      pkg_type="deb"
+      return 0
+  elif command -v rpm > /dev/null 2>&1; then
+      succeeded
+      pkg_type="rpm"
+      return 0
```
BinaryFissionGames, line 478: I'm noticing a lot of inconsistency with how we're handling errors.

In general, should the called functions be returning a status code (`return 0` on success, `return 1` on failure) and let the caller handle exiting and printing an error, or should the called function handle erroring. 

IMO, in general, we should just let the function handle erroring since it has finer granularity with what actually went wrong.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+}
+
+# This
+root_check()
+{
+  system_user_name=$(id -un)
+  if [ "${system_user_name}" != 'root' ]
+  then
+    failed
+    error_exit "Script needs to be run as root or with sudo permissions"
```
BinaryFissionGames, line 405: We've been inconsistently specifying "$LINENO" for error_exit. Is there a reason we only specify it occasionally, or should we be specifying it all the time?

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+          version=$2 ; shift 2 ;;
+        -l|--url)
+          url=$2 ; shift 2 ;;
+        -x|--proxy)
+          proxy=$2 ; shift 2 ;;
+        -U|--proxy-user)
+          proxy_user=$2 ; shift 2 ;;
+        -P|--proxy-password)
+          proxy_password=$2 ; shift 2 ;;
+        -r|--uninstall)
```
BinaryFissionGames, line 624: I notice we have `-l` for specifying the url, and `-r` for specifying uninstall. IMO, one of these should be `-u`

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+Usage:
+  $(fg_yellow '-v, --version')
+      Defines the version of the observIQ OpenTelemetry Collector.
+      If not provided, this will default to the latest version.
+      Example: '-v 1.2.12' will download 1.2.12.
+
+  $(fg_yellow '-l, --url')
+      Defines the URL that the components will be downloaded from.
+      If not provided, this will default to observIQ OpenTelemetry Collector\'s GitHub releases.
+      Example: '-l http://my.domain.org/observiq-otel-collector' will download from there.
```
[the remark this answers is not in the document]
cpheps: I originally removed this and the proxy but left it in as it's not unreasonable to think a customer may host a package they've done security checks on. We could always change out the binary at that URL so hosting something you've approved if security is your concern isn't out of the normal.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+    proxy="$https_proxy"
+  fi
+
+  if [ -z "$full_proxy" ]; then
+    full_proxy="$proxy"
+  fi
+}
+
+# This will set the os based on the current runtime environment.
+# Accepted values are darwin and linux. This value cannot be overriden.
```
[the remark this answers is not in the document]
cpheps: Valid I'll clean those up.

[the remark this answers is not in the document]
cpheps: I've seen the duplicate pattern in this script of `check` and `set`. I'm going to remove the darwin comment but let the set. Will it ever error? Almost certainly not but leaves us with a framework to add darwin or something crazier down the line.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+}
+
+# This will set the urls to use when downloading the collector and its plugins.
+# These urls are constructed based on the --version flag or COLLECTOR_VERSION env variable.
+# If not specified, the version defaults to "latest".
+set_download_urls()
+{
+  if [ -z "$version" ] ; then
+    # shellcheck disable=SC2153
+    version=$COLLECTOR_VERSION
```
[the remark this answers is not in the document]
cpheps: It is a bit hard to follow. I didn't change it since I copied it from a known working script that had this pattern the benefit to clean it up and refactor the entire script to use locals and pass them along didn't seem worth the effort. For better or worse this is not an anti-pattern in bash scripts.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+      package_type="deb"
+  elif command -v rpm > /dev/null 2>&1; then
+      package_type="rpm"
+  else
+      error "Could not find dpkg or rpm on the system"
+  fi
+}
+
+# This will set the urls to use when downloading the collector and its plugins.
+# These urls are constructed based on the --version flag or COLLECTOR_VERSION env variable.
```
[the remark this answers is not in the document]
cpheps: No we don't document the env. It's not uncommon to have envs and flags together. If I have automation running a script it's usually easier to just add an ENV to the automation than it is to modify the core command string for a one off case.

scripts/install/install_unix.sh
cpheps: I'm not sure to be honest. 

I don't know what happens to `out_file_path` if it has spaces. I think that's unlikely (not impossible). Most linux admins are not going to have a directory path with spaces in it. I can test it out though to see what happens.

scripts/install/install_unix.sh
cpheps: yeah I know. There was already a build in function from the previous script called `install_package`. Got any other name suggestions? `unpack_package`?

scripts/install/install_unix.sh
BinaryFissionGames: Yes it is, I'm just blind.

scripts/install/install_unix.sh
BinaryFissionGames: Ugh, I see what you mean. I think that other "install_package" function should be something like just "install". I mean, it's definitely doing more than just installing the package. Do we need to keep that other function name the same?

scripts/install/install_unix.sh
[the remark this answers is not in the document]
BinaryFissionGames: Fair enough with env vs command line.

I think we should document the env variables in the usage command. 

If having an env var that does the same thing as arguments is very common, then I guess we should probably do the same for proxy support and the package URL, which I think was a feature of the other scripts too.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,638 @@
...
+  root_check
+  succeeded
+
+  info "Stopping service..."
+  systemctl stop observiq-otel-collector > /dev/null || error_exit "$LINENO" "Failed to stop service"
+  succeeded
+
+  info "Disabling service..."
+  systemctl disable observiq-otel-collector > /dev/null 2>&1 || error_exit "$LINENO" "Failed to disable service"
+  succeeded
```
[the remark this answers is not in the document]
cpheps: The script enabled the service though, why would it not disable it?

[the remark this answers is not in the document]
BinaryFissionGames: Yep, you're right. I was thinking about deleting the service file. This makes sense.

scripts/install/install_unix.sh
cpheps: So yeah I made the choice to redirect stderr and stdout to dev null so this no-ops if nothing exists. I think this is probably a product question on the behavior. I'll ask.

scripts/install/install_unix.sh
cpheps: I think this was left over from the old script. It's fair to say why but it's also not a detriment to do so.

scripts/install/install_unix.sh
cpheps: I'll fixup the comment and error message.

I actually tried what that stackoverflow link suggested and it didn't work. I think this is fine it's an edge case.

scripts/install/install_unix.sh
cpheps: They were in the old script looks like they don't do anything. I'll remove.

scripts/install/install_unix.sh
cpheps: Added it to all

scripts/install/install_unix.sh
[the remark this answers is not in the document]
BinaryFissionGames: I'm fine with keeping it in, it just seems weird that they would go through the trouble of setting up a server to host their package, then want to use this install script over just directly installing the package.

scripts/install/install_unix.sh
jsirianni: I am not sure, my suspicion is that it has always been there, looking at some of our past agents / collectors. 

I am generally a fan of disabling shellcheck rules where they happen instead of globally.

I just ran shellcheck against the script with those two lines removed and it passes fine (it has two unrelated issues)
```
In scripts/install/install_unix.sh line 260:
      while [ -z "$proxy_password" ] && [ ! "$accept_defaults" = "yes" ]; do
                                             ^--------------^ SC2154: accept_defaults is referenced but not assigned.


In scripts/install/install_unix.sh line 297:
      os="linux"
      ^-- SC2034: os appears unused. Verify use (or export if used externally).
```

jsirianni: It might be a good idea to have shellcheck run as part of our workflow. The check is super fast and won't have a big impact on CI.

scripts/install/install_unix.sh
BinaryFissionGames: Ok. I feel like it's confusing to new customers, is all. The only reason, from what I can see, that we made these flags weird, is because there was a `--service-user` option in previous scripts.

scripts/install/install_unix.sh
```diff
@@ -13,165 +13,620 @@
...
+    set_download_urls
+    set_proxy
+    set_file_name
+
+    success "Configuration complete!"
+    decrease_indent
+}
+
+set_file_name() {
+    local package_file_name="${PACKAGE_NAME}_${arch}.${package_type}"
```
BinaryFissionGames, line 269: apparently the `local` keyword isn't POSIX shell; I think we're fine to remove that, my concern was more that the variable wasn't being used anywhere, when it actually was.

cpheps: roger, I'll remove it.

scripts/install/install_unix.sh
cpheps: I think if we have more shell scripts I would do so but we have only a few that likely won't change much I don't think it's worth the extra CI and tooling right now.

BinaryFissionGames: It sounds like we're OK with these two lines then. We don't check against it in CI, but I know a lot of us use the shell-check extension in vscode to catch mistakes.

scripts/install/install_unix.sh
BinaryFissionGames: The comment is still left empty here

cpheps: Good catch. I think I got distracted by the other points.

scripts/install/install_unix.sh
cpheps: Yeah I think that's fine for now.

jsirianni (approved):

scripts/install/install_unix.sh
cpheps: So it doesn't work with spaces in the path. Our current stanza and BP scripts will have this same issue since I copied this line. I'm ok accepting this risk as spaces in directory names is not common in linux.

@jsirianni @BinaryFissionGames are you also ok with this restriction?

scripts/install/install_unix.sh
cpheps: Spoke with Joe H. he's ok with this being a no-op. Resolving this for now.

BinaryFissionGames (approved):

PR #328, opened 2022-04-04
feat(config): Modified default scrapers for hostmetric receiver in the sample config

cpheps:
### Proposed Change
Added the following scrapers to the host metric receiver in the sample config:

- filesystem
- memory
- network
- cpu
- disk

Added notes in config that `cpu` and `disk` are not compatible on MacOS. See table in [Getting Started](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver#getting-started) of hostmetric receiver.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/config/example.yaml
+++ b/config/example.yaml
@@ -13,7 +13,7 @@ receivers:
       filesystem:
       memory:
       network:
-      # These scrapers do not support MacOS please comment them out to reduce errors in collector logs
+      # Remove these scrapers for macOS collectors as they are not supported and will generate errors in the logs
       cpu:
       disk:
 
--- a/config/example.yaml
+++ b/config/example.yaml
@@ -8,7 +8,14 @@ receivers:
   hostmetrics:
     collection_interval: 1m
     scrapers:
+      # Theses scrapers work on all operating systems
       load:
+      filesystem:
+      memory:
+      network:
+      # These scrapers do not support MacOS please comment them out to reduce errors in collector logs
+      cpu:
+      disk:
 
   # The syslog receiver; this configuration will listen on every network interface on port 514
   # for UDP syslog messages using the rfc3164 protocol.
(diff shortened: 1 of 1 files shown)


jsirianni (commented):
Process metrics are very popular for us, but require `root` / `Administrator`, it might be worth adding process (commented out, with an explanation)?

config/example.yaml
```diff
@@ -8,7 +8,14 @@ receivers:
   hostmetrics:
     collection_interval: 1m
     scrapers:
+      # Theses scrapers work on all operating systems
       load:
+      filesystem:
+      memory:
+      network:
+      # These scrapers do not support MacOS please comment them out to reduce errors in collector logs
```
dehaansa, line 16: Rather than "please comment" maybe "Remove these scrapers for MacOS collectors as they are not supported and will generate errors in the logs"

cpheps: I like that.

antonblock: Also I believe `macOS` is the official spelling

cpheps:
> Process metrics are very popular for us, but require root / Administrator, it might be worth adding process (commented out, with an explanation)?

@jsirianni We chose not to add those in as they don't work for everyone out of the box and we wanted a config that always works on install to validate the collector.

I'm unsure if I want to add these in commented out. We do link to the receiver docs at the top. Were you thinking of a more comprehensive explanation or more something simple like below?

```yaml
# Uncomment the below section to get process level metrics. You'll need to restart the collector with root/Administrator permissions
process
```

If you have something better would you mind sharing and I'll add it in.

jsirianni:
> > Process metrics are very popular for us, but require root / Administrator, it might be worth adding process (commented out, with an explanation)?
> 
> @jsirianni We chose not to add those in as they don't work for everyone out of the box and we wanted a config that always works on install to validate the collector.
> 
> I'm unsure if I want to add these in commented out. We do link to the receiver docs at the top. Were you thinking of a more comprehensive explanation or more something simple like below?
> 
> ```yaml
> # Uncomment the below section to get process level metrics. You'll need to restart the collector with root/Administrator permissions
> #process:
> ```
> 
> If you have something better would you mind sharing and I'll add it in.

That makes sense. There is some extra configuration that would be required, and I suppose that is already covered here: https://github.com/observIQ/observiq-otel-collector/tree/main/config/google_cloud_exporter/hostmetrics

jsirianni (approved):

antonblock (approved):

PR #329, opened 2022-04-04
feat(installer): Removed version from MSI name

cpheps:
### Proposed Change
Removed the version from the `.msi` name so we can provide a repeatable one-liner install command that can download directly from GitHub.

Updated documentation to highlight the one-liner command.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


.github/workflows/release.yml
```diff
@@ -171,14 +168,11 @@ jobs:
...
-      - name: "Get Tag From Environment"
-        id: get-tag
-        run: printf '::set-output name=tag::%s' "$(printf '%s' "${{ github.ref }}" | sed 's/refs\/tags\///')"
       - name: Retrieve Windows MSI Installer
         uses: actions/download-artifact@v2
         with:
-          name: observiq-otel-collector-${{ steps.get-tag.outputs.tag }}.msi
-          path: observiq-otel-collector-${{ steps.get-tag.outputs.tag }}.msi
+          name: observiq-otel-collector.msi
+          path: observiq-otel-collector.msi
```
BinaryFissionGames, line 175: Forgot to mention this, but we'll also need to change the goreleaser yaml so that the glob includes this file.

cpheps: I double check I thought it was `observiq-otel-collector*.msi` so it would be covered.

cpheps: Ahh no it has a dash. Good catch!

BinaryFissionGames (approved):

schmikei (approved):

PR #336, opened 2022-04-05
Add redis example config

jsirianni:
### Proposed Change

Added redis example configuration for Google Cloud.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/config/google_cloud_exporter/redis/README.md
+++ b/config/google_cloud_exporter/redis/README.md
@@ -1,0 +1,11 @@
+# Redis Metrics with Google Cloud
+
+The Redis Receiver can be used to send Redis metrics to Google Cloud Monitoring. See the [documentation](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/redisreceiver) for more details.
+
+## Limitations
+
+The collector must be installed on the Redis system.
+
+## Prerequisites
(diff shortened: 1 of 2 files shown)


config/google_cloud_exporter/redis/README.md
```diff
@@ -0,0 +1,11 @@
+# Redis Metrics with Google Cloud
+
+The Redis Receiver can be used to send Redis metrics to Google Cloud Monitoring. See the [documentation](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/redisreceiver) for more details.
+
+## Limitations
+
+The collector must be installed on the Redis system.
```
cpheps, line 7: Is this true? Looks like the receiver takes an endpoint so it should work remotely. I could be wrong so please let me know if there's some limitation that makes this not possible.

jsirianni: For now, collecting against localhost is the safe option because Redis (and some other receivers) do not have resource attributes that can be used to uniquely identify where the metrics came from. This config uses the hostname of the agent system for unique identification, which will not work if a single agent is collecting against multiple remote Redis systems.

cpheps (approved):

PR #345, opened 2022-04-11
Add storage directory to release packages

jsirianni:
### Proposed Change

Stateful receivers will likely make use of the storage extension. For example, reading log files (saving state and buffering).

**Linux**

- Added directory `/opt/observiq-otel-collector/storage`
- Added environment `OIQ_OTEL_COLLECTOR_STORAGE` pointing to the directory, allowing configs to read `$OIQ_OTEL_COLLECTOR_STORAGE` instead of hard coding the value.

**Windows**

- Added storage directory to install directory. The `.include` file is required for WIX to copy the directory, afaik. 

Windows test output
```
  File C:/Program Files/observIQ OpenTelemetry Collector/plugins
     [PASS]  should exist
     [PASS]  should be directory
  File C:/Program Files/observIQ OpenTelemetry Collector/storage
     [PASS]  should exist
     [PASS]  should be directory
  File C:/Program Files/observIQ OpenTelemetry Collector/observiq-otel-collector.exe
     [PASS]  should exist
     [PASS]  should be file
  File C:/Program Files/observIQ OpenTelemetry Collector/config.yaml
     [PASS]  should exist
     [PASS]  should be file
  File C:/Program Files/observIQ OpenTelemetry Collector/plugins/aerospike.yaml
     [PASS]  should exist
     [PASS]  should be file
  File C:/Program Files/observIQ OpenTelemetry Collector/plugins/microsoft_iis.yaml
     [PASS]  should exist
     [PASS]  should be file
  File C:/Program Files/observIQ OpenTelemetry Collector/plugins/zookeeper.yaml
     [PASS]  should exist
     [PASS]  should be file
  Service observiq-otel-collector
     [PASS]  should be installed
     [PASS]  should be enabled
     [PASS]  should be running

Test Summary: 17 successful, 0 failures, 0 skipped
```

**Other**

Also updated make release-test to build with `--snapshot`, allowing the package to be installed over the previous tagged release. Useful when testing locally.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -47,6 +47,9 @@ archives:
       - src: release_deps/plugins/*
         dst: plugins
         strip_parent: true
+      - src: release_deps/storage/*
+        dst: storage
+        strip_parent: true
 
     format_overrides:
       - goos: windows
@@ -206,6 +209,7 @@ brews:
       bin.install "observiq-otel-collector"
       prefix.install "LICENSE", "config.yaml"
       prefix.install Dir["plugins"]
+      prefix.install Dir["storage"]
       lib.install "opentelemetry-java-contrib-jmx-metrics.jar"
     plist: |
       <?xml version="1.0" encoding="UTF-8"?>
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -206,7 +206,6 @@ brews:
       bin.install "observiq-otel-collector"
       prefix.install "LICENSE", "config.yaml"
       prefix.install Dir["plugins"]
-      FileUtils.mkdir_p(File.expand_path("~/observiq-otel-collector/storage")) unless File.exists?(File.expand_path("~/observiq-otel-collector/storage"))
       lib.install "opentelemetry-java-contrib-jmx-metrics.jar"
     plist: |
       <?xml version="1.0" encoding="UTF-8"?>
@@ -236,5 +235,3 @@ brews:
       </plist>
     # For local testing
     # skip_upload: "true"
-
-
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -110,7 +110,7 @@ nfpms:
       - dst: /opt/observiq-otel-collector/storage
         type: dir
         file_info:
-          mode: 0755
+          mode: 0750
           owner: observiq-otel-collector
           group: observiq-otel-collector
     scripts:
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -47,9 +47,6 @@ archives:
       - src: release_deps/plugins/*
         dst: plugins
         strip_parent: true
-      - src: release_deps/storage/*
-        dst: storage
-        strip_parent: true
 
     format_overrides:
       - goos: windows
@@ -209,7 +206,7 @@ brews:
       bin.install "observiq-otel-collector"
       prefix.install "LICENSE", "config.yaml"
       prefix.install Dir["plugins"]
-      prefix.install Dir["storage"]
+      FileUtils.mkdir_p(File.expand_path("~/observiq-otel-collector/storage")) unless File.exists?(File.expand_path("~/observiq-otel-collector/storage"))
       lib.install "opentelemetry-java-contrib-jmx-metrics.jar"
     plist: |
       <?xml version="1.0" encoding="UTF-8"?>
@@ -239,3 +236,5 @@ brews:
       </plist>
     # For local testing
     # skip_upload: "true"
+
+
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -100,6 +100,14 @@ nfpms:
       # nfpm failed: cannot write header of release_deps/plugins/amazon_eks.yaml to data.tar.gz: archive/tar: missed writing 1736 bytes
       - src: release_deps/plugins/*
         dst: /opt/observiq-otel-collector/plugins
+      # Storage dir is used by stateful receivers, such as filelog receiver. It allows
+      # receivers to track their progress and buffer data.
+      - dst: /opt/observiq-otel-collector/storage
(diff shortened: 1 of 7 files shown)


cpheps (commented):
Does this need to be added to homebrew also?

scripts/package/postinstall.sh
```diff
@@ -43,6 +43,7 @@ Type=simple
 User=observiq-otel-collector
 Group=observiq-otel-collector
 Environment=PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
+Environment=OIQ_OTEL_COLLECTOR_STORAGE=/opt/observiq-otel-collector/storage
```
BinaryFissionGames, line 46: Can we also add this env var to Windows?

.goreleaser.yml
```diff
@@ -102,6 +102,14 @@ nfpms:
       # nfpm failed: cannot write header of release_deps/plugins/amazon_eks.yaml to data.tar.gz: archive/tar: missed writing 1736 bytes
       - src: release_deps/plugins/*
         dst: /opt/observiq-otel-collector/plugins
+      # Storage dir is used by stateful receivers, such as filelog receiver. It allows
+      # receivers to track their progress and buffer data.
+      - dst: /opt/observiq-otel-collector/storage
+        type: dir
+        file_info:
+          mode: 0755
```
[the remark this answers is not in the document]
jsirianni: Yeah, not a bad idea. The collector **should** be writing files with 0640 but it doesn't hurt. Done.

jsirianni:
> Does this need to be added to homebrew also?

I pushed a commit, not sure how I feel about it. We need to bundle `storage/.include` with all archives to pull this off.

scripts/package/postinstall.sh
jsirianni: Would be nice, but I am not sure what that change would look like for wix.

.goreleaser.yml
```diff
@@ -198,6 +209,7 @@ brews:
       bin.install "observiq-otel-collector"
       prefix.install "LICENSE", "config.yaml"
       prefix.install Dir["plugins"]
+      prefix.install Dir["storage"]
```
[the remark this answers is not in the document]
jsirianni: I've reverted the homebrew change in favor of creating ~/observiq-otel-collector/storage. It is unclear to me if `FileUtils` and `File` are available (imported).

.goreleaser.yml
```diff
@@ -209,7 +206,7 @@ brews:
       bin.install "observiq-otel-collector"
       prefix.install "LICENSE", "config.yaml"
       prefix.install Dir["plugins"]
-      prefix.install Dir["storage"]
+      FileUtils.mkdir_p(File.expand_path("~/observiq-otel-collector/storage")) unless File.exists?(File.expand_path("~/observiq-otel-collector/storage"))
```
cpheps, line 209: `~/observiq-otel-collector` is not the collector path. Did you test it and it actually worked? I thought the file ops were limited to within the homebrew directory. Does the storage directory have to exist in relation to the collector if so it should be `#{bin.install}/storage`

.goreleaser.yml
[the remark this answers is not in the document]
cpheps: You can test by pointing this at your formula by running `brew install -build-from-source dist/observiq-otel-collector.rb`

scripts/package/postinstall.sh
jsirianni: Done

windows/wix.json
```diff
@@ -32,6 +32,19 @@
...
       "name": "plugins"
+    },
+    {
+      "name": "storage"
+    }
+  ],
+  "environments": [
+    {
+      "name": "OIQ_OTEL_COLLECTOR_STORAGE",
+      "value": "C:\\Program Files\\observIQ OpenTelemetry Collector",
```
BinaryFissionGames, line 43:
```suggestion
      "value": "[INSTALLDIR]storage",
```
We should do this so that changing the install dir during installation doesn't break anything

windows/wix.json
```diff
@@ -32,6 +32,19 @@
...
+      "permanent": "yes",
```
BinaryFissionGames, line 44:
```suggestion
      "permanent": "no",
```

We should remove this env variable on uninstall

windows/wix.json
```diff
@@ -32,6 +32,19 @@
...
+      "system": "yes",
+      "action": "set",
+      "part": "last"
```
BinaryFissionGames, line 47:
```suggestion
      "part": "all"
```
My understanding is that putting this as "all" will replace the environment variable on install, which is what we want (we don't want to append to the current value of the variable)

https://wixtoolset.org/documentation/manual/v3/xsd/wix/environment.html

BinaryFissionGames (commented):
Just some small tweaks on the env variable for windows, I tested with these suggested changes so they should be good.

jsirianni:
@cpheps I decided to revert the homebrew change for a couple reasons.

I could not successfully create the directory (probably ignorance on my part, with Ruby and Homebrew)

If created in the homebrew install directory, the user will need to identify that directory when configuring the storage extension. I think it is reasonable to expect the user to create their own storage directory instead. Ideally, the directory would live outside of homebrew's scope, allowing upgrades to occur without modifying the storage directory. Further, macOS is generally not used as a long term solution, while Linux and Windows are more production oriented. I think it is expected that we provide a storage directory on those platforms.

I did test the homebrew install directory when configuring a logging pipeline with storage, it works well in the root of the install directory if the user really wants to avoid creating a directory.

BinaryFissionGames (approved):

cpheps (approved):

PR #351, opened 2022-04-14
feat: Speed up container image build. Enable ARM64 image building.

jsirianni:
### Proposed Change

By pulling go build out of the docker build process, github actions can build amd64 and arm64 images in < 2 minutes.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


.github/workflows/release.yml
```diff
@@ -137,6 +137,8 @@ jobs:
       - name: Get Tag From Environment
         id: get-tag
         run: printf '::set-output name=tag::%s' "$(printf '%s' "${{ github.ref }}" | sed 's/refs\/tags\///')"
+      - name: Build Binaries
+        run: make release-test
```
BinaryFissionGames, line 141: I believe this needs the GITHUB_TOKEN environment variable set? I can't remember if that was only failing because the repo was private, or if it's still an issue even with the repo being public.
```yaml
env:
  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

BinaryFissionGames: Oh, this is doing a snapshot, that's why it doesn't error out (don't need GITHUB_TOKEN for snapshot builds).

The internal version for the collector won't be right w/ the snapshot builds, we should avoid the snapshot flag if possible here.

cpheps: I wanted to ask this also this runs the entire release process do we just want the binaries or do we need something special that comes with doing the entire release? If it's just the binaries we could use `make build-all` or [goreleaser build](https://goreleaser.com/cmd/goreleaser_build/).

jsirianni: Goreleaser `build` will only build for the goos/goarch of the runner. We could use `build-all`. We want to avoid a snapshot build, so `release-test` was the wrong choice.

The goal is to use the same binaries shipped with packages. I might experiment with copying the binaries from the build (linux, darwin, windows) stages, and avoid building new binaries all together. That seems like the best solution, maybe somewhat complex.

BinaryFissionGames: Goreleaser build will build based on the environment variables GOOS and GOARCH - you can set these for the job to build for a specific arch + platform, if you desire.

jsirianni:
Ive updated Goreleaser to build the image, this way we build binaries once and use them everywhere. You can view the test run here: https://github.com/observIQ/observiq-otel-collector/runs/6041681014?check_suite_focus=true

.github/workflows/release.yml
```diff
@@ -191,3 +132,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.ORG_GORELEASER_GITHUB_TOKEN }}
           COSIGN_PWD: ${{ secrets.ORG_COSIGN_PWD }}
+      - name: TEMP get images
```
cpheps, line 135: Is the word `TEMP` suppose to be here?

.goreleaser.yml
```diff
@@ -130,7 +166,7 @@ signs:
 release:
   draft: false
   extra_files:
-    - glob: "./observiq-otel-collector*.msi"
+    #- glob: "./observiq-otel-collector*.msi"
```
cpheps, line 169: should this be commented out?

.goreleaser.yml
```diff
@@ -113,6 +113,42 @@ nfpms:
     scripts:
       preinstall: "./scripts/package/preinstall.sh"
       postinstall: ./scripts/package/postinstall.sh
+
+# Build container images with docker buildx (mutli arch builds)
+dockers:
```
cpheps, line 118: I forgot go releaser could do docker this is a nice clean change.

.goreleaser.yml
jsirianni: Nope, was for testing only.

.github/workflows/release.yml
jsirianni: Nope, was for testing only.

jsirianni:
Once thing to mention, the image tagging scheme will create the following
```
observiq/observiq-otel-collector   0                 46317cc5d2f3   2 days ago      462MB
observiq/observiq-otel-collector   0.5               46317cc5d2f3   2 days ago      462MB
observiq/observiq-otel-collector   0.5.0             46317cc5d2f3   2 days ago      462MB
observiq/observiq-otel-collector   latest            46317cc5d2f3   2 days ago      462MB
```

Major and minor version tags will always be updated with the latest major / minor release. I think this is okay as it is an addition to how we do tagging. Historically we just tag once with major.minor.patch.

cpheps (approved):

BinaryFissionGames (approved):
Awesome job, cool to see the container build moved into goreleaser

PR #352, opened 2022-04-14
Use generic node for Mysql and Redis Google examples

jsirianni:
### Proposed Change

We will be using generic_node for all Google examples in the near future. Mysql and Redis will be part of a blog being written right now, so I have updated them earlier than expected.

To use [Google's generic_node](https://cloud.google.com/monitoring/api/resources#tag_generic_node), we need to use hostname as the node_id and set namespace / location. The `resourceattributetransposer` is no longer required as host.name is being preserved as `node_id`.

Screen Shot 2022-04-14 at 4 01 04 PM

Screen Shot 2022-04-14 at 4 01 38 PM

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/config/google_cloud_exporter/mysql/config.yaml
+++ b/config/google_cloud_exporter/mysql/config.yaml
@@ -21,7 +21,7 @@ processors:
       value: "mysql"
       action: upsert
     - key: location
-      value: "us-east1-b"
+      value: "global"
       action: upsert
 
   normalizesums:
--- a/config/google_cloud_exporter/mysql/config.yaml
+++ b/config/google_cloud_exporter/mysql/config.yaml
@@ -14,10 +14,15 @@ processors:
     system:
       hostname_sources: ["os"]
 
-  resourceattributetransposer:
-    operations:
-      - from: host.name
-        to: agent
+  # Used for Google generic_node mapping.
+  resource:
+    attributes:
+    - key: namespace
+      value: "mysql"
+      action: upsert
+    - key: location
+      value: "us-east1-b"
+      action: upsert
 
   normalizesums:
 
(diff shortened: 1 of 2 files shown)


config/google_cloud_exporter/mysql/config.yaml
```diff
@@ -14,10 +14,15 @@ processors:
...
-      - from: host.name
-        to: agent
+  # Used for Google generic_node mapping.
+  resource:
+    attributes:
+    - key: namespace
+      value: "mysql"
+      action: upsert
+    - key: location
+      value: "us-east1-b"
```
BinaryFissionGames, line 24: would `global` be more appropriate here, since we don't have a known zone to put here?

config/google_cloud_exporter/redis/config.yaml
```diff
@@ -12,10 +12,15 @@ processors:
...
-      - from: host.name
-        to: agent
+  # Used for Google generic_node mapping.
+  resource:
+    attributes:
+    - key: namespace
+      value: "redis"
+      action: upsert
+    - key: location
+      value: "us-east1-b"
```
BinaryFissionGames, line 22: Same here, does `global` make more sense?

cpheps (approved):
LGTM. I didn't see anything outside of @BinaryFissionGames' questions

config/google_cloud_exporter/mysql/config.yaml
jsirianni: I 100% agree with you, however, Google has instructed us to use generic_node.

I had the same question with location, their explanation was that it is where "the metrics are stored". It seems odd to me that this is a user facing parameter as it does not line up well with non GCP users.

I believe generic_node does have user facing advantages. The `namespace` parameter is used to split up metrics in the storage backend. I think this helps get around some limitations that global has.

jsirianni: After further discussion, we are talking about `global` as a location value, not the global monitored resource type.

BinaryFissionGames (approved):

PR #353, opened 2022-04-15
feat(plugin): Rabbitmq Logging plugin

cpheps:
### Proposed Change
- Added plugin file for RabbitMQ Logging
- Added test structure to ensure the plugins we supply with the collector load and render with the current version of the plugin receiver
- Ran `make tidy` at high level and caught some items in cleanup

**Note:** I will be adding several more plugins. Once the initial batch is done we will remove the logreceiver and change the release process to bundle the included plugins rather than the `stanza-plugins`.

##### Checklist
- [X] Changes are tested (Tested on a live rabbit system)
- [ ] CI has passed


plugins/rabbitmq_logs.yaml
```diff
@@ -0,0 +1,34 @@
...
+      attributes:
+        log_type: rabbitmq
+      operators:
+        - type: regex_parser
+          regex: '^(?P<timestamp>\d+-\d+-\d+ \d+:\d+:\d+\.\d+\+\d+:\d+) \[(?P<severity>\w+)\] \<(?P<process_id>\d+\.\d+\.\d+)\> (?P<message>.*)'
+          timestamp:
+            parse_from: body.timestamp
+            layout: '%Y-%m-%d %H:%M:%S.%s%j'
+          severity:
+            parse_from: body.severity
```
BinaryFissionGames, line 30: Does this cover all the severity mappings? There are some differences between base stanza's default severity map and otels default severity map.

BinaryFissionGames: For reference, here's what we have in the update-fields branch on stanza-plugins: https://github.com/observIQ/stanza-plugins/blob/5e885ea0023ef1df71194f63fee24d74d085f525/plugins/rabbitmq.yaml#L46

receiver/pluginreceiver/supplied_plugins_test.go
```diff
@@ -0,0 +1,49 @@
...
+		t.Run(fmt.Sprintf("Loading %s", entry.Name()), func(t *testing.T) {
+			t.Parallel()
+			fullFilePath, err := filepath.Abs(filepath.Join(pluginDirPath, entry.Name()))
+			assert.NoError(t, err, "Failed to determine path of file %s", entry.Name())
+
+			// Load the plugin
+			plugin, err := LoadPlugin(fullFilePath)
+			assert.NoError(t, err, "Failed to load file %s", entry.Name())
+
+			_, err = plugin.RenderComponents(map[string]interface{}{})
```
BinaryFissionGames, line 45: Will this work for every plugin?

It's a bridge we can cross when we get to it if not, but are we trying to make all plugins have defaults for all fields?

receiver/pluginreceiver/supplied_plugins_test.go
```diff
@@ -0,0 +1,49 @@
...
+			assert.NoError(t, err, "Failed to render components for plugin %s", entry.Name())
```
StefanKurek, line 46: are "other" errors acceptable for each of these assertions?

receiver/pluginreceiver/supplied_plugins_test.go
cpheps: I believe it should as it doesn't actually load anything from factories but rather applies the loaded config to structs and applies defaults.

I'm mostly just testing does this thing load without issues.

receiver/pluginreceiver/supplied_plugins_test.go
StefanKurek: ignore me. i assumed this method was taking in a string and expecting the errors to have that message

receiver/pluginreceiver/supplied_plugins_test.go
BinaryFissionGames: Yeah, but if a field isn't populated this will fail, right? Like, if one of the options doesn't have a default? Doesn't really matter for this PR, actually, it's fine as is.

plugins/rabbitmq_logs.yaml
```diff
@@ -0,0 +1,34 @@
+version: 0.0.1
+title: RabbitMQ
+description: Log parser for RabbitMQ
+parameters:
+  - name: daemon_log_path
+    type: string
+    default: "/var/log/rabbitmq/rabbit@*.log"
```
jsirianni, line 7: Something I tried to do with more recent stanza plugins, was to use `strings` instead of `string` for most input files. I think generally, globbing is fine but sometimes we end up in a situation where the user needs to specify multiple strings explicitly. 

With Rabbit, it might not matter but I think the cost / complexity associated with strings is not too high.

I think this would be worth standardizing on this early, if we agree.

plugins/rabbitmq_logs.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - name: start_at
+    type: string
+    supported:
+      - beginning
+      - end
+    default: end
+template: |
+  receivers:
+    filelog:
```
jsirianni, line 16: I think it might be worth exposing or setting the following:
- `include_file_path`
- `resource`

Im sure there are not semantic conventions for Rabbitmq logging, but if there were, I think there would be some default `resource` values.

plugins/rabbitmq_logs.yaml
cpheps: Good call out I didn't know we had that branch fixing up for that.

plugins/rabbitmq_logs.yaml
cpheps: That's an interesting take we could do this and I don't think there would be any issues as this maps to the include paths

jsirianni: Stanza plugins is somewhat inconsistent here, so I think we can standardize on it early.

plugins/rabbitmq_logs.yaml
cpheps: Those are good upgrades this initial push is just to get them ported over into the repo. We should apply best practices after this initial push. I think you probably know more about these than I do so I think we should sync up at some point and get these best practices applied.

receiver/pluginreceiver/supplied_plugins_test.go
cpheps: Talked about this offline just a misunderstanding.

receiver/pluginreceiver/supplied_plugins_test.go
cpheps: Possibly, I will cross that bridge when we come to it. I'm not sure to be honest if that's a case we have.

plugins/rabbitmq_logs.yaml
Dylan-M: I agree completely that we should standardize on ability to include multiple strings. Then, if appropriate, we can make exceptions the other way if there is good reason to do so.

Dylan-M (commented):
My one issue here, and I have this with all the base Stanza plugins too, is lack of comments within the plugin that explains what is going on. Those comments are of a huge help when someone new to the product comes along and is trying to decipher what is happening. I feel that all plugin yaml files should include comments that explain each piece, why it is there, and what it affects.

This becomes even more important with more complex plugins that include flow control by sending things to specific outputs.

For example, with syslog, we could include an automatic detection (which I've now done for a couple of customers) to determine which syslog format it is rather than making the customer specify.

In fact, we could probably roll all three syslog plugins into a single one, instead of having separate ones for `syslog`, `rsyslog`, and `syslog-ng`.

The simpler we can make things for the customer in the long run, the better adoptions can be. All while still allowing customer to override and gain complex capabilities if they need to. 

I guess that leads me to another capability that should be available: ability to override outputs within the plugin. For example, again with syslog, being able to provide a custom parser and skip the plugin supplied syslog_parser.

Dylan-M:
For reference, this is what I've done for Telus' second team we're currently working with on a POC regarding their rsyslog where they have appliances that send different formats and they cannot change currently. I've had to skip using the plugin for them, because of this. I've snipped the detection portion of their config and pasted it here. It also includes detection for stuff that doesn't match either format, and adds a label to it that makes it easy to find and identify.

```
  - id: rsyslog_router
    type: router
    routes:
      - expr: '$.syslog_entry matches "<\\d{2}>\\w{3} \\d{2} \\d{2}:\\d{2}:\\d{2}\\s.*"'
        output: rsyslog_rfc3164_parser
      - expr: '$.syslog_entry matches "<\\d{2}>\\d \\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}.\\d{3}Z\\s.*"'
        output: rsyslog_rfc5424_parser
    default: needs_parsing
 
  - id: rsyslog_rfc3164_parser
    type: syslog_parser
    protocol: rfc3164
    parse_from: syslog_entry
    output: output
 
  - id: rsyslog_rfc5424_parser
    type: syslog_parser
    protocol: rfc5424
    parse_from: syslog_entry
    output: output

  - id: needs_parsing
    type: add
    field: $labels.parsing
    value: "needs parsing"
    output: output
```

plugins/rabbitmq_logs.yaml
cpheps: @BinaryFissionGames @jsirianni any insight on the severity. I tested this out using the install scripts we had for ops-agent and the severity does not map like we have in in the severity parsing. 

Looks like the [ops-agent parsing](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/apps/rabbitmq.go#L49-L52) we did also just did normal mapping. I'm thinking of changing this back any concerns with that?

cpheps: Actually looks like the regex on that branch doesn't work at all. Taking another look.

cpheps: Looks like the timestamp doesn't work either. I'm reverting back to what I had before as that does look like it works and matches ops-agent.

BinaryFissionGames: Yeah, I would just use something similar to what's in the ops agent.

Just to explain/expand on the differences a bit and what's in that file; The mapping there are basically just to make up the difference with stanza (I believe these mappings get applied over the defaults? I forget the specifics) - so a lot of these mappings on that branch are probably unnecessary, and were just applied as a catch all so we didn't have to audit each specific technology.

What we have in the ops-agent would actually work just fine with the default, if we wanted those mappings ("noti" -> "default" seems weird to me, though, that seems more like it should be warn).

plugins/rabbitmq_logs.yaml
```diff
@@ -0,0 +1,34 @@
...
+      include: [ '{{ .daemon_log_path }}' ]
+      multiline:
+        line_start_pattern: '\d+-\d+-\d+ \d+:\d+:\d+\.\d+\+\d+:\d+'
+      start_at: {{ .start_at }}
+      attributes:
+        log_type: rabbitmq
+      operators:
+        - type: regex_parser
+          regex: '^(?P<timestamp>\d+-\d+-\d+ \d+:\d+:\d+\.\d+\+\d+:\d+) \[(?P<severity>\w+)\] \<(?P<process_id>\d+\.\d+\.\d+)\> (?P<message>.*)'
```
Dylan-M, line 25: This regex for the timestamp is fairly weak (too permissive). I can provide a better regex that limits to the correct max lengths for fields so that this cannot trigger accidentally. I just need to see a sample or three. :)

Essentially, if the time for example is always 2 digits (02:07:55 for example), then the time piece should be \d{2}:\d{2}:... This prevents false triggering, a serious danger with using the `+` in regex too freely.

Dylan-M (changes requested):

plugins/rabbitmq_logs.yaml
cpheps: updated this. Left the millisecond as `\d+` as I was confident that precision was always going to be that many digits long.

cpheps:
@Dylan-M I appreciate all your feedback and thoughts on making these ultimately more user friendly. This work is being done in two phases. This first phase it to port these over from our [stanza-plugins](https://github.com/observiq/stanza-plugins) repo. The follow on phase will be to polish them and apply any standardizations we need across them all.

plugins/rabbitmq_logs.yaml
```diff
@@ -0,0 +1,38 @@
...
+        # 2022-04-18 12:03:53.856675+00:00 [info] <0.230.0> Running boot step rabbit_looking_glass defined by app rabbit
+        - type: regex_parser
+          regex: '^(?P<timestamp>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d+\+\d{2}:\d{2}) \[(?P<severity>[a-z]+)\] \<(?P<process_id>\d+\.\d+\.\d+)\> (?P<message>.*)'
+          timestamp:
+            parse_from: timestamp
+            layout: '%Y-%m-%d %H:%M:%S.%s%j'
+          severity:
+            parse_from: severity
+            mapping:
+              trace: 'noti'
```
BinaryFissionGames, line 34: `noti` is somewhere between info and warning, I think having it at the trace level will be confusing (might be filtered out in the backend often). IMO it should be mapped to something like `info2` or `warning` instead of trace.

cpheps: I think `info2` sounds good. In ops-agent we put it at `default`. The messages for `noti` made it hard to get a feeling for what it should be.

BinaryFissionGames (approved):

jsirianni (approved):

PR #355, opened 2022-04-18
feat: Wrap Google Cloud Exporter

jmwilliams89:
### Proposed Change
This will add an exporter that wraps the official Google Cloud Exporter. The intent of this feature is to reduce the complexity of configuration when sending telemetry data to GCP. The exporter includes a number of processors considered essential to the user experience. The README explains how this data is processed before it is exported.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


exporter/googlecloudexporter/config.go
```diff
@@ -0,0 +1,175 @@
...
+	attributerConfig := attributerFactory.CreateDefaultConfig().(*resourceprocessor.Config)
+	return attributerConfig
+}
+
+// createDefaultTransposerConfig creates a default transposer config
+func createDefaultTransposerConfig() *resourceattributetransposerprocessor.Config {
+	transposerFactory := resourceattributetransposerprocessor.NewFactory()
+	transposerConfig := transposerFactory.CreateDefaultConfig().(*resourceattributetransposerprocessor.Config)
+	transposerConfig.Operations = []resourceattributetransposerprocessor.CopyResourceConfig{
+		{
```
cpheps, line 143: I'm not as knowledgeable on these but are these relative to the host receiver? If not where do they come from?

exporter/googlecloudexporter/README.md
```diff
@@ -0,0 +1,60 @@
...
+| `endpoint`          |                       | `false`  | The endpoint used when exporting telemetry data. |
+| `location`          | `us-east1-b`          | `false`  | The GCP location attribute attached to telemetry data. |
+| `namespace`         | `{hostname}`          | `false`  | The GCP namespace attribute attached to telemetry data. |
+| `user_agent`        | `observIQ-otel-agent` | `false`  | Overrides the user agent used when making requests. |
+| `timeout`           | `12s`                 | `false`  | The timeout for API calls. |
+| `resource_mappings` | [See below](#resource-mapping-default)         | `false`  | Defines a mapping of resources from source to target. |
+| `retry_on_failure`  |                       | `false`  | Handle retries when sending data to Google Cloud fails. |
+| `sending_queue`     |                       | `false`  | Determines how telemetry data is buffered before exporting. |
+| `batch`             |                       | `false`  | The config of the exporter's [batch processor](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor). |
```
cpheps, line 18: This is a bit of a pain but it's probably worth making sure this documentation points towards the documentation of the version of these processors that we support. With how fast OTel changes we could have documentation that isn't correct for what we have packaged.

cpheps:
You'll need to add an entry to `dependabot.yml`. Copy the `resourceattributetransposerprocessor` one we mostly just want to ignore being notified of otel updates.

exporter/googlecloudexporter/config.go
jmwilliams89: Yes, they're relative to the host receiver. @jsirianni compiled this initial list. I imagine this mapping will grow with subsequent sources. But for now, I wanted to keep it simple. Especially if the Google Cloud exporter is updated to handle resource attributes better. That will allow us to remove this processor altogether.

exporter/googlecloudexporter/README.md
jmwilliams89: Yeah, I believe that's a worthwhile change. The only issue will be we'll have to remember to update this documentation every time we up the version.

cpheps: Yeah, I agree this is the same issue we have with the current docs. Ideally if they're all on the same version it's easy to find replace the entire repo but I doubt that's the case. Might be worth trying to create tooling to generate these.

cpheps (approved):

exporter/googlecloudexporter/config.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourceprocessor"
+	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/processor/batchprocessor"
+	"go.uber.org/multierr"
+)
+
+const (
+	defaultMetricPrefix = "workloads.googleapis.com"
+	defaultUserAgent    = "observIQ-otel-collector"
+	defaultLocation     = "us-east1-b"
```
BinaryFissionGames, line 34: IMO, we should use `global` as the location, it's just more generic and is what's currently in our examples that map to generic_node.

exporter/googlecloudexporter/config.go
```diff
@@ -0,0 +1,175 @@
...
+	err = multierr.Append(err, c.NormalizeConfig.Validate())
+	err = multierr.Append(err, c.DetectorConfig.Validate())
+	err = multierr.Append(err, c.AttributerConfig.Validate())
+	err = multierr.Append(err, c.TransposerConfig.Validate())
+	return err
+}
+
+// createDefaultConfig creates the default config for the exporter
+func createDefaultConfig() config.Exporter {
+	defaultNamespace, _ := os.Hostname()
```
BinaryFissionGames, line 65: Is this what namespace should really be by default? I'm not really sure, semantically, what the namespace should be, but the resource detection processor + node_id already fills this role of labelling w/ the hostname, right?

exporter/googlecloudexporter/exporter.go
```diff
@@ -0,0 +1,150 @@
...
+func (e *Exporter) ConsumeLogs(ctx context.Context, ld pdata.Logs) error {
+	if e.logsConsumer == nil {
+		return nil
+	}
+	return e.logsConsumer.ConsumeLogs(ctx, ld)
+}
+
+// Capabilities returns the capabilities of the exporter
+func (e *Exporter) Capabilities() consumer.Capabilities {
+	return consumer.Capabilities{MutatesData: true}
```
BinaryFissionGames, line 67: This shouldn't be true, right? In general, exporters should not mutate log records.

BinaryFissionGames: Actually, you're right, this does mutate records, which has me wondering about how it affects adjacent exporters, or if this is actually OK.

Need to take a closer look.

exporter/googlecloudexporter/config.go
jsirianni: Confirmed, global is working well. We were not aware of global previously.

exporter/googlecloudexporter/exporter.go
BinaryFissionGames: Here's where the warning about doing this is: https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter#data-ownership

exporter/googlecloudexporter/config.go
jsirianni: Im leaning more toward the source name ("postgresql", "mysql", etc) but I believe it is not currently possible unless we set a resource somewhere. I think hostname is okay.

Google uses `namespace` to organize data on the backend. They instructed us to use "something" other than default.

BinaryFissionGames: If something other than `default` is needed, then this is probably OK then. It just seems redundant from a user perspective, wasn't aware of the technical reason.

exporter/googlecloudexporter/exporter.go
```diff
@@ -0,0 +1,150 @@
...
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/model/pdata"
+)
+
+// Exporter is a google cloud exporter wrapped with additional functionality
+type Exporter struct {
```
BinaryFissionGames, line 27: I think this should be unexported, generally only the factory is exported in these components.

go.mod
```diff
@@ -413,14 +413,14 @@ require (
 	golang.org/x/sys v0.0.0-20220209214540-3681064d5158
 )
 
-replace github.com/open-telemetry/opentelemetry-collector-contrib/exporter/googlecloudexporter v0.46.0 => github.com/observiq/opentelemetry-collector-contrib/exporter/googlecloudexporter v0.0.0-20220304152956-bb36c08bd895
-
```
BinaryFissionGames, line 417: Removing this breaks logs: 
```
2022/04/19 09:46:16 collector server run finished with error: cannot build exporters: error creating googlecloud exporter: failed to create logs exporter: telemetry type is not supported
```

When building, only the root modules replace directives take effect.

jmwilliams89: Good catch. I didn't consider how that replace precedence would work.

exporter/googlecloudexporter/config.go
jmwilliams89: Can do. I was not aware that global was an option.

exporter/googlecloudexporter/exporter.go
jmwilliams89: Interesting. Do you understand the use case for setting up parallel pipelines with the same receivers? This seems like an extreme edge case.

jmwilliams89: Nevermind. I think I understand the use case after looking closer. Effectively, the issue is having multiple exporters in a single pipeline. There's potential to mutate data that is being sent to the other exporter. Do you see any way around this limitation?

BinaryFissionGames: It's not that strange to want to send data to two locations at once; for instance, you might want to send to some local prometheus server for short term viewing, but send to GCP maybe for longer term storage (or possibly even another location for REALLY long storage). Another example is debugging using the logging exporter, for example.

BinaryFissionGames: I think we've had customers ask about this in the past, also, but I'm not sure on the specifics.

BinaryFissionGames: Only way around the limitation I see is to make a copy of the record before passing to the first consumer.

BinaryFissionGames (approved):

PR #358, opened 2022-04-19
feat: Remove logreceiver

cpheps:
### Proposed Change
As we're porting over the [stanza-plugins](https://github.com/observiq/stanza-plugins) to internally shipped plugins via the plugins receiver we no longer have a need for the `logsreceiver`. The `logsreceiver` also locked us to an older version of OpenTelemetry as it was stuck on `opentelemetry-log-collection` v0.24.0. We'll now be able to update to the latest version of OpenTelemetry.

Updated the release scripts to ship the internal plugins rather than `stanza-plugins`

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


cpheps:
@BinaryFissionGames Can you check the release scripts and verify I didn't mess up including plugins in the MIS build? I wasn't sure how that worked fully.

dehaansa (approved):
I think technically full removal isn't exactly deprecation but 🙃

BinaryFissionGames (approved):

PR #366, opened 2022-04-21
fix: Wrapped google exporter metric prefix

jsirianni:
### Proposed Change

- `workloads.googleapis.com` needs to be `workload.googleapis.com`. 
- Update hostmetrics config to leverage wrapped Google exporter

Screenshot from 2022-04-21 14-27-19

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/googlecloudexporter/config.go
+++ b/exporter/googlecloudexporter/config.go
@@ -15,8 +15,6 @@
 package googlecloudexporter
 
 import (
-	"os"
-
 	"github.com/GoogleCloudPlatform/opentelemetry-operations-collector/processor/normalizesumsprocessor"
 	"github.com/mitchellh/mapstructure"
 	"github.com/observiq/observiq-otel-collector/processor/resourceattributetransposerprocessor"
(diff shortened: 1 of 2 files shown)


exporter/googlecloudexporter/config.go
```diff
@@ -96,7 +92,7 @@ func createDefaultGCPConfig() *gcp.Config {
 					TargetKey: "location",
 				},
 				{
-					SourceKey: "namespace",
+					SourceKey: "host.name",
 					TargetKey: "namespace",
```
BinaryFissionGames, line 96: These changes to namespace make it unconfigurable; Changing namespace in the config will not change it on exported metrics. I think we do need the 2 separate keys to handle this case.

jsirianni: Yeah, I've reverted the change. It is out of scope anyway.

BinaryFissionGames (approved):

schmikei (approved):

PR #370, opened 2022-04-25
feat: mysql, redis, mongodb prometheus exporter example configs

algchoo:
### Proposed Change

Added example configs for the prometheus exporter
* MySQL
* Redis
* MongoDB

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/config/prometheus/README.md
+++ b/config/prometheus/README.md
@@ -1,0 +1,15 @@
+# Prometheus Exporter Prerequisites
+
+## Configure Prometheus to scrape the collector
+
+Using the basic configuration provided in [these docs](https://prometheus.io/docs/prometheus/latest/getting_started/#configuring-prometheus-to-monitor-itself) instead of:
+
+```
+static_configs:
+    - targets: ['localhost:<PROMETHEUS_PORT>']
+```
+edit the config target the collector:
+```
+static_configs:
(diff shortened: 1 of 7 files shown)


config/prometheus/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+```
+static_configs:
+    - targets: ['localhost:<PROMETHEUS_PORT>']
+```
+edit the config target the collector:
+```
+static_configs:
+    - targets: ['localhost:<COLLECTOR_PORT>']
+```
```
jsirianni, line 15: I do not think the linked doc is very helpful. We are not configuring prometheus to scrape itself.

I think this section can just say that "prometheus should be configured to scrape the exporter's configured ip and port. Its possible that the user's prometheus server is also scraping other targets.

config/prometheus/README.md
```diff
@@ -0,0 +1,15 @@
+# Prometheus Exporter Prerequisites
+
```
jsirianni, line 2: Add small blurb explaining that prometheus exporter can be scraped by a prometheus server. See the google managed prometheus example.

jsirianni (changes requested):

jsirianni (approved):

PR #372, opened 2022-04-26
deps: Removed varnish as it was moved to the operations collector

cpheps:
### Proposed Change
Removed Varnish as it was added to the operations collector in https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/94. This will be imported into the collector in a future PR.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


schmikei (approved):
Is there any reason to not import now rather than later?

dehaansa (approved):

cpheps:
> Is there any reason to not import now rather than later?

Yeah two reasons:
1. We weren't importing it before so. If we were I would have done the full swap.
2. There's a chance we'll hit some issues with otel imports that we'll have to solve when v0.50.0 releases this week so I'm waiting on that to only solve them once.

PR #374, opened 2022-04-26
feat(plugin): Added Mongodb log plugin

cpheps:
### Proposed Change
Added the Mongodb log plugin MVP as a port from `stanza-plugins`. 

**Note1:** The stanza-plugin included the ability to read from k8s. There is no way I could find in otel to get logs from k8s and add the operator parsing. The [k8seventsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8seventsreceiver) does not support operators. This would be doable if the logs transforms processor is approved open-telemetry/opentelemetry-collector-contrib#9133.

**Note2:** The remove operator is currently not present in the filelog receiver. There is a PR to add it so I commented out that operator until the PR is approved. Tracked in open-telemetry/opentelemetry-collector-contrib#9524

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/mongodb_logs.yaml
+++ b/plugins/mongodb_logs.yaml
@@ -55,6 +55,8 @@ template: |
                 - D5
           output: add_log_type
 
+        # Example log line:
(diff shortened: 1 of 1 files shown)


BinaryFissionGames (approved):

jsirianni (approved):
I think it is okay to remove k8s support for now, allowing us to re-asses how we handle it going forward.

PR #376, opened 2022-05-02
feat(nginx): Add nginx logging plugin

BinaryFissionGames:
### Proposed Change
* Add a logging plugin for NGINX access + error logs. (ported from stanza-plugins)

Tested against a running NGINX instance to make sure access + error logs were parsing correctly.

As noted for Mongodb, k8s support has been removed.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/nginx_logs.yaml
+++ b/plugins/nginx_logs.yaml
@@ -121,9 +121,9 @@ template: |
       pipelines:
         logs:
           receivers:
-          {{ if .enable_access_log }}
-          - filelog/access_log
-          {{ end }}
-          {{ if .enable_error_log }}
-          - filelog/error_log
-          {{ end }}
+            {{ if .enable_access_log }}
+            - filelog/access_log
+            {{ end }}
+            {{ if .enable_error_log }}
+            - filelog/error_log
+            {{ end }}
--- a/plugins/nginx_logs.yaml
+++ b/plugins/nginx_logs.yaml
@@ -1,0 +1,129 @@
+version: 0.0.1
+title: NGINX
+description: Log parser for NGINX
+parameters:
+  # The observiq log format is defined as:
+  # log_format observiq '{"remote_addr":"$remote_addr","remote_user":"$remote_user","time_local":"$time_local","request":"$request","status":"$status","body_bytes_sent":"$body_bytes_sent","http_referer":"$http_referer","http_user_agent":"$http_user_agent","request_length":"$request_length","request_time":"$request_time","upstream_addr":"$upstream_addr","upstream_response_length":"$upstream_response_length","upstream_response_time":"$upstream_response_time","upstream_status":"$upstream_status","proxy_add_x_forwarded_for":"$proxy_add_x_forwarded_for","bytes_sent":"$bytes_sent","time_iso8601":"$time_iso8601","upstream_connect_time":"$upstream_connect_time","upstream_header_time":"$upstream_header_time","http_x_forwarded_for":"$http_x_forwarded_for"}';
+  # See https://docs.nginx.com/nginx/admin-guide/monitoring/logging/#setting-up-the-access-log for more information on configuring log format
+  - name: log_format
+    type: string
+    supported:
+      - default
+      - observiq
+    default: default
+  - name: enable_access_log
+    type: bool
+    default: true
+  - name: access_log_paths
+    type: "[]string"
+    default: 
+      - "/var/log/nginx/access.log*"
+  - name: enable_error_log
+    type: bool
+    default: true
+  - name: error_log_paths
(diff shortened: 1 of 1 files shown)


plugins/nginx_logs.yaml
```diff
@@ -0,0 +1,129 @@
...
+        attributes:
+          log_type: 'nginx.access'
+        start_at: {{ .start_at }}
+        encoding: {{ .encoding }}
+        operators:
+          {{ if eq .log_format "default" }}
+          - type: regex_parser
+            regex: '^(?P<http_request_remoteIp>[^ ]*) (?P<host>[^ ]*) (?P<user>[^ ]*) \[(?P<time_local>[^\]]*)\] "(?P<http_request_requestMethod>\S+)(?: +(?P<http_request_requestUrl>[^\"]*?)(?: +(?P<http_request_protocol>\S+))?)?" (?P<http_request_status>[^ ]*) (?P<http_request_responseSize>[^ ]*)(?: "(?P<http_request_referer>[^\"]*)" "(?P<http_request_userAgent>[^\"]*)")?$'
+            timestamp:
+              parse_from: "$body.time_local"
```
cpheps, line 61: I hit this issue with other plugins but should this be `$body.time_local`? I had to use `attributes.time_local` when the collector was using the latest logging library.

cpheps: Same question everywhere `$body` exists.

plugins/nginx_logs.yaml
```diff
@@ -0,0 +1,129 @@
...
+                info2: 3xx
+                warn: 4xx
+                error: 5xx
+      {{end}} # end error log receiver
+
+    service:
+      pipelines:
+        logs:
+          receivers:
+          {{ if .enable_access_log }}
```
cpheps, line 124: Does this need to be indented one?

plugins/nginx_logs.yaml
BinaryFissionGames: I think you're right, I was testing with v0.5.0, not latest.

plugins/nginx_logs.yaml
BinaryFissionGames: indented seems more consistent with other plugins, good catch

plugins/nginx_logs.yaml
BinaryFissionGames: Fixed this and tested with the latest collector; Thanks for pointing this out, didn't realise those breaking changes weren't yet made on our latest release.

cpheps (approved):

armstrmi (approved):

BinaryFissionGames:
Had to update this PR, noticed `preserve_to` is still present in this version of the collector. In v0.50.0 of the otel collector, it should be removed, so we'll have to make some changes to the plugins to reflect that once we've upgraded our dependencies.

Here's what the logs look like now:
Access log:
```
LogRecord #99
ObservedTimestamp: 1970-01-01 00:00:00 +0000 UTC
Timestamp: 2022-05-03 16:14:02 +0000 UTC
Severity: Warn
Body: 172.17.0.2 - - [03/May/2022:12:14:02 -0400] "GET /index.html HTTP/1.1" 404 152 "-" "Go-http-client/1.1"
Attributes:
     -> http_request_userAgent: STRING(Go-http-client/1.1)
     -> http_request_requestMethod: STRING(GET)
     -> http_request_requestUrl: STRING(/index.html)
     -> http_request_remoteIp: STRING(172.17.0.2)
     -> http_request_protocol: STRING(HTTP/1.1)
     -> http_request_status: STRING(404)
     -> http_request_responseSize: STRING(152)
     -> http_request_referer: STRING(-)
     -> host: STRING(-)
     -> user: STRING(-)
     -> log_type: STRING(nginx.access)
     -> log.file.name: STRING(access.log)
Trace ID: 
Span ID: 
Flags: 0
```

Error log:
```
LogRecord #79
ObservedTimestamp: 1970-01-01 00:00:00 +0000 UTC
Timestamp: 2022-05-03 16:20:29 +0000 UTC
Severity: Info
Body: 2022/05/03 12:20:29 [info] 10549#10549: *160747 client 172.17.0.2 closed keepalive connection
Attributes:
     -> log.file.name: STRING(error.log)
     -> subrequest: STRING()
     -> pid: STRING(10549)
     -> tid: STRING(10549)
     -> referer: STRING()
     -> host: STRING()
     -> upstream: STRING()
     -> connection: STRING(160747)
     -> message: STRING(client 172.17.0.2 closed keepalive connection)
     -> server: STRING()
     -> log_type: STRING(nginx.error)
     -> client: STRING()
     -> request: STRING()
```

PR #388, opened 2022-05-04
feat(syslog): Add a syslog plugin with a low-flow mode

BinaryFissionGames:
### Proposed Change
* Add a syslog plugin
* Add a low-flow mode to the plugin
  * Filters out debug logs
  * Filters out factory_id, process_id, and priority

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


cpheps (approved):

StefanKurek (approved):
This one's kind of annoying to review since we don't have the original config right here to compare against. Either way looks good enough, but I have not manually tested it...

PR #390, opened 2022-05-04
fix(nginx): Fix incorrect type for error_log_paths

BinaryFissionGames:
### Proposed Change
* Change error_log_paths to be of type string
* Add a test to verify that default arguments match the types of their parameters
* Fix the default plugin render test to actually test all plugins, instead of the last one in the directory listing

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/pluginreceiver/supplied_plugins_test.go
+++ b/receiver/pluginreceiver/supplied_plugins_test.go
@@ -33,18 +33,18 @@ func TestValidateSuppliedPlugins(t *testing.T) {
 	require.NoError(t, err)
 
 	for _, entry := range entries {
-		entry := entry
+		entryName := entry.Name()
 		t.Run(fmt.Sprintf("Loading %s", entry.Name()), func(t *testing.T) {
 			t.Parallel()
-			fullFilePath, err := filepath.Abs(filepath.Join(pluginDirPath, entry.Name()))
-			assert.NoError(t, err, "Failed to determine path of file %s", entry.Name())
+			fullFilePath, err := filepath.Abs(filepath.Join(pluginDirPath, entryName))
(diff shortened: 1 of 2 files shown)


receiver/pluginreceiver/supplied_plugins_test.go
```diff
@@ -33,6 +33,7 @@ func TestValidateSuppliedPlugins(t *testing.T) {
 	require.NoError(t, err)
 
 	for _, entry := range entries {
+		entry := entry
```
cpheps, line 36: is this on purpose? shadowing a variable with a copy of itself?

BinaryFissionGames: Yes, this is on purpose. I've seen this pattern a few times (e.g. in google's ops agent), it's to fix this issue:

https://stackoverflow.com/questions/69984933/goroutine-inside-for-loop-takes-only-the-last-value-of-slice

cpheps: Oh I see because of the `t.Parrallel`. Yeah I was wondering if you were trying to solve the iterator pointer problem but I didn't see any concurrency at first.

Can we just grab the entry name into a variable since that seems all we want, or rename the variable being assigned to so we don't shadow. I know it should be ok but I'd rather try to avoid confusion where possible.

BinaryFissionGames: Sure, we can do that. The pattern actually comes from the golang FAQ, btw: 

https://go.dev/doc/faq#closures_and_goroutines

cpheps (approved):

PR #393, opened 2022-05-05
fix: Config in linux package should be marked "config|noreplace"

BinaryFissionGames:
### Proposed Change
* Move the "config|noreplace" type from LICENSE to config.yaml

This is needed to preserve config files between upgrades.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -78,13 +78,13 @@ nfpms:
           mode: 0640
           owner: observiq-otel-collector
           group: observiq-otel-collector
+        type: config|noreplace
       - src: LICENSE
         dst: /opt/observiq-otel-collector/LICENSE
         file_info:
           mode: 0644
           owner: observiq-otel-collector
           group: observiq-otel-collector
-        type: config|noreplace
       - src: release_deps/opentelemetry-java-contrib-jmx-metrics.jar
         dst: /opt/opentelemetry-java-contrib-jmx-metrics.jar
         file_info:


cpheps (approved):

.goreleaser.yml
```diff
@@ -78,13 +78,13 @@ nfpms:
           mode: 0640
           owner: observiq-otel-collector
           group: observiq-otel-collector
+        type: config|noreplace
```
StefanKurek, line 81: @BinaryFissionGames i don't know anything about this yaml file. can you give me a little more context of why this change is needed here (and why it's no longer needed for LICENSE)?

BinaryFissionGames: This is a yaml for [goreleaser](https://goreleaser.com/intro/), which we use to manage our builds and releases. This file specifies how we build releases (The main exception being the MSI, which is built separately, then included in the release in this file).

This specific section is for specifying how to build the linux packages, general stuff that can be here is here: https://goreleaser.com/customization/nfpm/

What's being changed here is marking the config file (config.yaml) as a config file to the package managers (dpkg and rpm), so that it isn't just replaced when updating:

https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html
https://www.debian.org/doc/debian-policy/ap-pkg-conffiles.html

LICENSE isn't a config file, as in, a user should not be changing it and it's authoritative as is. With config.yaml, we would like to preserve changes to it.

StefanKurek (approved):

PR #395, opened 2022-05-05
feat: Update macOS install script to allow generation of manager.yaml

BinaryFissionGames:
### Proposed Change
* Add flags for endpoint, labels, and secret-key
  * endpoint must be specified if labels or secret-key is specified
  * These parameters may also be passed in using environment variables
* If endpoint is specified, generate a "manager.yaml" file with a randomly generated UUID for agent_id, as well as labels and secret-key, if specified.
* Remove the manager.yaml on uninstall

This PR allows the install script to create a manager.yaml with supplied information, which will be used to configure the collector for connecting with an OpAMP compatible server.

Tested on macOS Big Sur v11.4

Note: The internal ticket for this specified `manager.yml`, however I opted to go with `manager.yaml` to keep consistency with the extension we've been using for yaml files in this repository.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/install/install_macos.sh
+++ b/scripts/install/install_macos.sh
@@ -163,13 +163,7 @@ Usage:
 
   $(fg_yellow '-u, --upgrade')
       Stops the collector services and upgrades the collector via brew.
-  
-  $(fg_yellow '-e, --endpoint')
-      Defines the endpoint of an OpAmp compatible management server for this collector install.
-      
-      Specifying this will install the collector in a managed mode, as opposed to the
-      normal headless mode.
-  
+   
   $(fg_yellow '-e, --endpoint')
       Defines the endpoint of an OpAMP compatible management server for this collector install.
       This parameter may also be provided through the ENDPOINT environment variable.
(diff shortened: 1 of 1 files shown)


cpheps (approved):

scripts/install/install_macos.sh
```diff
@@ -161,6 +163,32 @@ Usage:
...
   $(fg_yellow '-u, --upgrade')
       Stops the collector services and upgrades the collector via brew.
+  
+  $(fg_yellow '-e, --endpoint')
+      Defines the endpoint of an OpAmp compatible management server for this collector install.
+      
+      Specifying this will install the collector in a managed mode, as opposed to the
+      normal headless mode.
+  
+  $(fg_yellow '-e, --endpoint')
```
StefanKurek, line 173: @BinaryFissionGames is this supposed to be in here twice?

BinaryFissionGames: It was not supposed to be there twice, good catch!

StefanKurek (approved):

armstrmi (approved):

PR #404, opened 2022-05-10
feat: Update Windows MSI to allow generation of manager.yaml

BinaryFissionGames:
### Proposed Change
* Allow user to optionally configure OpAMP parameters through the installer UI:
Screen Shot 2022-05-10 at 11 05 54 AM
Screen Shot 2022-05-10 at 11 06 14 AM
  * Endpoint is required:
Screen Shot 2022-05-10 at 11 06 54 AM

* If a user is upgrading, skip install directory + opamp config steps
  * This change was made because we do not regenerate the opamp config when updating; So having these steps were confusing. It also led to a problem where the user config would be destroyed if they changed the directory they had installed to. This simplified workflow should mitigate that, while also being more painless for users.

* Installing w/ OpAMP configuration using MSIExec would look like this:
  * `msiexec /i "https://github.com/observIQ/observiq-otel-collector/releases/latest/download/observiq-otel-collector.msi" /quiet ENABLEMANAGEMENT=1 OPAMPENDPOINT=<your_endpoint> OPAMPSECRETKEY=<secret-key> OPAMPLABELS=<comma-separated-labels>`

##### Checklist

- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/windows/install/generate-manager-yaml.ps1
+++ b/windows/install/generate-manager-yaml.ps1
@@ -1,0 +1,29 @@
+param (
+    [string] $install_dir = $null,
+    [string] $endpoint = $null,
+    [string] $secret_key = $null,
+    [string] $labels = $null
+)
+
+Write-Output $install_dir
+Write-Output $endpoint
+Write-Output $secret_key
+Write-Output $labels
+
+if([string]::IsNullOrEmpty($endpoint)) {
+    # Don't generate yaml if endpoint isn't specified
+    Write-Output "Endpoint not specified; Not writing output yaml"
+    return 0
+}
+
+Write-Output "Writing manager yaml"
+$yaml="endpoint: `"$endpoint`"{0}" -f [environment]::NewLine
+if(![string]::IsNullOrEmpty($secret_key)) {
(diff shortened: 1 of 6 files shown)


cpheps (approved):

andykellr:
Saw OpAMP and figured I would take a look...

Two small pieces of feedback:

1. Instead of `Labels (Comma separated)`, `Labels (name1=value1,name2=value2,...)` might be more clear regarding the format.
2. Labels and Secret Key aren't technically OpAMP things. They're Iris things and Iris uses OpAMP. I'm not sure if we are concerned about that.

BinaryFissionGames:
Thanks for taking a look, @andykellr!

I've changed the wording to `Labels (name1=value1,name2=value2,...)`, that definitely is more clear.
Screen Shot 2022-05-11 at 9 17 11 AM

Re: Labels and secret key, this is a good callout. I've been avoiding calling out Iris until it's publicly available - I think it's fine as is for now, and we can re-evaluate the wording and structure here later when it's public.

armstrmi (approved):

PR #405, opened 2022-05-10
feat: iis example config for google cloud exporter

algchoo:
### Proposed Change

Added example configuration using iis receiver and exporting to google cloud
* tested configuration on Windows Server 2019

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/config/google_cloud_exporter/iis/config.yaml
+++ b/config/google_cloud_exporter/iis/config.yaml
@@ -13,7 +13,5 @@ service:
       receivers:
       - iis
       processors:
-      - normalizesums
-      - batch
       exporters:
       - googlecloud
--- a/config/google_cloud_exporter/iis/config.yaml
+++ b/config/google_cloud_exporter/iis/config.yaml
@@ -1,0 +1,30 @@
+receivers:
+  iis:
+    collection_interval: 10s
+
+processors:
+
+  normalizesums:
+
+  batch:
+
+exporters:
+  googlecloud:
+    retry_on_failure:
+      enabled: false
+    metric:
+      prefix: workload.googleapis.com
+    resource_mappings:
+      - source_type: ""
+        target_type: generic_node
+
+service:
+  pipelines:
+    metrics:
+      receivers:
+      - iis
+      processors:
+      - normalizesums
+      - batch
+      exporters:
+      - googlecloud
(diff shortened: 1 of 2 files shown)


config/google_cloud_exporter/iis/config.yaml
```diff
@@ -0,0 +1,30 @@
+receivers:
+  iis:
+    collection_interval: 10s
```
jsirianni, line 3: Lets be consistent with 60s collection interval. I believe the Google minimum is 30 seconds.

config/google_cloud_exporter/iis/config.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+processors:
```
jsirianni, line 5: We probably need resource detection, resource, and resource attributes transposer here. I am curious to see how the metrics look into the metric explorer.

jsirianni (changes requested):

config/google_cloud_exporter/iis/config.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+exporters:
+  googlecloud:
+    retry_on_failure:
+      enabled: false
+    metric:
+      prefix: workload.googleapis.com
+    resource_mappings:
+      - source_type: ""
+        target_type: generic_node
```
BinaryFissionGames, line 19: Now that v0.6.0 is released, we can simplify these a LOT. The exporter now has:
- Built in normalize sums processor (doesn't need to be configured separately)
- Built in Batch processor (doesn't need to be configured separately)
- Automatic mapping to generic_node
- metric prefix is default set to "workload.googleapis.com"
- retry_on_failure is default disabled

I think from now on, we should try using the new exporter defaults with these configs.

Let me know if any of that was unclear.

https://github.com/observIQ/observiq-otel-collector/tree/main/exporter/googlecloudexporter

config/google_cloud_exporter/iis/config.yaml
jsirianni: See Brandon's comment

cpheps (approved):

BinaryFissionGames (approved):

jsirianni (approved):

PR #420, opened 2022-05-17
feat: Opamp Client Connection

cpheps:
### Proposed Change
- Added branching logic in collector startup to load into a managed opamp connection if a `manager.yml` is present
- Added a `logging.yaml` flag (required for opamp) even though not needed right now.
- implemented iris (observIQ) specific OpAmp connection protocol

**NOTE**: This is still raw implementation. I have some follow on tasks that will cleanup/change some of the code flow. There is work coming for hot reloading configs to avoid a complete shutdown of the collector. This will likely remove the `Validator` style code in favor of a hot reload test.

##### Checklist
- [X] Changes are tested
- [ ] CI has passed


internal/os/os.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Hostname retrieves the hostname of the machine
+func Hostname() (string, error) {
+	info, err := host.Info()
+	if err != nil {
+		return unknownValue, err
+	}
+
+	return info.Hostname, nil
+}
```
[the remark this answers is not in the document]
cpheps: Yeah you're right. I considered making this a singleton and only have host.Info() called once and reused. I think that got lost in the haste of getting this done. That being said this is only called once per the life of the program so I don't feel like it's a concern if this is doing extra to get a single value.

I'll see if I can factor it into a singleton. This package was copied from agent so I didn't have to reinvent the wheel.

cmd/collector/main.go
```diff
@@ -46,12 +58,33 @@ func main() {
...
+		log.Println("Starting Management Path")
+		// Management config exists
+		configManger := observiq.NewAgentConfigManager(defaultLogger.Sugar())
+		configManger.AddConfig(opamp.CollectorConfigName, (*configPaths)[0], opamp.NewYamlValidator(make(map[string]interface{})))
+		configManger.AddConfig(opamp.ManagerConfigName, *managerConfigPath, opamp.NewYamlValidator(new(opamp.Config)))
+		configManger.AddConfig(opamp.LoggingConfigName, *loggingConfigPath, opamp.NoopValidator)
+
+		currCollector := collector.New((*configPaths)[0], version.Version(), []zap.Option{})
+
+		if err := runRemoteManaged(ctx, currCollector, configManger, *managerConfigPath); err != nil {
```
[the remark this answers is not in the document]
cpheps: We should talk about this I don't quite understand what you mean by not having a service handler.

opamp/observiq/identity.go
```diff
@@ -0,0 +1,99 @@
...
+
+	name, err := ios.Name()
+	if err != nil {
+		logger.Warn("Failed to retrieve host details on collector. Creating partial identity", zap.Error(err))
+	}
+
+	return &identity{
+		agentID:     config.AgentID,
+		agentName:   config.AgentName,
+		serviceName: "com.observiq.collector", // Hardcoded defines this type of agent to the server
```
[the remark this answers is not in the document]
cpheps: That's correct. The name `serviceName` is misleading but that's what iris already calls it so I was keeping backwards compatibility.

internal/os/os.go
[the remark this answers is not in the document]
cpheps: @BinaryFissionGames Refactored into rudimentary singleton pattern

opamp/client.go
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package opamp
+
+import "context"
+
+// Client implements a connection with OpAmp enabled server
+type Client interface {
```
StefanKurek, line 20: maybe this question doesn't make sense because i haven't seen all the code yet, but is there no interface that otel provides for you to use for a client?

cmd/collector/main.go
```diff
@@ -16,20 +16,32 @@ package main
...
+	"go.uber.org/zap/zapcore"
 )
 
+var defaultLogger *zap.Logger
+
 func main() {
-	var configPaths = pflag.StringSlice("config", []string{"./config.yaml"}, "the collector config path")
+	setupDefaultLogger()
+
+	configPaths := pflag.StringSlice("config", []string{"./config.yaml"}, "the collector config path")
```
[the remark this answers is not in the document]
cpheps: The otel collector can technically take in multiple configs. This maybe something we factor out in a future release but for now we're passing it straight through.

opamp/config.go
```diff
@@ -57,3 +60,36 @@ func ParseConfig(configLocation string) (*Config, error) {
...
+	}
+
+	return cmpStringPtr(c.Labels, o.Labels)
+}
+
+func cmpStringPtr(p1, p2 *string) bool {
+	switch {
+	case p1 == nil && p2 == nil:
+		return true
+	case p1 == nil && p2 != nil:
```
StefanKurek, line 86: maybe something is happening in go that i don't understand here, but are these "fallthrough" cases necessary?

opamp/config.go
```diff
@@ -57,3 +60,36 @@ func ParseConfig(configLocation string) (*Config, error) {
...
+func (c Config) GetSecretKey() string {
+	if c.SecretKey == nil {
+		return ""
+	}
+
+	return *c.SecretKey
+}
+
+// CmpUpdatableFields compares updatable fields for equality
+func (c Config) CmpUpdatableFields(o Config) (equal bool) {
```
StefanKurek, line 74: extremely minor and i wouldn't hold up a pr for it, but is cmp a standard abbreviation for compare? maybe it is? all i know is i didn't know what this was doing until i looked at the code/read the description

opamp/client.go
cpheps: There is but since we're wrapping the otel client in our own flavor of it I wanted to have a generic contract for anyone interacting with our opamp package. So the observiq flavor of that client implements the interface. We could have a splunk or newrelic version that also implements that same interface and wraps the opamp client with logic specific to those platforms.

StefanKurek: gotcha

opamp/config.go
cpheps: So switch statements in go don't fallthrough by default like most languages do. You have to tell it to do so. Since all cases return false I added fallthrough. I could also just put `return false` in every case.

cmd/collector/main.go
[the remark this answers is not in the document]
cpheps: Yes in the managed mode we are. It's a bit of a hold over and I wanted to not mess with standalone mode code as much as possible. I think we need to revisit how we do cli and flags overall so we should clean it up then. I'm fine accepting this as tech debt.

opamp/config.go
StefanKurek: must just be my brain misfiring 😄

opamp/config.go
StefanKurek: @cpheps i mean what would happen if you completely removed those to 1 thing == nil cases

cpheps: Not sure I follow what what you're asking.

opamp/observiq/observiq_client.go
```diff
@@ -0,0 +1,192 @@
...
+	Config        opamp.Config
+	Collector     collector.Collector
+
+	ManagerConfigPath   string
+	CollectorConfigPath string
+	LoggerConfigPath    string
+}
+
+// NewClient creates a new OpAmp client
+// The passed in configmanager should be preloaded with all known configs.
```
StefanKurek, line 65: this comment is a little confusing. i don't see a "passed in" configmanager?

opamp/config.go
StefanKurek: cleared this up. just didn't understand exactly how fallthrough works in go

opamp/managed_config.go
```diff
@@ -0,0 +1,32 @@
...
+
+// ReloadFunc is a function that handles reloading a config given the new contents
+type ReloadFunc func([]byte) (changed bool, err error)
+
+// NoopReloadFunc used as a noop reload function if unsure of how to reload
+func NoopReloadFunc([]byte) (bool, error) {
+	return false, nil
+}
+
+// ManagedConfig is a structure that can manager an on disk config file
```
StefanKurek, line 25: minor: typo "manage"

opamp/observiq/observiq_client.go
```diff
@@ -0,0 +1,190 @@
...
+		Reload:     loggerReload(observiqClient, args.LoggerConfigPath),
+	})
+
+	return observiqClient, nil
+}
+
+// Connect initiates a connection to the OpAmp server
+func (c *Client) Connect(ctx context.Context) error {
+	// Compose and set the agent description
+	if err := c.opampClient.SetAgentDescription(c.ident.ToAgentDescription()); err != nil {
```
StefanKurek, line 115: @cpheps is this here because configs could change now during the opamp lifecycle of the collector?

opamp/observiq/observiq_client.go
```diff
@@ -0,0 +1,190 @@
...
+	c.collector.Stop()
+	return c.opampClient.Stop(ctx)
+}
+
+// client callbacks
+
+func (c *Client) onConnectHandler() {
+	c.logger.Info("Successfully connected to server")
+
+	if err := c.opampClient.SetAgentDescription(c.ident.ToAgentDescription()); err != nil {
```
StefanKurek, line 162: @cpheps is this here because configs could change now during the opamp lifecycle of the collector?

cpheps: actually that's a good point we don't need to set this in the onConnect as we do it in the Connection function. I'll remove it. You only need to `SetAgentDescription` when it changes. So we only do that when we first start and on any changes to the manager.yml

cpheps: Just pushed an update that removed this

StefanKurek (approved):
I have not tested it, but it seems like it overall makes sense

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,167 @@
...
+func (a *AgentConfigManager) AddConfig(configName string, managedConfig *opamp.ManagedConfig) {
+	a.configMap[configName] = managedConfig
+}
+
+// ComposeEffectiveConfig reads in all config files and calculates the effective config
+func (a *AgentConfigManager) ComposeEffectiveConfig() (*protobufs.EffectiveConfig, error) {
+	contentMap := make(map[string]*protobufs.AgentConfigFile, len(a.configMap))
+
+	// Used to track config names to alphabetize later in hash compution
+	configNames := make([]string, 0, len(a.configMap))
```
StefanKurek, line 72: is this actually used or can we ditch this?

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// ApplyConfigChanges compares the remoteConfig to the existing and applies changes
+func (a *AgentConfigManager) ApplyConfigChanges(remoteConfig *protobufs.AgentRemoteConfig) (effectiveConfig *protobufs.EffectiveConfig, changed bool, err error) {
+	effectiveConfig, err = a.ComposeEffectiveConfig()
+	if err != nil {
+		return nil, false, err
+	}
+
+	currentConfigMap := effectiveConfig.GetConfigMap().GetConfigMap()
```
StefanKurek, line 105: since this works, i'm assuming this isn't an error (double GetConfigMap())?

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,167 @@
...
+	// CollectorConfigName is the key of the collector config in OpAmp
+	CollectorConfigName = "collector.yaml"
+	// ManagerConfigName is the key of the manager config in OpAmp
+	ManagerConfigName = "manager.yaml"
+	// LoggingConfigName is the key of the logging config in OpAmp
+	LoggingConfigName = "logging.yaml"
+)
+
+// acceptableConfigs is a lookup of configs that are able to be written/updated
+var acceptableConfigs = map[string]struct{}{
```
StefanKurek, line 38: won't hold up the PR if the answer is yes (as long as we make a tech debt bug), but should acceptableConfigs be used in the other two public methods as well?

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,167 @@
...
+	// No remote config Map
+	if remoteConfigMap == nil {
+		return
+	}
+
+	// loop through all remote configs and compare then with existing configs
+	for configName, remoteContents := range remoteConfigMap {
+		// For security check the log file we want is acceptable
+		if _, ok := acceptableConfigs[configName]; !ok {
+			a.logger.Info("Not support config received skipping", zap.String("config", configName))
```
StefanKurek, line 117: minor, feels like this could be a warn?

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,167 @@
...
+
+	return &protobufs.EffectiveConfig{
+		ConfigMap: &protobufs.AgentConfigMap{
+			ConfigMap: contentMap,
+		},
+	}, nil
+}
+
+// ApplyConfigChanges compares the remoteConfig to the existing and applies changes
+func (a *AgentConfigManager) ApplyConfigChanges(remoteConfig *protobufs.AgentRemoteConfig) (effectiveConfig *protobufs.EffectiveConfig, changed bool, err error) {
```
StefanKurek, line 99: minor: this method's a little large. would be nice if we could refactor to break it up a bit

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,167 @@
...
+
+			// Write out the file
+			if err := os.WriteFile(configName, remoteContents.GetBody(), 0600); err != nil {
+				return nil, false, fmt.Errorf("failed to write new config file %s: %w", configName, err)
+			}
+
+			// Track new config
+			a.AddConfig(configName, &opamp.ManagedConfig{
+				ConfigPath: filepath.Join(".", configName),
+				Reload:     opamp.NoopReloadFunc,
```
StefanKurek, line 135: are there any issues with this new config not having a reload func?

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+
+		// Update the config file
+		managedConfig := a.configMap[configName]
+		newContents := remoteContents.GetBody()
+
+		a.logger.Info("Applying changes to config file", zap.String("config", configName))
+		configChanged, err := managedConfig.Reload(newContents)
+		if err != nil {
+			return effectiveConfig, false, fmt.Errorf("failed to reload config: %s: %w", configName, err)
```
StefanKurek, line 152: why return effectiveConfig here vs returning nil when failing to write a brand new config? what's the reasoning?

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,167 @@
...
+		if bytes.Equal(currentContents.GetBody(), remoteContents.GetBody()) {
+			continue
+		}
+
+		// Update the config file
+		managedConfig := a.configMap[configName]
+		newContents := remoteContents.GetBody()
+
+		a.logger.Info("Applying changes to config file", zap.String("config", configName))
+		configChanged, err := managedConfig.Reload(newContents)
```
StefanKurek, line 150: would there ever be a case where there was new config data, but this returned NO config change and NO error? in this case we'd just quietly move on and the end result of this function could be no change no error and not even a log message explaining why the heck the config didn't change when there was in fact a remote change.

does that make sense?

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// If files have changed recompute effective config
+	if changed {
+		effectiveConfig, err = a.ComposeEffectiveConfig()
+		if err != nil {
+			return nil, true, err
+		}
+	}
+
+	return
```
StefanKurek, line 166: is this some weird go thing where we don't have to return `effectiveConfig, changed, nil`?

StefanKurek: forgot about named return values. nevermind

opamp/observiq/observiq_client.go
```diff
@@ -0,0 +1,186 @@
...
+func (c *Client) onConnectFailedHandler(err error) {
+	c.logger.Error("Failed to connect to server", zap.Error(err))
+}
+
+func (c *Client) onErrorHandler(errResp *protobufs.ServerErrorResponse) {
+	c.logger.Error("Server returned an error response", zap.String("Error", errResp.GetErrorMessage()))
+}
+
+func (c *Client) onRemoteConfigHandler(_ context.Context, remoteConfig *protobufs.AgentRemoteConfig) (*protobufs.EffectiveConfig, bool, error) {
+	c.logger.Debug("Remote config handler")
```
StefanKurek, line 172: minor. seems like this could be info level

opamp/observiq/observiq_client.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (c *Client) onConnectFailedHandler(err error) {
+	c.logger.Error("Failed to connect to server", zap.Error(err))
+}
+
+func (c *Client) onErrorHandler(errResp *protobufs.ServerErrorResponse) {
+	c.logger.Error("Server returned an error response", zap.String("Error", errResp.GetErrorMessage()))
+}
+
+func (c *Client) onRemoteConfigHandler(_ context.Context, remoteConfig *protobufs.AgentRemoteConfig) (*protobufs.EffectiveConfig, bool, error) {
```
StefanKurek, line 171: will we still have to set a new agent description here if a new remote manager config came in?

cmd/collector/main.go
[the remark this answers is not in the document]
cpheps: This is being address in #433 and will be merged into here to fix windows service.

opamp/observiq/agent_config_manager.go
cpheps: Nope artifact of a previous implementation. Good catch.

opamp/observiq/agent_config_manager.go
cpheps: No it's generated proto code. The proto structure for EffectiveConfig has a field called [ConfigMap](https://github.com/open-telemetry/opamp-go/blob/888d3adc29d859534099b0807b7d9cd8bafd772e/protobufs/opamp.pb.go#L1966) which also has a field called [ConfigMap](https://github.com/open-telemetry/opamp-go/blob/888d3adc29d859534099b0807b7d9cd8bafd772e/protobufs/opamp.pb.go#L2428). Those getter functions do nil checks so if anything at all in that chain is nil it won't panic and just return nil at the end.

opamp/observiq/agent_config_manager.go
cpheps: `acceptableConfigs` is used as a quick reference/security check to verify the config the server is sending down to update/add is one of the supported ones. We only use this for that check. It doesn't matter to use if you add more than these to be managed but they won't be able to be remote managed unless added to this list.

opamp/observiq/agent_config_manager.go
cpheps: I'll factor out the new file addition. I'd like to keep the actual config change logic inline though.

opamp/observiq/agent_config_manager.go
cpheps: This is more of a place holder for now. We don't know anything about this file so it's hard to say how we should reload it. Maybe at some point we add a generic just overwrite file type thing. This also won't make it into the `acceptableConfigs` map either. I'd like to leave this as is right now for security reasons until we know more about how this could be used or exploited.

opamp/observiq/agent_config_manager.go
cpheps: I actually updated another error spot to return it. The logic is that's the last config we computed on disk. If something goes wrong we can still hand the server our last good config and it can try again. I think the opamp library ignores it if we return an error in this function but I'd rather send it incase.

opamp/observiq/agent_config_manager.go
cpheps: I'm not sure I follow. If there's new data we trigger the reload function. If a reload fails we submit and error. We're trusting the reload function to do its job. If it decides there's no changes then there's nothing to report. The spec is pretty open ended on this. It even calls out the agent doesn't have to take any configuration changes from the platform.

I get what you're saying but not sure I see a case where that could happen.

opamp/observiq/observiq_client.go
cpheps: This could go off a lot I don't want to spam the logs with it. It's more just for us to know what's happening if things are acting weird.

opamp/observiq/observiq_client.go
cpheps: That happens if the manager.yaml is reloaded. It happens in the `managerReload` function.

opamp/observiq/agent_config_manager.go
StefanKurek: @cpheps practically though, will we even be able to run with unacceptable configs right now (if running in managed mode)?

or another question. what happens if we have an unacceptable config which gets sent back to iris. what does that mean?

opamp/observiq/agent_config_manager.go
StefanKurek: looks better. i'd still probably factor out the stuff in the for loop into a method like `compareRemoteConfig` or something as the method still seems a little large but i won't nitpick on it

opamp/observiq/agent_config_manager.go
StefanKurek: @cpheps i'm just saying that this method doesn't have control over how the managedConfig.Reload method is implemented. but i guess if we're saying this is a false condition that should never be implemented that way, then sure nothing necessarily needs to be done here

opamp/observiq/reload_funcs.go
```diff
@@ -0,0 +1,216 @@
...
+
+		// Read in existing config file
+		currConfig, err := opamp.ParseConfig(managerConfigPath)
+		if err != nil {
+			return false, err
+		}
+
+		// Check if the updatable fields are equal
+		// If so then exit
+		if currConfig.CmpUpdatableFields(newConfig) {
```
StefanKurek, line 44: @cpheps should we be comparing to what is currently in the manager.yml file or what is currently stored in the agent config manager?

what if the user had changed the manager.yml and set some new values. this does nothing right? then they change those same values in iris to be sent down. now when this comparison happens, they are equal and no changes actually occur right?

opamp/observiq/agent_config_manager.go
cpheps: Talked this through on slack. I don't think this is a current concern but could be in the future.

opamp/observiq/reload_funcs.go
cpheps: Yes this opens up an issue where if the on-disk copy matches iris but the in memory doesn't it'll never be updated. I'll look into adding in memory protection.

opamp/observiq/observiq_client.go
cpheps: Talked this through on slack.

cmd/collector/main.go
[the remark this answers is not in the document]
cpheps: Added in latest commits

cpheps:
Added logic to wrap in a `RunnableService`. Also, added logic to check in case memory matches the platform but disk doesn't.

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		err = fmt.Errorf("failed to reload config: %s: %w", configName, err)
+		return
+	}
+
+	// If the config changed recompute the hash for it
+	if changed {
+		err = managedConfig.ComputeConfigHash()
+		if err != nil {
+			err = fmt.Errorf("failed hash compute for config %s: %w", configName, err)
```
StefanKurek, line 166: this seems like a pretty bad case. like we've updated our config info in memory, but something is jacked where we couldn't calculate the hash (missing file on disk?).

actually thinking this through maybe it's not a problem. cause if there was an old hash that didn't match the current config (because computing the new hash failed), then the next time iris requests aggregated data we won't give it the missing file data and it'll know there's a problem and send another config maybe which hopefully we succeed next time.

but if the error was purely with calculating the hash and it was temporary (i don't know why this would be the case), that might be more problematic. like if the new file was written, but the hash wasn't computed and the stored one is old. could this get funky? do we need a comment here or a backlog tech debt to look into this?

opamp/observiq/observiq_client.go
```diff
@@ -0,0 +1,195 @@
...
+
+// Client represents a client that is connected to Iris via OpAmp
+type Client struct {
+	opampClient   client.OpAMPClient
+	logger        *zap.SugaredLogger
+	ident         *identity
+	configManager opamp.ConfigManager
+	collector     collector.Collector
+
+	currentConfig opamp.Config
```
StefanKurek, line 48: just off the bat wondering why the config from the configManager couldn't be used in place of this?

StefanKurek: nevermind, i guess it would be a pain to have to parse it into the struct every time you wanted to reference it

cpheps: Yeah, config manager is meant to manage the file and contents on disk. I'm added this reference mostly for hot reloading.

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,189 @@
...
+// NewAgentConfigManager creates a new AgentConfigManager
+func NewAgentConfigManager(defaultLogger *zap.SugaredLogger) *AgentConfigManager {
+	return &AgentConfigManager{
+		configMap: make(map[string]*opamp.ManagedConfig),
+		logger:    defaultLogger.Named("config manager"),
+	}
+}
+
+// AddConfig adds a config to be tracked by the config manager.
+// It will also compute the current hash of the config
```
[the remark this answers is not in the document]
cpheps: Yeah I was going to have it do so then I came up with a better design for got to remove it.

opamp/observiq/agent_config_manager.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// Nothing to update
+	if bytes.Equal(managedConfig.GetCurrentConfigHash(), remoteHash) {
+		// Verify disk matches memory and remote. This will ensure if any changes were made to the config on disk
+		// the will be overwritten by what is expected.
+		return verifyDiskContents(managedConfig.ConfigPath, managedConfig.GetCurrentConfigHash(), newContents)
+	}
+
+	a.logger.Info("Applying changes to config file", zap.String("config", configName))
+	changed, err = managedConfig.Reload(newContents)
```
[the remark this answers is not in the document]
cpheps: Yeah I kept the word changed since that's what it is in the opamp library. They call it `configChanged` so I wanted to keep the terminology.

>it doesn't actually represent that anything was changed. just that the file has properly been overwritten with the remote config data

This is not always true and it depends on the reload function. All of our reload functions do a hot reload so in the case of the manager it changes the identity of the agent and for the logger it changes the logger object. 

I think adding a comment though to the main struct is not a bad idea just to make it clear.

opamp/observiq/agent_config_manager.go
cpheps: So the hash function fails if we fail to open the file to read. Which means we don't recompute the hash. It does put things in a weird state where likely you are running with the correct config and have the correct one on disk at this point. So on the next apply remote config it'll show a mismatch and apply the config again.

Due to the amount of file ops the entire reload logic entails there are several errors points. It's more likely though to have all of them fail rather than one randomly unless the system is running out of file descriptors. I think this is an acceptable risk and the case where we're wrong seems like it would be self correcting eventually.

opamp/managed_config.go
```diff
@@ -0,0 +1,72 @@
...
+	// currentConfigHash is the hash of the config currently being used
+	currentConfigHash []byte
+}
+
+// GetCurrentConfigHash retrieves the current config hash
+func (m *ManagedConfig) GetCurrentConfigHash() []byte {
+	return m.currentConfigHash
+}
+
+// ComputeConfigHash reads in the config file and computes the hash for it saving it on the ManagedConfig
```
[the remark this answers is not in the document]
cpheps: I think I'm going to keep the function name but updated the comment.

opamp/observiq/reload_funcs.go
```diff
@@ -0,0 +1,262 @@
...
+			if rollbackErr := rollbackFunc(); rollbackErr != nil {
+				client.logger.Error("Rollback failed for manager config", zap.Error(rollbackErr))
+			}
+			client.ident = rollbackIdent
+			client.currentConfig = *rollBackCfg
+			return false, fmt.Errorf("failed to reformat manager config: %w", err)
+		}
+
+		// Save config file to disk
+		if err := updateConfigFile(ManagerConfigName, managerConfigPath, newContents); err != nil {
```
StefanKurek, line 82: just a thought. so we could have files on disk that have new "non updatable fields" right? if a weird config came in and "everything" was different (including agent name/labels), we'd only actually update the identity's name/labels, but the file would contain new out of sync info for the stuff that can't be changed on the fly.

this practically doesn't matter while the code continues executing, but if the program is every stopped and reran, then it WILL pick up ALL the completely new info in the managed config.yaml.

i have no idea if this is even a problem, just noticing it for the first time and making sure it is/isn't a problem

opamp/observiq/agent_config_manager.go
StefanKurek: @cpheps yeah, i think no matter what this ends up self correcting. i'll resolve

opamp/observiq/reload_funcs.go
```diff
@@ -0,0 +1,262 @@
...
+	return func(contents []byte) (bool, error) {
+		// Unmarshal config and only pull fields out that are allowed to be updated.
+		var newConfig opamp.Config
+		if err := yaml.Unmarshal(contents, &newConfig); err != nil {
+			return false, fmt.Errorf("failed to validate config %s", ManagerConfigName)
+		}
+
+		// Check if the updatable fields are equal
+		// If so then exit
+		if client.currentConfig.CmpUpdatableFields(newConfig) {
```
StefanKurek, line 39: also, because of this, if we have different hashes but the correct config we will hit this over and over again never correctly. in this case would it be better to only overwrite the file but not mess with hot reloading?

opamp/observiq/reload_funcs.go
cpheps: That's a good call. It totally negates the security concern.

opamp/observiq/reload_funcs.go
cpheps: Every time we have have an incorrect hash we will reload, giving it another chance to correctly compute. If it's unlikley that we are able to fully do a reload then fail the hash since there are so many file ops present in the reload.

opamp/observiq/observiq_client.go
```diff
@@ -0,0 +1,195 @@
...
+	ErrUnsupportedURL = errors.New("unsupported URL")
+)
+
+// Ensure interface is satisfied
+var _ opamp.Client = (*Client)(nil)
+
+// Client represents a client that is connected to Iris via OpAmp
+type Client struct {
+	opampClient   client.OpAMPClient
+	logger        *zap.SugaredLogger
```
BinaryFissionGames, line 43: Why is this logger sugar'ed? It looks like we're always using the non-sugared, field based methods anyways.

opamp/observiq/reload_funcs.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+		}()
+
+		// Write new config file
+		if err := updateConfigFile(CollectorConfigName, collectorConfigPath, contents); err != nil {
+			return false, err
+		}
+
+		// Reload collector
+		if err := client.collector.Restart(context.Background()); err != nil {
```
BinaryFissionGames, line 127: Do we have any concerns about losing the original context for the collector here? It seems like it works fine for now, but I'm wondering if it could cause problems down the line.

opamp/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+		Value: &protobufs.AnyValue{
+			Value: &protobufs.AnyValue_StringValue{StringValue: value},
+		},
+	}
+}
+
+// ComputeHash computes a sha256 hash of the passed in data
+func ComputeHash(data []byte) []byte {
+	return sha256.New().Sum(data)
+}
```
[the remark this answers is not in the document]
cpheps: Oh dang good call I'll fix that.

We use sha256 because that's the way opamp hashes things. Theres no reason I just figured we should be consistent. It's purely for comparing hashes.

opamp/observiq/observiq_client.go
cpheps: I think this is a hold over from the agent code I copied. I'll factor it out.

opamp/observiq/reload_funcs.go
cpheps: There's a small concern. We call collector stop on a signal regardless. The collector just don't shut itself down on a signal now. I think it's fine. The alternative is to keep a reference to the signal context on the client and pass that through. I didn't like that but we could do it.

opamp/observiq/observiq_client.go
cpheps: Ohh looks like the opamp library logging interface isn't fulfilled by a non-sugared logger. I can just sugar it when it's passed into the implementation

opamp/observiq/reload_funcs.go
StefanKurek: yep, i gotcha. this won't be an issue because the collector only runs a manager config file on startup through a flag.

opamp/helpers.go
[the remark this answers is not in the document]
BinaryFissionGames: Cool, if it's for consistency that sounds like a good enough reason.

I ask because FNV is less expensive than SHA2, so I tend towards that for non-cryptographic purposes, but I don't think this is really a hot path, so it doesn't matter much here.

opamp/observiq/reload_funcs.go
BinaryFissionGames: Yeah, I think it's probably fine too, just oddly inconsistent between the first start of the collector and subsequent starts.

IMO, we should either do what you suggested (keep the ctx on the client and pass it in here), or pass in the background context on the initial startup; Either way, I think we should try to keep it consistent.

opamp/helpers.go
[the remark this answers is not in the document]
cpheps: Yeah this path involves writing and reading from disk so it's not fast at all

opamp/observiq/reload_funcs.go
cpheps: Yeah I agree. I think we put the background context initially. That way it's clear the our program needs to shut down the collector.

BinaryFissionGames (approved):
This LGTM.

opamp/observiq/reload_funcs.go
StefanKurek: talked about this offline. not a major concern and no need to change anything

PR #422, opened 2022-05-17
feat: Add CSV plugin

dmikolay:
### Proposed Change
Added a CSV plugin.

##### Checklist
- [X] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/csv_logs.yaml
+++ b/plugins/csv_logs.yaml
@@ -30,12 +30,6 @@ parameters:
   - name: log_type
     type: string
     default: csv
-  - name: start_at
-    type: string
-    supported:
-    - beginning
-    - end
-    default: end
 template: |
   receivers:
     filelog:
@@ -50,9 +44,7 @@ template: |
         - '{{ $fp }}'
         {{end}}
       {{end}}
-      {{ if .encoding }}
       encoding: '{{ .encoding }}'
-      {{end}}
       attributes:
         log_type: '{{ .log_type }}'
       operators:
--- a/plugins/csv_logs.yaml
+++ b/plugins/csv_logs.yaml
@@ -1,0 +1,64 @@
+version: 0.0.1
+title: CSV
+description: Log parser for CSV
+parameters:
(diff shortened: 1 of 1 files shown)


plugins/csv_logs.yaml
```diff
@@ -0,0 +1,64 @@
...
+    default: utf-8
+  - name: log_type
+    type: string
+    default: csv
+  - name: start_at
+    type: string
+    supported:
+    - beginning
+    - end
+    default: end
```
cpheps, line 38: `start_at` is in here twice. It looks like it is in the original too.

cpheps (commented):
Overall looks good, did you test this against any log files to verify?

plugins/csv_logs.yaml
```diff
@@ -0,0 +1,64 @@
...
+      start_at: {{ .start_at }}
+      {{ if .exclude_file_log_path }}
+      exclude:
+        {{ range $fp := .exclude_log_paths }}
+        - '{{ $fp }}'
+        {{end}}
+      {{end}}
+      {{ if .encoding }}
+      encoding: '{{ .encoding }}'
+      {{end}}
```
jsirianni, line 55: Encoding has a default, no need to wrap with an `if`.

```suggestion
      encoding: '{{ .encoding }}'
```

plugins/csv_logs.yaml
dmikolay: I just removed one of the repeated start-at. I redid some testing and it appear to still work just fine.

plugins/csv_logs.yaml
dmikolay: Just removed the if statement wrapping encoding, thanks!

plugins/csv_logs.yaml
dmikolay: Also, I did test against some log files, it seemed to function correctly.

cpheps (approved):

StefanKurek (approved):

PR #433, opened 2022-05-20
feat: Add logging config through logging.yaml

BinaryFissionGames:
### Proposed Change
* Create a service handler for windows
  * This was necessary because the OpenTelemetry one ALWAYS overwrote the core with a windows event logging one
  * This will also help with OpAmp integration, which will need to use this code to run correctly on Windows as a service
* Allow configuration through logging.yaml
  * Can be specified through the `--logging` flag
  * If not specified, stdout is logged to by default
  * If the logging flag is specified but the yaml file cannot be found, stdout will be logged to.
  * The filepath in the logging.yaml may use variable substitution.
  * Default install now logs to
    * /opt/observiq-otel-collector/log/collector.log on Linux
    * C:\Program Files\observIQ OpenTelemetry Collector\log\collector.log on Windows
    * $(brew --prefix observiq/observiq-otel-collector/observiq-otel-collector)/log/collector.log on macOS
* Remove windows event logging
  * We could potentially add this back in as another output at a later date.

I've tested Windows + Linux installs, and confirmed they both log to files by default now.

I'll need help testing the brew locally, I haven't done that before.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cmd/collector/main.go
```diff
@@ -41,28 +41,33 @@ func main() {
...
+	// logOpts will override options here
+	logger, err := zap.NewProduction(logOpts...)
 	if err != nil {
-		log.Fatal(err)
+		log.Fatalf("Failed to set up logger: %v", err)
 	}
 
-	if err := run(ctx, *settings); err != nil {
-		log.Fatal(err)
+	col := collector.New((*configPaths)[0], version.Version(), logOpts)
```
cpheps, line 55: The collector use to be able to take in multiple config paths through creating a settings. Looks like we've lost that functionality here?

internal/logging/config.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"gopkg.in/natefinch/lumberjack.v2"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	// FileOutput is an output option for logging to a file.
+	FileOutput string = "file"
```
cpheps, line 30: This isn't used outside of this package it looks like so it could be unexported. Same with `StdOutput` below.

.gitignore
```diff
@@ -14,9 +14,12 @@ gpg.*
...
 windows/config.yaml
+windows/logging.yaml
 windows/LICENSE
 windows/plugins
 opentelemetry-java-contrib-jmx-metrics.jar
 VERSION.txt
 release_deps
 tmp
+go.work
+go.work.sum
```
cpheps, line 25: We shouldn't have these in the .gitignore. We don't support this go1.18 feature but we might want to in the future.

.goreleaser.yml
```diff
@@ -290,11 +300,19 @@ brews:
...
+            <string>#{prefix}</string>
+          </dict>
+
           <key>ProgramArguments</key>
           <array>
             <string>#{bin}/observiq-otel-collector</string>
             <string>--config</string>
             <string>#{prefix}/config.yaml</string>
+            <string>--logging</string>
+            <string>#{prefix}/logging.yaml</string>
```
cpheps, line 315: This looks good. Should we remove the stdout and stderr redirect that is happening at the bottom of this plist file?

internal/logging/config.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// If the file doesn't exist, we just return the default config.
+	if _, err := os.Stat(configPath); errors.Is(err, os.ErrNotExist) {
+		return conf, nil
+	} else if err != nil {
+		return nil, err
+	}
+
+	confBytes, err := os.ReadFile(configPath)
```
cpheps, line 62: Gosec is mad here. You need to do a `filepath.Clean` on the path to satisfy gosec.

internal/service/mocks/collector.go
```diff
@@ -0,0 +1,77 @@
+// Code generated by mockery v2.12.2. DO NOT EDIT.
```
cpheps, line 1: This Collector isn't part of the service package. Seems like it should be in its own `mocks` directory under the `collector` package. I use this in opamp so having it closer to the interface it implements seems more intuitive for anyone wanting to use it.

internal/service/service.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+}
+
+// Start starts the collector
+func (s StandaloneCollectorService) Start(ctx context.Context) error {
+	collectorStartedChan := make(chan struct{})
+	go func() {
+		// Start's context is only valid for the lifetime of "Start",
+		// but the collector expects a context that is valid for the lifetime of the service.
+		err := s.col.Run(context.Background())
```
cpheps, line 69: The comment on the context here confuses me. Why is `Start`'s context only good for the lifetime of start? How do we pass in a context that is listening for signals?

cmd/collector/main.go
cpheps: We also had to set the set the `DisableGracefulShutdown` on the collector or else it would take over signal intercept from our process.

internal/service/service.go
```diff
@@ -0,0 +1,159 @@
...
+				s.errChan <- status.Err
+			} else if !status.Running {
+				// If we aren't running, bail out. Otherwise the collector is effectively a "zombie" process.
+				s.errChan <- errors.New("collector unexpectedly stopped running")
+			}
+		case <-s.doneChan:
+			return
+		}
+	}
+}
```
cpheps, line 106: This might not be needed anymore. I fixed the collector in OTel a while back so that Run will return an error if anything is wrong during startup. It'll also set the status to `Closed` if startup fails instead of leaving it as `Starting` like previously.

cpheps: So ideally you just kick off `col.Run` in a goroutine and if it returns an error you notify the caller.

internal/service/service_others.go
```diff
@@ -13,16 +13,22 @@
...
-	"go.opentelemetry.io/collector/service"
+	"go.uber.org/zap"
 )
 
-func run(ctx context.Context, params service.CollectorSettings) error {
-	return runInteractive(ctx, params)
+// RunService runs the given service, calling its start and stop functions.
+func RunService(logger *zap.Logger, rSvc RunnableService) error {
+	stopSignal := make(chan os.Signal, 1)
+	signal.Notify(stopSignal, syscall.SIGINT, syscall.SIGTERM)
```
cpheps, line 30: Curious why we didn't want to use the signal.NotifyContext but rather use a channel?

internal/service/service.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil
+	case <-ctx.Done():
+		return fmt.Errorf("failed while waiting for service shutdown: %w", ctx.Err())
+	}
+}
+
+// runServiceInteractive runs the service in an "interactive" mode (responds to SIGINT and SIGTERM).
+// This mode is always used in linux, and is used in Windows when the collector
+// is not running as a service.
+func runServiceInteractive(logger *zap.Logger, stopSignal <-chan os.Signal, svc RunnableService) error {
```
cpheps, line 135: This should probably take in a context that is listening for signals and use that as the base for the `startupTimeoutCtx`. That way if a signal comes in or we timeout the context will end.

scripts/package/postinstall.sh
```diff
@@ -70,7 +71,7 @@ The collector's config file can be found here:
   /opt/observiq-otel-collector/config.yaml
 
 To view logs from the collector, run:
-  sudo journalctl -u observiq-otel-collector.service
+  sudo tail -F /opt/observiq-otel-collector/log/collector.log
```
cpheps, line 74: Do we need to update the mac or windows docs to reflect this also?

cmd/collector/main.go
BinaryFissionGames: `DisableGracefulShutdown` is set (col.Run calls collector.NewSettings, which sets this option)

As far as multiple configs, good point, I think I had it in my head that we were only using the first config previously, but we were using multiple. I'll change collector.New to accept the full slice.

internal/service/service.go
BinaryFissionGames: The context for start contains timeout information (the Start operation should be quick and not long-lasting, I want to bail out and restart if it hangs).

Maybe it should be long running, though. I think the timeout could be handled outside this function.

Either way, Shutdown WILL be called when a signal is caught, which I think we need to do anyways to ensure the collector is fully shut down before exiting the program.

cpheps: I saw this later on but I still think we can use the signal context as a parent. See https://github.com/observIQ/observiq-otel-collector/pull/433#discussion_r878097759

internal/service/service.go
BinaryFissionGames: Hmm, you might be right. I think the reason why I have this is context cancellation, actually. If the context passed into s.col.Run is cancelled unexpectedly, no error is returned, and the process becomes a zombie. 

I put this as a failsafe if there were any similar scenarios where the underlying collector might stop but not report an error.

cpheps: Might be something to talk through. This logic was originally a big protection as the collector would die and still report it's attempting to start.

StefanKurek (approved):
Given that I'm not super familiar with the problem being solved and that this PR is my first real exposure to channels/goroutines, this LGTM! A few minor comments/questions, but nothing that I'd hold the PR up over.

internal/service/service_others.go
BinaryFissionGames: Good point, changed this to a context.

internal/service/service.go
BinaryFissionGames: I kept it with the short running ctx (signal context becomes the parent). It becomes overcomplicated and doesn't really seem necessary, since Shutdown is called. Let me know if there's strong disagreement with that.

internal/logging/config.go
```diff
@@ -0,0 +1,110 @@
...
+
+// NewLoggerConfig returns a logger config. If configPath is not
+// set, stdout logging will be enabled.
+func NewLoggerConfig(configPath string) (*LoggerConfig, error) {
+	conf := &LoggerConfig{
+		Output: StdOutput,
+		Level:  zapcore.InfoLevel,
+	}
+
+	if configPath == "" {
```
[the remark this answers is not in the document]
BinaryFissionGames: Actually, I think this is because of filepath.Clean, which I introduced after this due to gosec. So I think you're correct, but now we do need it because of the clean.

internal/service/service.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// monitor status for errors, so we don't zombie the service
+	s.wg.Add(1)
+	go s.monitorStatus(s.wg)
+	return nil
+}
+
+// monitorStatus monitors the collector's status for errors, and reports them
+// to the error channel to trigger a shutdown.
+func (s StandaloneCollectorService) monitorStatus(wg *sync.WaitGroup) {
```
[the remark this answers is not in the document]
BinaryFissionGames: Good callout, fixed this.

internal/service/service.go
```diff
@@ -0,0 +1,159 @@
...
+// is not running as a service.
+func runServiceInteractive(logger *zap.Logger, stopSignal <-chan os.Signal, svc RunnableService) error {
+	startupTimeoutCtx, startupCancel := context.WithTimeout(context.Background(), startTimeout)
+	defer startupCancel()
+
+	if err := svc.Start(startupTimeoutCtx); err != nil {
+		return fmt.Errorf("failed to start service: %w", err)
+	}
+
+	var err error
```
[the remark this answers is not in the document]
BinaryFissionGames: I do, I think it's confusing because it's named err and that's not a very specific name; I changed it to be more clear with what that error actually is.

internal/service/service.go
BinaryFissionGames: Refactored to be long-running after discussing on a call.

scripts/package/postinstall.sh
BinaryFissionGames: Took a pass at docs; Updated install docs and where it was referenced in install scripts too.

cpheps (approved):

BinaryFissionGames:
Did one last manual test with installing the collector on Windows/macOS/Linux, all install by default logging to a file.

PR #441, opened 2022-05-24
fix(bew): Improved Homebrew Installation

cpheps:
### Proposed Change
- Improved homebrew installation by making the install script more robust
- Added versioning to homebrew formulae (this will be manually managed until we add automation)
- Changed to use brew services over plist file
- Upgraded version of goreleaser with fixes that are needed to generate versioned brew file correctly

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


.github/workflows/release.yml
```diff
@@ -102,7 +102,7 @@ jobs:
       - name: Set up Go
```
BinaryFissionGames, line 102: There's another `1.5.0` version of goreleaser being used in this job too.

cpheps:
@BinaryFissionGames Updated goreleaser versions in release action

.github/workflows/release.yml
```diff
@@ -1,8 +1,8 @@
 name: release
 on:
   push:
-   tags:
-     - "v*"
+    tags:
+      - "v*"
 
 jobs:
   build-msi:
```
BinaryFissionGames, line 8: Looks like this job also sets up go 1.17 instead of 1.18

armstrmi (approved):

PR #442, opened 2022-05-24
feat(IIS): Add IIS logging plugin

armstrmi:
### Proposed Change

* Add a logging plugin for Microsoft IIS access logs. (ported from stanza-plugins)
Tested against a running Microsoft IIS instance to make sure access logs were parsing correctly.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

Sample log
image


plugins/iis_logs.yaml
```diff
@@ -0,0 +1,84 @@
...
+description: Log parser for IIS
+parameters:
+  - name: file_path
+    type: "[]string"
+    default: 
+      - "C:/inetpub/logs/LogFiles/W3SVC*/**/*.log"
+  - name: exclude_file_log_path
+    type: "[]string"
+    default: 
+      - ""
```
cpheps, line 12: Default should be `default: []`

plugins/iis_logs.yaml
```diff
@@ -0,0 +1,84 @@
...
+          layout: '%Y-%m-%d %H:%M:%S'
+          location: {{.location}}
+        severity: 
+          parse_from: attributes.status
+          mapping:
+            info: 2xx
+            info2: 3xx
+            warn: 4xx
+            error: 5xx
+        output: end_filter
```
cpheps, line 74: nit: this `output` isn't needed a it just goes to the next operator

plugins/iis_logs.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+      - id: end_filter
+        type: filter
+        expr: 'body == ""'
```
cpheps, line 78: Do you need this? I added it in other plugins that had a diverging pipeline in order to bring them back together. Do you expect some logs to have a blank body and want to filter those out?

plugins/iis_logs.yaml
StefanKurek: @cpheps can you explain this more?

cpheps (approved):

PR #443, opened 2022-05-24
feat: Updates docs with human readable names for components

### LINKED ISSUE #419: List of receivers, exporters, processors could be more helpful
The list of receivers shows the otel name for the receiver. A more useful list would be in table form and include the otel name as well as the human-readable name and possibly icon. Helpful for the user scanning for the integration they're looking for. Same with exporters, processors, etc

StefanKurek:
Resolves #419 

### Proposed Change
Updates documentation with two column tables for extensions, receivers, exporters, and processors. The first column has a human readable name and the second has the original otel name along with a link to the github README for that component.

#### Sample
image

### DIFF
--- a/docs/exporters.md
+++ b/docs/exporters.md
@@ -2,30 +2,31 @@
 
 Below is a list of supported exports with links to their documentation pages.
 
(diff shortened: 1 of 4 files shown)


cpheps (approved):
This looks good. Appreciate getting the correct name and casing.

BinaryFissionGames (approved):

PR #454, opened 2022-06-01
fix: Embed tzdata

jsirianni:
### Proposed Change

Some systems do not have a timezone database supported by Go.
- Container environments
- Windows

Go supports [embedding tzdata](https://pkg.go.dev/time/tzdata) which will be used as a fallback when the operating system does not provide a database.

We have several build systems (Gorelease, make collector) so I have opted to use the `import` option instead of the build tag.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	_ "time/tzdata"
 
 	"github.com/observiq/observiq-otel-collector/collector"
 	"github.com/observiq/observiq-otel-collector/internal/logging"


cmd/collector/main.go
```diff
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	_ "time/tzdata"
```
cpheps, line 22: Is it possible to only do this for the windows build?

cpheps (approved):

BinaryFissionGames (approved):

PR #455, opened 2022-06-02
deps: Update opamp library with new protobuf message format

andykellr:
### Proposed Change
Most of the changes are handled directly in the library implementation. There was also a change to remove the AuthorizationHeader and replace it with a general Header field to allow additional headers to be passed. This adds the User-Agent header setting it to `observiq-otel-collector/version`

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/opamp/observiq/observiq_client_test.go
+++ b/opamp/observiq/observiq_client_test.go
@@ -21,7 +21,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"reflect"
 	"testing"
 
 	colmocks "github.com/observiq/observiq-otel-collector/collector/mocks"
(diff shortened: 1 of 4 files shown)


opamp/observiq/observiq_client_test.go
```diff
@@ -251,7 +257,7 @@ func TestClientConnect(t *testing.T) {
 				mockOpAmpClient.On("Start", mock.Anything, mock.Anything).Return(nil).Run(func(args mock.Arguments) {
 					settings := args.Get(1).(types.StartSettings)
 					assert.Equal(t, expectedSettings.OpAMPServerURL, settings.OpAMPServerURL)
-					assert.Equal(t, expectedSettings.AuthorizationHeader, settings.AuthorizationHeader)
+					assert.True(t, reflect.DeepEqual(expectedSettings.Header, settings.Header), "Header elements should match")
```
cpheps, line 260: This can just be `assert.Equal` as it does a [reflect.DeepEqual](https://github.com/stretchr/testify/blob/083ff1c0449867d0d8d456483ee5fab8e0c0e1e6/assert/assertions.go#L58-L76)

PR #460, opened 2022-06-06
Run as root by default on Linux

jsirianni:
### Proposed Change

In order to support reading log files out of the box, we need the collector to run as the root user on Linux. The `observiq-otel-collector` does not have permission to read arbitrary log files. For example: Elasticsearch, Mongodb, Mysql, Nginx, Postgres, Rabbitmq, and redis are all owned by different users.

Changes
- Updated systemd unit file to use `root` user
- Updated post install script to configure the log file permissions
  - The logger will not modify permissions. If the file does not exist, the log file will be owned by root. We want to pre configure the permissions in order to allow users to switch to the `observiq-otel-collector` user.
- Updated linux doc to reflect these changes, and documented the process to switching to the unprivileged user.

I have tested the following
- Fresh install Runs as root
- Upgrades from v1.0.0 switch from oiq user to root user
- Documented systemd override procedure works (switching to unprivileged user)
- Log file rotates and maintains it's permissions (owned by observiq-otel-collector user)
- Manager config maintains its permissions (owned by observiq-otel-collector user) after being re-written by opamp.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/package/postinstall.sh
+++ b/scripts/package/postinstall.sh
@@ -40,7 +40,7 @@ StartLimitIntervalSec=120
 StartLimitBurst=5
 [Service]
 Type=simple
-User=observiq-otel-collector
+User=root
 Group=observiq-otel-collector
 Environment=PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
 Environment=OIQ_OTEL_COLLECTOR_HOME=/opt/observiq-otel-collector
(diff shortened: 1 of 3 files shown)


scripts/package/postinstall.sh
```diff
@@ -119,6 +119,13 @@ finish_permissions() {
     # We also change the owner of the binary to observiq-otel-collector
     chown -R observiq-otel-collector:observiq-otel-collector /opt/observiq-otel-collector/observiq-otel-collector /opt/observiq-otel-collector/plugins/*
     chmod 0640 /opt/observiq-otel-collector/plugins/*
+
+  # Initialize the log file to ensure it is owned by observiq-otel-collector.
+  # This prevents the service (running as root) from assigning ownership to
+  # the root user. By doing so, we allow the user to switch to observiq-otel-collector
+  # user for 'non root' installs.
+  touch /opt/observiq-otel-collector/log/collector.log
+  chown observiq-otel-collector:observiq-otel-collector /opt/observiq-otel-collector/log/collector.log
```
BinaryFissionGames, line 128: Is there any concern about log file rotation here? Like, when it rotates, does collector.log get new permissions or retain the old file's permissions? Or is the same file reused? I'm not sure how the rotation works ATM.

cpheps:
Please hold off merging this until we reconcile discussion with product.

PR #470, opened 2022-06-07
Adds manual msi build github action

StefanKurek:
### Proposed Change
Add new GitHub action that can be manually triggered on any branch to build a msi

https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.github/workflows/manual_msi_build.yml
+++ b/.github/workflows/manual_msi_build.yml
@@ -1,0 +1,91 @@
+name: manual_msi_build
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Collector Version"
+        required: true
+        default: 'v0.0.1'
+
+jobs:
+  build-msi:
+    runs-on: windows-2019
+    steps:
+      - name: Checkout Sources
+        uses: actions/checkout@v2
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: "1.18"
+          check-latest: true
+      - name: Cache Go Modules
+        uses: actions/cache@v2
+        with:
+          path: |
+            %LocalAppData%\go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/manual_msi_build.yml
```diff
@@ -0,0 +1,89 @@
...
+        run: |
+          curl -f -L -o go-msi.exe https://github.com/observIQ/go-msi/releases/download/v2.1.0/go-msi.exe
+          curl -f -L -o wix310-binaries.zip http://wixtoolset.org/downloads/v3.10.3.3007/wix310-binaries.zip
+          unzip wix310-binaries.zip
+        working-directory: C:/build
+      - name: Get Tag From Environment
+        id: get-tag
+        run: printf '::set-output name=tag::%s' "$(printf '%s' "${{ github.ref }}" | sed 's/refs\/tags\///')"
+      - name: "Build MSI from Tagged Release"
+        run: go-msi.exe make -m observiq-otel-collector.msi --version ${{ steps.get-tag.outputs.tag }} --arch amd64
```
BinaryFissionGames, line 65: Totally forgot about this stuff; We should remove this step (Get Tag From Environment), then we should have this version either as a workflow variable or hardcoded to something (otherwise the version won't makes sense, not sure what the implications of that is)

.github/workflows/manual_msi_build.yml
```diff
@@ -0,0 +1,89 @@
+name: manual_msi_build
+on:
+  workflow_dispatch:
```
cpheps, line 3: can we add a parameter to specify the branch/commit? Assuming this only works from main when merged?

StefanKurek: @cpheps if you look at that link in the summary, i think it lets you select the branch you want to use

.github/workflows/manual_msi_build.yml
BinaryFissionGames: Hardcoding the version for right now is probably simplest (just hardcoding to v0.0.1 or something)

StefanKurek: @BinaryFissionGames added an input (again similar to what you see in the link) for version with a default of "v0.0.1")

cpheps (approved):

PR #473, opened 2022-06-08
fix: Changes working directory to directory of executable on startup (for Windows)

StefanKurek:
### Proposed Change
On startup set the working directory to the directory of the executable. This fixes the Windows Service issue where the working directory is not where the executable lives.

```  
plugin:
  path: ./plugins/csv_logs.yaml
  parameters:
    log_paths:
      - ./input.csv
    start_at: beginning
    header: "1,2,3"
```

The above config produces errors related to not being able to find plugins/csv_logs.yaml after a windows install.  It works fine now after the proposed changes.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/internal/service/service_windows.go
+++ b/internal/service/service_windows.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"syscall"
 
 	"go.uber.org/zap"
@@ -42,6 +43,17 @@ func RunService(logger *zap.Logger, rSvc RunnableService) error {
 	}
 
 	if isService {
+		// Change working directory to executable directory
+		ex, err := os.Executable()
+		if err != nil {
+			logger.Warn("Failed to retrieve executable directory", zap.Error(err))
+		} else {
+			execDirPath := filepath.Dir(ex)
+			if err := os.Chdir(execDirPath); err != nil {
+				logger.Warn("Failed to modify current working directory", zap.Error(err))
+			}
+		}
+
 		// Service name doesn't need to be specified when directly run by the service manager.
 		return svc.Run("", newWindowsServiceHandler(logger, rSvc))
 	} else {


cmd/collector/main.go
```diff
@@ -75,6 +76,17 @@ func main() {
...
+	// Change working directory to executable directory (for Windows services)
+	ex, err := os.Executable()
+	if err != nil {
+		logger.Warn("Failed to retrieve executable directory", zap.Error(err))
+	} else {
+		execDirPath := filepath.Dir(ex)
+		if err := os.Chdir(execDirPath); err != nil {
+			logger.Warn("Failed to modify current working directory", zap.Error(err))
+		}
+	}
```
BinaryFissionGames, line 88: I think this could lead to some surprising behavior when running as not-a-service, or modifying the CWD for services on linux (which isn't often, but it could happen).

Can we make this logic only run for windows as a service? Possibly move it into internal/service/service_windows.go, before svc.Run is called.

StefanKurek: @BinaryFissionGames can you give me a more specific scenario that you are worried about? either way, you're right that it's probably not a bad idea to keep the change a little more compartmentalized

BinaryFissionGames: It's very edge case, but I'm thinking of a case where we run the collector from a different directory than where the exe is.

e.g. I have a folder on my system where I keep some sample configs, and run manually /opt/observiq-otel-collector/observiq-otel-collector --config ./test-config.yaml

It'd be very surprising to me that ./test-config.yaml does not refer to the directory relative to my CWD, but relative to the binary. Could see it being something the field team could possibly run into while doing support.

StefanKurek: @BinaryFissionGames yeah that's totally a fair and reasonable worry. I made the changes you requested and will test them out.

StefanKurek: @BinaryFissionGames alright made the changes and tested them. great catch!

cpheps (approved):

BinaryFissionGames (approved):

PR #474, opened 2022-06-08
deps: Updated to OTel v0.53.0

cpheps:
### Proposed Change
Updated to Otel v0.53.0

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):
I'm not sure if we should care about my comments here. A bunch of packages have been "deprecated" (https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.53.0), but not yet removed.

Don't know if we should be pro-active about moving off the deprecated stuff or not in these PRs, might be something to follow up on as things are removed instead. So I'll leave it up to you whether you think it's necessary to update these things now or not.

cpheps:
@BinaryFissionGames 

>I'm not sure if we should care about my comments here. A bunch of packages have been "deprecated" (https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.53.0), but not yet removed.

In general we should try to update as we know they're deprecated. Sometime we may punt them until the actual removal if we need to release quickly and the changes are more involved.

StefanKurek (approved):

PR #475, opened 2022-06-08
feat: Adds macOS plugin

StefanKurek:
### Proposed Change
Adds the macOS plugin

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/macos_logs.yaml
+++ b/plugins/macos_logs.yaml
@@ -1,0 +1,95 @@
+version: 0.0.1
+title: macOS
+description: Log parser for macOS
+parameters:
+  - name: enable_system_log
+    type: bool
+    default: true
+  - name: system_log_path
+    type: "[]string"
+    default:
+      - "/var/log/system.log"
+  - name: enable_install_log
+    type: bool
+    default: true
+  - name: install_log_path
+    type: "[]string"
+    default: 
+      - "/var/log/install.log"
+  - name: start_at
+    type: string
+    supported:
+      - beginning
+      - end
+    default: end
+template: |
+  receivers:
+    {{ if .enable_system_log }}
+    filelog/system_logs:
+      include: 
+        {{ range $fp := .system_log_path }}
+        - '{{ $fp }}'
+        {{ end }}
+      multiline: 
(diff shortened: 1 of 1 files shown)


plugins/macos_logs.yaml
```diff
@@ -0,0 +1,91 @@
+version: 0.0.1
+title: MacOS
```
cpheps, line 2: should probably be `macOS`

plugins/macos_logs.yaml
```diff
@@ -0,0 +1,91 @@
...
+description: Log parser for MacOS
```
cpheps, line 3: Should be `macOS`

plugins/macos_logs.yaml
```diff
@@ -0,0 +1,91 @@
...
+            parse_from: attributes.timestamp
+            layout_type: gotime
+            layout: '2006-01-02 15:04:05-07'
+        - id: time_parser_without_timezone
+          type: regex_parser
+          regex: '^(?P<timestamp>\w{3}\s*\d{1,2} \d{2}:\d{2}:\d{2}) (---|(?P<host>[^ ]*))? ((?P<process>[^\[]*)\[(?P<pid>[^\]]*)\])?( \((?P<subprocess>[^\[]*)(\[(?P<spid>[^\]]*)\])?\))?(: )?(?P<message>[\w\W]*)'
+          timestamp:
+            parse_from: attributes.timestamp
+            layout_type: gotime
+            layout: 'Jan _2 15:04:05'
```
cpheps, line 79: I think you need and add processor here just to point the two timestamp blocks back to a single point. See [mysql](https://github.com/observIQ/observiq-otel-collector/blob/main/plugins/mysql_logs.yaml#L634-L637) as example.

StefanKurek: fixed

plugins/macos_logs.yaml
StefanKurek: fixed

plugins/macos_logs.yaml
StefanKurek: fixed

cpheps (approved):

BinaryFissionGames (approved):

PR #477, opened 2022-06-08
feat: Update to latest GoogleCloud exporter config

### LINKED ISSUE #418: Update example config to use a simple config
in the base README we show a complex configuration file. Update this to a simple config to make it easier for those just getting started to understand

jmwilliams89:
### Proposed Change
- Updates the google cloud exporter to use the newest config structure
- Removes unused processors
- Removes the `location` and `namespace` config parameters
- Resolves #418 

##### Checklist
- [x] Changes are tested
- [x] CI has passed


exporter/googlecloudexporter/factory.go
```diff
@@ -38,10 +34,6 @@ const typeStr = "googlecloud"
 
 // NewFactory creates a factory for the googlecloud exporter
 func NewFactory() component.ExporterFactory {
-	// Hard code to use legacy config for now until we can fully test and understand new OTLPDirect behavior.
-	// Feature flag is in an init function so overridding must occur after inits are processed.
-	featuregate.GetRegistry().Apply(map[string]bool{"exporter.googlecloud.OTLPDirect": false})
```
StefanKurek, line 43: @jmwilliams89 does this mean the feature gate will be on or off for us?

exporter/googlecloudexporter/config.go
```diff
@@ -74,121 +67,42 @@ func (c *Config) setClientOptions() {
...
-	batchFactory := batchprocessor.NewFactory()
-	batchConfig := batchFactory.CreateDefaultConfig().(*batchprocessor.Config)
-	return batchConfig
-}
-
-// createDefaultNormalizerConfig creates a default normalizer config
-func createDefaultNormalizerConfig() *normalizesumsprocessor.Config {
-	normalizeFactory := normalizesumsprocessor.NewFactory()
-	return normalizeFactory.CreateDefaultConfig().(*normalizesumsprocessor.Config)
+	factory := batchprocessor.NewFactory()
```
StefanKurek, line 96: extremely minor: any reason why you give these variables prefixes in other create config methods but not here?

exporter/googlecloudexporter/factory.go
jmwilliams89: The default of OTLPDirect is true now. We had to add this before to force it to be false, which forced the exporter to use the legacy config.

exporter/googlecloudexporter/config.go
jmwilliams89: I think it was just stylistic. I can switch it back if you think it's better the other way. My thinking was that it's in a function called createDefaultBatchConfig and it's the only factory being made, so no need to specify what kind of factory it is. I think I should have done it this way the first time around.

StefanKurek: @jmwilliams89 i don't think one way is better than the other; my comment was more about consistency. i noticed at least one function which had it named the other way.

exporter/googlecloudexporter/README.md
```diff
@@ -6,57 +6,36 @@ This exporter can be used to send metrics, traces, and logs to Google Cloud Moni
 ## Configuration
 | Field               | Default               | Required | Description |
 | ---                 | ---                   | ---      | ---         |
+| `metric`            |                       | `false`  | The [metric](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/googlecloudexporter#configuration-reference) settings of the exporter. |
```
cpheps, line 9: These links should be to the version we're supporting so v0.53.0

exporter/googlecloudexporter/config.go
```diff
@@ -74,121 +67,42 @@ func (c *Config) setClientOptions() {
...
-	}
-
-	return gcpConfig
+func createDefaultGCPConfig() *gcp.Config {
+	factory := gcp.NewFactory()
+	config := factory.CreateDefaultConfig().(*gcp.Config)
+	config.RetrySettings.Enabled = false
+	config.UserAgent = defaultUserAgent
+	config.MetricConfig.Prefix = defaultMetricPrefix
+	config.LogConfig.DefaultLogName, _ = os.Hostname()
```
cpheps, line 90: Highly unlikely this errors but it'll return an empty string if it does. Is that ok?

jmwilliams89: Yeah. The default value if not set is already an empty string. So this is a noop if that happens. We only attempt to set it here to help customers avoid errors when sending logs. The current exporter will fail to send if the log entry doesn't have an attribute named gcp.log_name _or_ a default log name is not set.

exporter/googlecloudexporter/README.md
jmwilliams89: This should be fixed now.

exporter/googlecloudexporter/config.go
jmwilliams89: Good catch. I copied that function over from the old implementation and forgot to make that same style change. It should be consistent now.

exporter/googlecloudexporter/config.go
```diff
@@ -32,39 +28,36 @@ import (
...
-	Credentials             string                                       `mapstructure:"credentials"`
-	CredentialsFile         string                                       `mapstructure:"credentials_file"`
-	Location                string                                       `mapstructure:"location"`
-	Namespace               string                                       `mapstructure:"namespace"`
-	GCPConfig               *gcp.LegacyConfig                            `mapstructure:",squash"`
-	BatchConfig             *batchprocessor.Config                       `mapstructure:"batch"`
-	NormalizeConfig         *normalizesumsprocessor.Config               `mapstructure:"normalize"`
-	DetectorConfig          *resourcedetectionprocessor.Config           `mapstructure:"detector"`
-	AttributerConfig        *resourceprocessor.Config                    `mapstructure:"attributer"`
-	TransposerConfig        *resourceattributetransposerprocessor.Config `mapstructure:"transposer"`
```
BinaryFissionGames, line 51: This is a breaking change and will break existing users who are currently using these removed fields on update (from my understanding); What's our plan on migrating them?

cpheps (approved):

BinaryFissionGames (commented):
I imagine this also invalidates a LOT of our sample configs now, since the legacy config compatibility is broken. I also think metrics map to generic task now instead of generic node ?

Point being, this changes a lot, haha.

docs/getting-started.md
```diff
@@ -118,59 +118,19 @@ receivers:
...
-  # The batch processor aggregates incoming metrics into a batch,
-  # releasing them if a certain time has passed or if a certain number
-  # of entries have been aggregated.
-  batch:
-
 # Exporters send the data to a destination, in this case GCP.
 exporters:
   googlecloud:
     retry_on_failure:
       enabled: false
```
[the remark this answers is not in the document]
jmwilliams89: Good call. I've removed these from the sample config.

exporter/googlecloudexporter/config.go
jmwilliams89: It's actually not as much of a breaking change as you'd expect. We removed quite a few embedded processors, but no customers or any of our sample configs were actually configuring those values before. So we don't foresee any problems with those removals.

The two biggest breaking changes will be removing location and namespace as config options. We had to do that because the newest exporter no longer supports custom resource mappings. Instead, it automatically converts resources and if nothing matches, it falls back to generic node. Based on the logic in that conversion, the location and namespace can be defined by the `cloud.region` and `service.namespace` attributes respectively. So going forward, users will need to use a processor to configure this, instead of the exporter.

The field team has already been notified about this and reviewed the differences between the two exporters yesterday afternoon. DevOps also verified that the newest exporter worked without having to make changes to their configs.

BinaryFissionGames: I think what concerns me is that we have sample configs like:
https://github.com/observIQ/observiq-otel-collector/blob/9c5c21a40635d6a4cb5a3184b0c333b26d891c21/config/google_cloud_exporter/redis/config.yaml

Which are broken (haven't tested it, but it should fail to unmarshal since resource_mappings is no longer a valid key).

We might consider fixing up these configs "follow-up" work, but I feel like we should have these sample configs in a working state before release. (Unless I'm completely wrong and these configs work fine)

BinaryFissionGames (approved):

PR #482, opened 2022-06-09
fix: Modify goreleaser to have a safe config.yaml that will not be overwritten

StefanKurek:
### Proposed Change
goreleaser now creates a brew formula which puts the config file in the /etc directory of brew. It is given a observiq prefix (as this is a shared directory) and symlinked back to the standard brew install directory.

Also updated the macos install script to remove these configs on uninstall.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/docs/installation-mac.md
+++ b/docs/installation-mac.md
@@ -65,6 +65,10 @@ brew uninstall observiq/observiq-otel-collector/observiq-otel-collector
 
 # If you moved the opentelemetry-java-contrib-jmx-metrics.jar
 sudo rm -f /opt/opentelemetry-java-contrib-jmx-metrics.jar
+
+# To remove configs
+rm -f $(brew --prefix)/etc/observiq_config.yaml
+rm -f $(brew --prefix)/etc/observiq_config.yaml.default
 ```
 
 ### Uninstall script
--- a/scripts/install/install_macos.sh
+++ b/scripts/install/install_macos.sh
@@ -20,6 +20,7 @@ FORMULA_NAME="observiq/observiq-otel-collector/observiq-otel-collector"
 SERVICE_NAME="com.observiq.collector"
 
 # Script Constants
+BREW_ETC=$(brew --prefix)/etc/
 PREREQS="printf brew sed uname uuidgen tr"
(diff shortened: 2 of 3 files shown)


docs/installation-mac.md
```diff
@@ -65,6 +65,14 @@ brew uninstall observiq/observiq-otel-collector/observiq-otel-collector
...
 # If you moved the opentelemetry-java-contrib-jmx-metrics.jar
 sudo rm -f /opt/opentelemetry-java-contrib-jmx-metrics.jar
+
+# To remove configs run:
+# - on Mac M1:
+rm -f /opt/homebrew/etc/observiq_config.yaml
+rm -f /opt/homebrew/etc/observiq_config.yaml.default
+# - on Mac Intel:
+rm -f /usr/local/etc/observiq_config.yaml
+rm -f /usr/local/etc/observiq_config.yaml.default
```
cpheps, line 75: Is there a brew command that will give this to you?

cpheps: looks like it's `$(brew --prefix)/etc/observiq_config.yaml`

scripts/install/install_macos.sh
```diff
@@ -542,6 +545,11 @@ uninstall()
   info "Untapping formula..."
   brew untap observiq/homebrew-observiq-otel-collector > /dev/null 2>&1 || error_exit "$LINENO" "Failed to untap formula"
   succeeded
+
+  info "Removing config files..."
+  rm -f $BREW_ETC/observiq_config.yaml.default > /dev/null 2>&1
```
cpheps, line 550: `BREW_ETC` can be replaced with `$(brew --prefix)/etc/`. There are other examples in here using the `brew --prefix`

docs/installation-mac.md
StefanKurek: @cpheps yep, a much smarter way to do it. i'll make the changes

scripts/install/install_macos.sh
StefanKurek: @cpheps fixed

cpheps (approved):

PR #483, opened 2022-06-09
Move systemd unit file from heredoc to file (Linux packages)

jsirianni:
### Proposed Change

The current install method for the systemd unit file forces an overwrite. This was acceptable as the user can use a [systemd override](https://wiki.archlinux.org/title/systemd). 

Some users prefer to modify the service file directly, and want to rely on the package manager to determine if the configuration should be upgraded or left alone. We can accomplish this by moving the service file from the post install script to a static file managed by the built package.

#### Changes

- Added systemd unit file as a "config" type to be packaged with nfpm
- Post install script will continue to reload systemd

We loose the ability to detect the systemd directory, which might be okay as I tested on the following without issue
- centos 7, 8
- ubuntu 18, 20, 22
- debian 9, 10, 11
- sles 12, 15

If I modify the unit file directly, a package upgrade does not replace the file.

#### Considerations

Supporting pre systemd systems (ubuntu 14) will likely require that we build a second package that excludes systemd and includes an init script. The post install script provided this flexibility but has not been implemented yet.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


cpheps (approved):

BinaryFissionGames:
I'm REALLY apprehensive about this change. Having the customer able to change the service file directly means we CANNOT change our service configuration in the future (meaning, if we add new flags that need to be used, or add new environment variables that need to be used), because we can't overwrite their file.

Are we 100% sure we are OK with that? I don't think I understand why the customer wants this.

cpheps (commented):
@BinaryFissionGames Brings up a good point. I'm revoking my approval until we talk about this more.

jsirianni:
Closing for now. Will re-open if needed.

.goreleaser.yml
```diff
@@ -138,8 +138,15 @@ nfpms:
           mode: 0750
           owner: observiq-otel-collector
           group: observiq-otel-collector
+      - src: scripts/package/service/systemd/observiq-otel-collector.service
+        dst: /usr/lib/systemd/system/observiq-otel-collector.service
+        type: config
```
BinaryFissionGames, line 143: I'm kind of torn on whether this should be `config` or `config|noreplace`. I think `config|noreplace` is what is expected (the users changes are preferred no matter what, even if we change the service file).

The way it is right now, if we change the service file in our package, it will overwrite the user's changes (for RPM, dpkg will prompt the user to decide either way).

I did some digging, and ElasticSearch marks their service file as `config|noreplace`, so maybe we should too.

scripts/package/service/systemd/observiq-otel-collector.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
```
BinaryFissionGames, line 1: Do you mind moving this into the top-level `service` directory? We added the macOS plist file and I think they should be next to eachother.

BinaryFissionGames (commented):
Tested this on ubuntu 18.04 'cuz I remembered in test kitchen finding that `/usr/lib/systemd/system` didn't exist by default, but it runs as expected.

jsirianni:
> Tested this on ubuntu 18.04 'cuz I remembered in test kitchen finding that `/usr/lib/systemd/system` didn't exist by default, but it runs as expected.

Yes, the directory is not there by default but it gets created and functions as expected. 

I will make the changes you suggested 👍

BinaryFissionGames (approved):

PR #485, opened 2022-06-10
deps: Updated jmx jar version to v1.14.0

cpheps:
### Proposed Change
Updated JMX Jar to [v1.14.0](https://github.com/open-telemetry/opentelemetry-java-contrib/releases/tag/v1.14.0)

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (commented):
Can we also bump the version in the Dockerfile? Ideally we'd only have to do it here, but I don't think we pass it in right now.

BinaryFissionGames (approved):

armstrmi (approved):

PR #486, opened 2022-06-10
feat: Update GCP example configs to use latest Google exporter

jsirianni:
### Proposed Change

The latest Google exporter allows us to simplify most of the example configs

**Metrics**

- Removed batch and normalizesums, as these are handled by the exporter now
- Generally removed the need for RAT processor. It is still useful sometimes, but the new exporter allows us to copy resource attributes with the `metric.resource_filters[].prefix` setting.
- Generally removed the need for resource detection, as the exporter handles this. This could change in the future, and it will be trivial to re-add resource detection to these configs.
- Removed all resource mapping configs, this is no longer exposed.
- Use `resource_attributes` option on all jmx receivers.
- Set Google `metric.prefix` when metric names can conflict with other receivers

**Logs**

- Removed `project` requirement, no longer required

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/config/google_cloud_exporter/cadvisor/config.yaml
+++ b/config/google_cloud_exporter/cadvisor/config.yaml
@@ -2,44 +2,37 @@ receivers:
   prometheus:
     config:
       scrape_configs:
-      - job_name: 'cadvisor' # A friendly name, will be added as 'service.name' resource label
+      - job_name: 'cadvisor'
         scrape_interval: 60s
         static_configs:
         - targets:
           - cadvisor:8080
 
 processors:
+  # Detect the running containers hostname
+  resourcedetection:
+    detectors: ["system"]
+    system:
+      hostname_sources: ["os"]
+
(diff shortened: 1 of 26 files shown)


config/google_cloud_exporter/cadvisor/config.yaml
```diff
@@ -2,54 +2,45 @@ receivers:
...
 
 processors:
+  # Detect the running containers hostname
+  resourcedetection:
+    detectors: ["system"]
+    system:
+      hostname_sources: ["os"]
+
+  # Copy host.name to cadvisor.collector_hostname and then
+  # remove the origonal. Google will map to "generic_task",
```
cpheps, line 19:
```suggestion
  # remove the original. Google will map to "generic_task",
```

cpheps (approved):

jmwilliams89 (approved):

PR #487, opened 2022-06-10
feat(example config): New Relic Elasticsearch and Redis

jsirianni:
### Proposed Change

- Added Elasticsearch metrics for New Relic
- Added Redis metrics for New Relic
- Fixed Cassandra for New Relic (properties.otel.resource.attributes was replaced by resource_attributes)
  - Updated the attribute name to `cassandra.endpoint` to follow the pattern used in other jmx based configs

Redis includes the resource detection processor as it does not include identifiable attributes out of the box. Elasticsearch has node name and cluster name attributes that can be used for identification.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/config/new_relic_otlp/elasticsearch/README.md
+++ b/config/new_relic_otlp/elasticsearch/README.md
@@ -1,0 +1,7 @@
+# Elasticsearch Metrics with New Relic
+
+The Elasticsearch Receiver can be used to send Elasticsearch metrics to New Relic.
+
+## Prerequisites
+
+You must update this example config with the endpoint and license key (api key) for your account. See the [doc](../README.md) for New Relic prerequisites.
--- a/config/new_relic_otlp/elasticsearch/config.yaml
+++ b/config/new_relic_otlp/elasticsearch/config.yaml
@@ -1,0 +1,26 @@
+receivers:
+  elasticsearch:
+    nodes: ["_local"]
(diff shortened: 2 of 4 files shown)


cpheps (approved):

config/new_relic_otlp/cassandra/config.yaml
```diff
@@ -4,9 +4,8 @@ receivers:
     endpoint: localhost:7199
     target_system: cassandra,jvm
     collection_interval: 60s
-    properties:
-      # Attribute 'endpoint' will be used for generic_node's node_id field.
-      otel.resource.attributes: endpoint=localhost:7199
+    resource_attributes:
+      cassandra.endpoint: localhost:7199
```
StefanKurek, line 8: i don't know much about the new relic stuff. this doesn't break anything?

StefanKurek (commented):
Looks fine as far as I can tell other than my question

config/new_relic_otlp/cassandra/config.yaml
jsirianni: It will if someone is searching on that key. Other than that it is fine. The keys can be arbitrary. This change follows the pattern we are using for other jmx based configs.

These configs should be used as a reference only, so I think the change is okay.

StefanKurek (approved):

PR #488, opened 2022-06-10
feat(vCenter): Add vCenter logging plugin

armstrmi:
### Proposed Change

Add a logging plugin for VMWare vCenter logs .
Tested running a Fedora VM forwarding the tcp receiver into the syslog parser.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

Example Log:
image


plugins/vcenter_logs.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+  # Set Defaults
+  # {{$listen_address := default "0.0.0.0:5140" .listen_address}}
+  # {{$length := len $listen_address}}
+  # {{$listen_ip := default "0.0.0.0" .listen_ip}}
+  # {{$listen_port := default 5140 .listen_port}}
+  # {{$max_buffer_size := default "1024kib" .max_buffer_size}}
+  # {{$insecure := default false .insecure}}
+  # {{$certificate_file := default "" .certificate_file}}
+  # {{$key_file := default "" .key_file}}
```
cpheps, line 34: This isn't needed

plugins/vcenter_logs.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+        - id: vcenter_parser
+          type: syslog_parser
+          protocol: rfc5424
+
+  service:
+    pipelines:
+      logs:
+        receivers: [tcplog]
+
```
cpheps, line 81: add a newline

plugins/vcenter_logs.yaml
```diff
@@ -0,0 +1,81 @@
...
+              expr: 'body matches "^\\d* "'
+          default: vcenter_parser
+
+        - id: pre_parser
+          type: regex_parser
+          regex: '^(?P<drop>\d* )(?P<syslog_message>[\w\W]*)'
+
+        - id: remove_drop
+          type: remove
+          field: body.drop
```
cpheps, line 66: Is this `body.drop`? I'm wondering if it should be `attributes.drop` instead based on regex parses usually parsing into `attributes`

plugins/vcenter_logs.yaml
```diff
@@ -0,0 +1,70 @@
...
+          regex: '^(?P<drop>\d* )(?P<syslog_message>[\w\W]*)'
+
+        - id: remove_drop
+          type: remove
+          field: attributes.drop
+        
+        - id: move_syslog_message
+          type: move
+          from: attributes.syslog_message
+          to: attributes
```
cpheps, line 60: this won't work I think you probably want this to go to `body`

plugins/vcenter_logs.yaml
```diff
@@ -0,0 +1,70 @@
+version: 0.0.1
+title: VMware vCenter
+description: Log parser for VMware vCenter
+parameters:
+  - name: listen_port
+    type: int
+    default: 5142
```
StefanKurek, line 7: I'm assuming this was intentionally changed from the stanza plugin?

StefanKurek (commented):
LGTM with a small question

plugins/vcenter_logs.yaml
StefanKurek: so do operators resolve in order? so in this case we have a router which can 'jump' the sequential order for one case to the very last operator. but otherwise we go through each operator? this is just a knowledge question for me

cpheps: @StefanKurek 

> so do operators resolve in order? so in this case we have a router which can 'jump' the sequential order for one case to the very last operator. but otherwise we go through each operator? this is just a knowledge question for me

Yes, they do. If the `output` field on an operator isn't specified it just goes to the one below it. In the case of a router you'll need something to tie the split back together.

cpheps: @armstrmi That being said can you set the `output` of this to the `vcenter_parser` just to be explicit?

plugins/vcenter_logs.yaml
armstrmi: Yes, for the sake of testing this plugin, this was the port to point to for the tcp receive logs from the vCenter box I had loaded up, might be worth going back not entirely sure

plugins/vcenter_logs.yaml
StefanKurek: @cpheps why set the output of just this one and not any of the ones earlier in the chain? i almost feel like that makes it more arbitrary/confusing. just my opinion.

plugins/vcenter_logs.yaml
StefanKurek: @armstrmi ok, so has the default for these logs changed from whenever until today? just making sure we are picking a good default that makes sense.

armstrmi: I understand where you're coming from, looking back now it probably is best to change that back to the original port

PR #493, opened 2022-06-14
Added GitHub action to trigger homebrew action

cpheps:
### Proposed Change
Adds a GitHub action that will trigger on an edit to a release. This will occur when we change a pre-release to a release. The action sends a `repository_dispatch` event to the https://github.com/observIQ/homebrew-observiq-otel-collector repo to trigger the action to update the homebrew formulas.

Triggers action detailed in observIQ/homebrew-observiq-otel-collector#1.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.github/workflows/homebrew-update-current.yml
+++ b/.github/workflows/homebrew-update-current.yml
@@ -1,8 +1,8 @@
 name: update_homebrew_current
 on:
   release:
-    # edited is the event triggered when changing from a pre-release to a release on a published release
-    type: [edited]
+    # released is triggered when a pre-release is changed to a release see documentation https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#release
+    type: [released]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/homebrew-update-current.yml
```diff
@@ -0,0 +1,25 @@
+name: update_homebrew_current
+on:
+  release:
+    # edited is the event triggered when changing from a pre-release to a release on a published release
+    type: [edited]
```
BinaryFissionGames, line 5: Will this run if we do other edits to the release e.g. edit to add extra notes to the release?

This documentation: https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#release Seems to indicate that `released` triggers when a pre-release is changed to a release, which might make more sense.

BinaryFissionGames (approved):

PR #495, opened 2022-06-16
feat: Oracle Database log plugin

jsirianni:
### Proposed Change

Ported Oracle Database plugin from Stanza. Tested against Oracle 12.2c, no parser errors. I did not change anything other than adding `add` for log_type. 

##### Checklist
- [x] Changes are tested
- [x] CI has passed


plugins/oracle_database_logs.yaml
```diff
@@ -0,0 +1,209 @@
...
+          value: oracledb.listener
+    {{ end }}
+
+  service:
+    pipelines:
+      logs:
+        receivers:
+          - filelog/audit
+          - filelog/alert
+          - filelog/listener
```
BinaryFissionGames, line 209: I think these need conditionals based on the enable parameters (e.g. if enable_listener_log is false, filelog/listener shouldn't be included here)

plugins/oracle_database_logs.yaml
```diff
@@ -0,0 +1,209 @@
...
+  receivers:
+    {{ if .enable_audit_log }}
+    filelog/audit:
+      include:
+        {{ range $fp := .audit_log_path }}
+        - '{{ $fp }}'
+        {{end}}
+      start_at: {{ .start_at }}
+      attributes:
+        plugin_id: oracledb
```
cpheps, line 42: I think we've been removing the `plugin_id` attribute as we port these.

plugins/oracle_database_logs.yaml
jsirianni: Good call

BinaryFissionGames (approved):

cpheps (approved):

PR #512, opened 2022-06-21
feat(w3c): Add w3c logging plugin

armstrmi:
### Proposed Change

Add a logging plugin for w3c logs .
Tested running collector locally and testing logs from [iis](https://github.com/observIQ/log-library/blob/master/library/msiis/iis.log)

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

## Disclaimer
In the original stanza plugin, a timestamp was not parsed out and in order to maintain the original functionality while porting we did likewise, which is why it appears the timestamp does not parse out

Example Log:
image

### DIFF
--- a/plugins/w3c_logs.yaml
+++ b/plugins/w3c_logs.yaml
@@ -1,0 +1,149 @@
+version: 0.0.1
+title: W3C
+description: File Input W3C Parser
+min_stanza_version: 1.2.0
+parameters:
+  - name: file_log_path
+    type: strings
+    required: true
+  - name: exclude_file_log_path
+    type: strings
+    default: []
+  - name: encoding
+    type: enum
+    valid_values:
+      - utf-8
+      - utf-16le
+      - utf-16be
+      - ascii
+      - big5
+    default: utf-8
+  - name: log_type
+    type: string
+    default: w3c
+  - name: start_at
+    type: enum
+    valid_values:
+      - beginning
+      - end
+    default: end
+  - name: delete_after_read
+    type: bool
+    default: false
+  - name: max_concurrent_files
+    type: int
+    default: 512
+    advanced_config: true
+  - name: include_file_name
+    type: bool
+    default: true
+  - name: include_file_path
+    type: bool
+    default: false
+  - name: include_file_name_resolved
+    type: bool
+    default: false
+  - name: include_file_path_resolved
+    type: bool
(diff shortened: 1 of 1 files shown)


plugins/w3c_logs.yaml
```diff
@@ -0,0 +1,109 @@
...
+    default: false
+  - name: header
+    type: "string"
+    required: true 
+  - name: delimiter
+    type: string
+    default: "\t"
+  - name: header_delimiter
+    type: string
+    default: "\t"
```
BinaryFissionGames, line 55: Looks like this isn't used (otel doesn't expose this option); We should remove it.

plugins/w3c_logs.yaml
```diff
@@ -0,0 +1,109 @@
...
+parameters:
+  - name: file_log_path
+    type: "[]string"
+    required: true
+    default: 
+      - "/var/log/zookeeper/zookeeper.log"
+  - name: exclude_file_log_path
+    type: "[]string"
+    default: 
+      - ""
```
BinaryFissionGames, line 13: This default should be an empty list
```suggestion
    default: []
```

plugins/w3c_logs.yaml
```diff
@@ -0,0 +1,109 @@
+version: 0.0.1
+title: W3C
+description: File Input W3C Parser
```
BinaryFissionGames, line 3: I think we've been following the format of `Log parser for <technology>`
```suggestion
description: Log Parser for W3C
```

plugins/w3c_logs.yaml
```diff
@@ -0,0 +1,103 @@
...
+
+        - id: quote_handler_parser
+          type: regex_parser
+          parse_from: body
+          regex: '(?P<message1>[^"]*)(?P<first_quote>[\"])(?P<message2>[^"]*)(?P<second_quote>[\"])(?P<message3>.*)'
+
+        - id: quote_handler_restructurer
+          type: add
+          field: body
+          value: 'EXPR(body.message1 + body.message2 + body.message3)'
```
cpheps, line 93: this should be `attributes.message1`, `attributes.message2`, `attributes.message3`. After the regex parse it goes to the attributes.

cpheps (approved):

BinaryFissionGames (approved):

PR #519, opened 2022-06-23
feat: Rework macOS script to install from tarball

BinaryFissionGames:
### Proposed Change
* Add a plist file to the archive for the macOS service.
  * Like with the windows one, there doesn't seem to be a way to restrict this to specific tarballs, so it's packaged with all archives.
* Rework the macOS script to install from tarball instead of brew
* Remove brew references from docs
* Remove brew from goreleaser

Since the new service file is required in the tarball, I've been testing using the `-l` flag to specify a local tarball built with `make release-test`.
- Configs (config.yaml, logging.yaml, manager.yaml) are not overwritten on re-install.
- On uninstall, config.yaml is preserved as config.bak.yaml, which is similar to behavior when uninstalling through RPM.
- Installing on top of an existing install will overwrite files of the existing install, similar to the linux behavior.
- The collector is installed under `/opt/observiq-otel-collector`, similar to the linux installs.
- The collector is installed as root, meaning all the files belong to the "root" user and the "wheel" group. MacOS groups and users seem a little different than on linux (useradd/groupadd don't exist).
  - This might be something we need to take a closer look at, but I think we'd have similar issues either way, since the collector runs as root and creates its log files as root.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -53,6 +53,9 @@ archives:
       - src: release_deps/VERSION.txt
         dst: "."
         strip_parent: true
+      - src: release_deps/com.observiq.collector.plist
+        dst: "install"
+        strip_parent: true
 
     format_overrides:
       - goos: windows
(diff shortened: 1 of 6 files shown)


docs/installation-mac.md
```diff
@@ -1,79 +1,40 @@
...
+The collector uses `launchctl` to control the collector lifecycle using the following commands.
 
 ```sh
 # Start the collector
-brew services start com.observiq.collector
+launchctl start com.observiq.collector
 
 # Stop the collector
-brew services stop com.observiq.collector
+launchctl stop com.observiq.collector
```
cpheps, line 32: Do these need `sudo` prefixes?

scripts/install/install_macos.sh
```diff
@@ -360,106 +415,104 @@ set_opamp_secret_key()
   fi
 }
 
-set_formula_name()
+# latest_version gets the tag of the latest release.
+# It's also super hacky; It assumes that the output is formatted prettily.
+# Ideally we'd use jq, but we want to avoid external dependencies
```
cpheps, line 420: Can we remove these two lines? It would be fine if this was a dev comment in source but this is a customer facing script that they can easily read.

scripts/install/install_macos.sh
```diff
@@ -360,106 +415,104 @@ set_opamp_secret_key()
...
-  found_name=$(brew list --full-name | grep "observiq/observiq-otel-collector/observiq-otel-collector")
-  set -e
-
-  if [ -n "$found_name" ]; then
-    FORMULA_NAME=$found_name
-  fi
-}
+  # Remove temporary directory, if it exists
+  rm -rf "$TMP_DIR"
+  mkdir -p "$TMP_DIR"
```
cpheps, line 436: This isn't needed as `-p` creates all missing directories so we can just call `mkdir -p "$TMP_DIR/artifacts"`

service/com.observiq.collector.plist
```diff
@@ -0,0 +1,23 @@
...
+    <key>OIQ_OTEL_COLLECTOR_HOME</key>
+    <string>[INSTALLDIR]</string>
+    </dict>
+    <key>ProgramArguments</key>
+    <array>
+    <string>[INSTALLDIR]observiq-otel-collector</string>
+    <string>--config</string>
+    <string>[INSTALLDIR]config.yaml</string>
+    <string>--logging</string>
+    <string>[INSTALLDIR]logging.yaml</string>
```
cpheps, line 18: We can probably add the flag for manager.yaml. If the file is missing it goes into standalone mode.

docs/installation-mac.md
BinaryFissionGames: Yep, good catch.

service/com.observiq.collector.plist
BinaryFissionGames: Good catch, I copy pasted this from the old service definition we had (before goreleaser added options for this).

I added in KeepAlive and WorkingDirectory, as well as the manager.yaml command line.

StefanKurek (approved):
LGTM

BinaryFissionGames:
@cpheps I've made a few changes:
* Move root check to be under the "Checking Prerequisites" section
* Only uninstall if the collector is already installed

I also moved around the ordering of the config backup, to make it slightly more robust (but it shouldn't matter with the install check).

.goreleaser.yml
```diff
@@ -53,6 +53,9 @@ archives:
       - src: release_deps/VERSION.txt
         dst: "."
         strip_parent: true
+      - src: release_deps/com.observiq.collector.plist
+        dst: "install"
+        strip_parent: true
```
armstrmi, line 58: Is there a guide or read me that goes over the different fields in this file, it seems pretty straight forward in regards to knowing that this is talking about the archives and what not, but am unsure what all other fields are included with this. Basically wondering if there is a readme like there is for otel operators.

Makefile
```diff
@@ -158,6 +158,7 @@ release-prep:
 	@cp -r ./plugins release_deps/
 	@cp config/example.yaml release_deps/config.yaml
 	@cp config/logging.yaml release_deps/logging.yaml
+	@cp service/com.observiq.collector.plist release_deps/com.observiq.collector.plist
```
armstrmi, line 161: Why is there the need to add this command? like why copy the service file into the release dep file for the sake of prepping for a release?

.goreleaser.yml
BinaryFissionGames: This specific section (archive) is here: https://goreleaser.com/customization/archive/
You should be able to find all the docs on that site.
LMK if you have some specific questions on any of these fields.

Makefile
BinaryFissionGames: This is a good question; @cpheps Do you remember why we implemented this pattern of copying to release_deps instead of referencing the original path directly in goreleaser? It makes sense for the stuff we download, not sure why we do this for plugins or config.yaml, actually.

cpheps: I think it was for downloading. Plugins use to be downloaded as well as the JMX jar. We probably don't need it now if there's a known static location for things.

cpheps (approved):

Makefile
cpheps: We also do renames in there looks like for the `config.yaml`

BinaryFissionGames: I'll leave this as-is for right now to keep it consistent with the other artifacts; Might be something to revisit later.

PR #520, opened 2022-06-24
feat: Add TLS Support to Opamp

dmikolay:
### Proposed Change
- Added TLS support to Opamp connector

##### Checklist
- [X] Changes are tested
- [X] CI has passed


cpheps:
Tested this against a live BindPlane OP Instance to verify it works in Insecure, TLS, and mTLS mode.

opamp/observiq/observiq_client.go
```diff
@@ -128,13 +128,18 @@ func (c *Client) Connect(ctx context.Context) error {
 		return err
 	}
 
+	tlsCfg, err := c.currentConfig.ToTLS()
+	if err != nil {
```
StefanKurek, line 132: should we add a new test case to check for this condition?

opamp/config.go
```diff
@@ -29,19 +32,82 @@ var (
...
+func (c Config) ToTLS() (*tls.Config, error) {
+	if c.TLS == nil {
+		return nil, nil
+	}
+
+	tlsConfig := &tls.Config{
+		MinVersion: tls.VersionTLS12,
+	}
+
+	if c.TLS.Insecure {
```
StefanKurek, line 79: looks like the "Insecure" config param is being used synonymously as "InsecureSkipVerify". is there a reason for this? for example in otel's tls config options they have both an "insecure" to turn TLS on/off and a "insecure_skip_verify" to avoid verifying the certificate

opamp/config.go
```diff
@@ -58,11 +124,37 @@ func ParseConfig(configLocation string) (*Config, error) {
...
+		// If CA file is specified
+		if config.TLS.CAFile != nil {
+			// Validate CA file exists on disk
+			if _, err := os.Stat(*config.TLS.CAFile); errors.Is(err, os.ErrNotExist) {
+				return nil, errors.New(errInvalidCAFile)
+			}
+		}
+
+		switch {
+		case config.TLS.CertFile == nil && config.TLS.KeyFile == nil: // Not using mTLS
```
StefanKurek, line 138: any reason why this doesn't give an error too?

opamp/config.go
cpheps: We could change it.

jsirianni (commented):
Exciting to see this. Left a couple comments.

opamp/config.go
```diff
@@ -29,19 +32,82 @@ var (
...
+}
+
+// ToTLS converts the config to a tls.Config
+func (c Config) ToTLS() (*tls.Config, error) {
+	if c.TLS == nil {
+		return nil, nil
+	}
+
+	tlsConfig := &tls.Config{
+		MinVersion: tls.VersionTLS12,
```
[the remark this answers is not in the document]
cpheps: I think might be a down the road thing. It doesn't look like BP allows this to be specified yet so once we add that We should add it to the collector also.

StefanKurek (approved):

jsirianni (approved):

PR #529, opened 2022-06-27
feat: Create Updater artifact

BinaryFissionGames:
### Proposed Change
* Create a new "updater" binary
* Add make target for updater
* Modify other build targets to include building the updater
* Add to MSI build
  * It's automatically added to other builds (linux packages, archives)
* Implement version print flag for updater

Tested on linux & windows, checked the tar archive for darwin to make sure that the updater exists.

Tested manual build for Win MSI.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


cpheps:
Curious if there are any concerns with this not being in its own package and being it's own go.mod? It's currently sharing a mod file with the collector. Do we expect a lot of overlap between the two that it's worth sharing a module?

BinaryFissionGames:
That's something that I haven't considered; It might make sense to split into a separate module. I'll try it out and see what it's like.

cpheps:
>That's something that I haven't considered; It might make sense to split into a separate module. I'll try it out and see what it's like.

@BinaryFissionGames you might hit issues with goreleaser. I thought you need the pay version to do multiple binaries in submodules. I think it's the [monorepo](https://goreleaser.com/customization/monorepo/) feature set.

BinaryFissionGames:
Yeah, looks like go does not play nice with the updater binary being in its own module:
```
cd ./updater/; go build -ldflags "-s -w -X github.com/observiq/observiq-otel-collector/internal/version.version=v1.2.0" -o ./dist/updater_darwin_amd64 ./cmd/updater
main module (github.com/observiq/observiq-otel-collector) does not contain package github.com/observiq/observiq-otel-collector/updater/cmd/updater
```

Might makes sense to make a new module for all the underlying logic, but it doesn't seem to work for the binary/

.github/dependabot.yml
```diff
@@ -44,3 +44,14 @@ updates:
...
       prefix: "deps"
       include: "scope"
+  - package-ecosystem: "gomod"
+    directory: "/updater"
+    schedule:
+      interval: "weekly"
+    ignore:
+      # Opentelemetry updates will be done manually
+      - dependency-name: "github.com/open-telemetry/opentelemetry-collector*"
+      - dependency-name: "go.opentelemetry.io/collector/*"
```
cpheps, line 54: We likely won't need these but we can keep them for now.

updater/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/observiq/observiq-otel-collector/updater
+
+go 1.17
```
cpheps, line 3: Could do go 1.18 if you want not necessary

updater/cmd/updater/main.go
```diff
@@ -0,0 +1,35 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+
+	"github.com/observiq/observiq-otel-collector/internal/version"
```
cpheps, line 20: Do it make sense to break this out into its own package? Seems weird to share a internal package from another module.

.github/dependabot.yml
BinaryFissionGames: Yeah, I was thinking we PROBABLY won't need these, but it shouldn't hurt.

updater/cmd/updater/main.go
BinaryFissionGames: Honestly, not really sure.

It does feel weird, but OTel does it all the time, and it seems to work fine for them. 

I don't really have any major objections to breaking it out, but it also feels weird to have this as a public package when it requires specific build steps and isn't really meant to be used outside this repo.

Makefile
```diff
@@ -40,27 +52,27 @@ build-windows: build-windows-amd64
 
 .PHONY: build-linux-amd64
 build-linux-amd64:
-	GOOS=linux GOARCH=amd64 $(MAKE) collector
+	GOOS=linux GOARCH=amd64 $(MAKE) build-binaries -j2
```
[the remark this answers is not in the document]
BinaryFissionGames: `-j` specifies concurrency; This is saying that it can do 2 targets at a time, so it will build the updater + collector concurrently.

updater/cmd/updater/main.go
BinaryFissionGames: @StefanKurek Do you also think this should be broken into a separate package? Or possibly duped into a package internal to the new module?

IMO the way it is now is the most straightforward, but I can see how it's a little wonky.

StefanKurek: @BinaryFissionGames @cpheps honestly i'll go by what corbin says as he's the go master

StefanKurek (approved):

updater/cmd/updater/main.go
cpheps: it is wonky, I think the most correct go way would be to copy the version package into an internal package in the updater. Right now the updater is importing the collector as a dependency for that one package even though it doesn't actually need the collector.

cpheps:
@armstrmi Because the updater is meant to be its own binary having it in its own module we can better separate it from the collector binary. By having its own module to we can separate the dependencies needed for the updater from the collector dependencies. Since the collector by, nature of what it is, has a large number of dependencies this could over complicate managing dependencies for two separates binaries.

If we come to find there there is some shared code between the two binaries we could break that shared code out into a third module that is shared between the two.

armstrmi (approved):

PR #532, opened 2022-06-29
feat: Versioned Linux Artifacs

cpheps:
### Proposed Change
- Added version numbers to `deb` and `rpm` release artifacts
- Updated `install_unix.sh` script to handle these
- Updated linux documentation to mention version when running `dpkg` and `rpm` manually

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/install/install_unix.sh
+++ b/scripts/install/install_unix.sh
@@ -16,7 +16,7 @@
 set -e
 
 # Collector Constants
-PACKAGE_NAME="observiq-otel-collector_linux"
+PACKAGE_NAME="observiq-otel-collector"
 DOWNLOAD_BASE="https://github.com/observiq/observiq-otel-collector/releases"
 
 # Script Constants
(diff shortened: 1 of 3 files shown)


scripts/install/install_unix.sh
```diff
@@ -354,23 +358,25 @@ set_package_type()
...
-
   if [ -z "$url" ] ; then
-    url=$DOWNLOAD_BASE
+    if [ -z "$version" ] ; then
+      # shellcheck disable=SC2153
+      version=$COLLECTOR_VERSION
+    fi
 
-    # Only if we're using the base download url apply version formatting
     if [ -z "$version" ] ; then
```
StefanKurek, line 370: @cpheps is this the same check two times in a row with different results?

StefanKurek: ignore me. i get it now 😄

cpheps: Yeah it's mostly if we don't know the collector version the package it downloads too isn't named weird.

StefanKurek (approved):

PR #533, opened 2022-06-29
fix: Rename the official full name of the collector to "observIQ Distro for OpenTelemetry Collector"

cpheps:
### Proposed Change
Per https://github.com/open-telemetry/community/issues/1093 we are rebranding the collector to `observIQ Distro for OpenTelemetry Collector`

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cpheps:
@BinaryFissionGames Hoping you can advise on if/what we should change for the windows install. We do not want to change the install path of `C:\Program Files\observIQ OpenTelemetry Collector\` or change service names as this will break backwards compatibility with current Windows installs. If there are any display name only fields we can update Let me know and I can do so.

BinaryFissionGames:
@cpheps I'll look into this more tomorrow; ProductName seems tied to the default install directory as-is, but there's probably a way to change that in the product.wxs.

As far as the service goes, the `display-name` for the service in wix.json should be fine to change, it's independent of any commands we give.

cpheps:
>  I noticed there was also a few fields in the Windows installation file that were not changed, but it seems like we are trying not to changes those correct?

@armstrmi  yeah we're trying to not change install paths or any services names that may break customer environments when they install the next version of the collector.

BinaryFissionGames:
I managed to get the Windows stuff working;
I tested:
* Upgrading from v1.2.0 to the new version
* Upgrading from v1.2.0 to the new version with a non-default install path
* Upgrading from the new version to a newer version
* Upgrading from the new version to a newer version with a non-default install path

I checked to see if config.yaml and manager.yaml were persisted as well.

StefanKurek (approved):

cpheps:
@BinaryFissionGames Does any of the Windows documentation needed to be updated to reflect the new service name?

PR #535, opened 2022-06-30
feat: Added additional headers to OpAMP client

cpheps:
### Proposed Change
Added the following headers to the observIQ client:
- `OpAMP-Version`: internally determined version of OpAMP
- `AgentID`: UUID of Agent
- `Agent-Hostname`: Hostname of agent
- `Agent-Version`: Version of agent

These headers are meant as a fail safe for OpAMP communication with server. Since OpAMP is not stable yet it's possible proto definitions will change and that a user may have a sever and agent that are incompatible. These headers will be a way to allow the server to recognize if the version of OpAMP the client is using has compatible protos. If not the server can take action accordingly. We also include some identifying information in the headers so the server can have some identity on the server even if there are OpAMP compatibility issues.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/opamp/observiq/observiq_client.go
+++ b/opamp/observiq/observiq_client.go
@@ -139,7 +139,7 @@ func (c *Client) Connect(ctx context.Context) error {
 			"Authorization":  []string{fmt.Sprintf("Secret-Key %s", c.currentConfig.GetSecretKey())},
 			"User-Agent":     []string{fmt.Sprintf("observiq-otel-collector/%s", version.Version())},
 			"OpAMP-Version":  []string{opamp.Version()},
-			"AgentID":        []string{c.ident.agentID},
+			"Agent-ID":       []string{c.ident.agentID},
 			"Agent-Version":  []string{version.Version()},
 			"Agent-Hostname": []string{c.ident.hostname},
 		},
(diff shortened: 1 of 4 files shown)


opamp/observiq/observiq_client.go
```diff
@@ -136,8 +136,12 @@ func (c *Client) Connect(ctx context.Context) error {
 	settings := types.StartSettings{
 		OpAMPServerURL: c.currentConfig.Endpoint,
 		Header: http.Header{
-			"Authorization": []string{fmt.Sprintf("Secret-Key %s", c.currentConfig.GetSecretKey())},
-			"User-Agent":    []string{fmt.Sprintf("observiq-otel-collector/%s", version.Version())},
+			"Authorization":  []string{fmt.Sprintf("Secret-Key %s", c.currentConfig.GetSecretKey())},
+			"User-Agent":     []string{fmt.Sprintf("observiq-otel-collector/%s", version.Version())},
+			"OpAMP-Version":  []string{opamp.Version()},
+			"AgentID":        []string{c.ident.agentID},
```
andykellr, line 142: I think I would use `Agent-ID` to be consistent with the other `Agent-` headers. Otherwise I think this looks great.

cpheps: @andykellr Updated.

andykellr (approved):

PR #536, opened 2022-06-30
feat: Manager Env Config

armstrmi:
### Proposed Change

Implement dynamic manager.yaml loading/creation based upon pre-existing manager.yaml file or preconfigured environment variables.

Logic for loading/creating a yaml file:

1. Check if manager.yaml exists in flag specified location
2. If not check if endpoint env exists.
  - If so check all other envs
  - If agent id is not specified via env then create a UUID
  - Create opamp.Config with env values
  - Marshal created config to yaml and write out to disk at `./manager.yaml`
3. If file does not exist and no envs are set go into standalone mode

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -33,10 +33,11 @@ import (
 )
 
 const (
-	ENDPOINT   = "OPAMP_ENDPOINT"
-	AGENT_ID   = "OPAMP_AGENT_ID"
-	SECRET_KEY = "OPAMP_SECRET_KEY"
-	LABELS     = "OPAMP_LABELS"
+	// env variable name constants
+	endpoint  = "OPAMP_ENDPOINT"
+	agentID   = "OPAMP_AGENT_ID"
+	secretKey = "OPAMP_SECRET_KEY"
+	labels    = "OPAMP_LABELS"
 )
 
 func main() {
@@ -114,21 +115,21 @@ func checkManagerConfig(configPath *string) error {
 	} else if errors.Is(err, os.ErrNotExist) {
 		// manager.ymal file does *not* exist, create file using env variables
 		newConfig := &opamp.Config{}
-		if endpoint, ok := os.LookupEnv(ENDPOINT); ok {
-			newConfig.Endpoint = endpoint
+		if ep, ok := os.LookupEnv(endpoint); ok {
+			newConfig.Endpoint = ep
 
-			var agentID string
-			if agentID, ok = os.LookupEnv(AGENT_ID); !ok {
-				agentID = uuid.New().String()
+			var ai string
+			if ai, ok = os.LookupEnv(agentID); !ok {
+				ai = uuid.New().String()
 			}
-			newConfig.AgentID = agentID
+			newConfig.AgentID = ai
 
-			if secret_key, ok := os.LookupEnv(SECRET_KEY); ok {
-				newConfig.SecretKey = &secret_key
+			if sk, ok := os.LookupEnv(secretKey); ok {
+				newConfig.SecretKey = &sk
 			}
 
-			if labels, ok := os.LookupEnv(LABELS); ok {
-				newConfig.Labels = &labels
+			if label, ok := os.LookupEnv(labels); ok {
+				newConfig.Labels = &label
 			}
 
 			var data []byte
@@ -139,11 +140,9 @@ func checkManagerConfig(configPath *string) error {
 			if err := os.WriteFile(*configPath, data, 0777); err != nil {
 				panic(err)
 			}
-
 			return err
-		} else {
-			return os.ErrNotExist
 		}
+		return os.ErrNotExist
 	}
 	return os.ErrInvalid
 }
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -116,21 +116,22 @@ func checkManagerConfig(configPath *string) error {
 		// We found the file just return nil
 		return nil
 	case errors.Is(statErr, os.ErrNotExist):
-		var ep, ai string
 		var ok bool
-		// Endpoint is only required env
-		if ep, ok = os.LookupEnv(endpointENV); !ok {
-			// Envs were not found and statErr is os.ErrNotExist so return that
-			return statErr
-		}
+
 		// manager.ymal file does *not* exist, create file using env variables
 		newConfig := &opamp.Config{}
 
-		newConfig.Endpoint = ep
-		if ai, ok = os.LookupEnv(agentIDENV); !ok {
-			ai = uuid.New().String()
+		// Endpoint is only required env
+		newConfig.Endpoint, ok = os.LookupEnv(endpointENV)
+		if !ok {
+			// Envs were not found and statErr is os.ErrNotExist so return that
+			return statErr
 		}
-		newConfig.AgentID = ai
+
+		newConfig.AgentID, ok = os.LookupEnv(agentIDENV)
+		if !ok {
+			newConfig.AgentID = uuid.New().String()
+		}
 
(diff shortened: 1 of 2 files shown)


cmd/collector/main.go
```diff
@@ -21,12 +21,22 @@ import (
...
 	"github.com/spf13/pflag"
 	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	ENDPOINT   = "OPAMP_ENDPOINT"
+	AGENT_ID   = "OPAMP_AGENT_ID"
+	SECRET_KEY = "OPAMP_SECRET_KEY"
+	LABELS     = "OPAMP_LABELS"
```
cpheps, line 39: these should be lower cased. The linter is yelling at you because they are exported with no comment.

cmd/collector/main.go
```diff
@@ -95,3 +105,45 @@ func logOptions(loggingConfigPath *string) ([]zap.Option, error) {
...
+				newConfig.SecretKey = &secret_key
+			}
+
+			if labels, ok := os.LookupEnv(LABELS); ok {
+				newConfig.Labels = &labels
+			}
+
+			var data []byte
+			if data, err = yaml.Marshal(newConfig); err != nil {
+				panic(err)
```
cpheps, line 136: don't panic here return and error

cmd/collector/main.go
```diff
@@ -95,3 +105,45 @@ func logOptions(loggingConfigPath *string) ([]zap.Option, error) {
...
+			}
+			// write data to a manager.yaml file, with 0777 file permission
+			if err := os.WriteFile(*configPath, data, 0777); err != nil {
+				panic(err)
```
cpheps, line 140: don't panic here return an error

cpheps (changes requested):

cmd/collector/main.go
```diff
@@ -21,12 +21,23 @@ import (
...
 	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	// env variable name constants
+	endpoint  = "OPAMP_ENDPOINT"
+	agentID   = "OPAMP_AGENT_ID"
+	secretKey = "OPAMP_SECRET_KEY"
+	labels    = "OPAMP_LABELS"
```
cpheps, line 40: We should add an addition env for agent name

cmd/collector/main.go
```diff
@@ -95,3 +106,43 @@ func logOptions(loggingConfigPath *string) ([]zap.Option, error) {
...
+				newConfig.SecretKey = &sk
+			}
+
+			if label, ok := os.LookupEnv(labels); ok {
+				newConfig.Labels = &label
+			}
+
+			var data []byte
+			if data, err = yaml.Marshal(newConfig); err != nil {
+				return err
```
cpheps, line 137: Can we wrap this error to give more context something like `fmt.Errorf("failed to marshal config: %w", err)`

cmd/collector/main.go
```diff
@@ -95,3 +106,43 @@ func logOptions(loggingConfigPath *string) ([]zap.Option, error) {
...
+			}
+			// write data to a manager.yaml file, with 0777 file permission
+			if err := os.WriteFile(*configPath, data, 0777); err != nil {
+				return err
```
cpheps, line 141: Same here something like `fmt.Errorf("failed to write config file created from ENVs: %w", err)`

cmd/collector/main.go
```diff
@@ -95,3 +106,43 @@ func logOptions(loggingConfigPath *string) ([]zap.Option, error) {
 
 	return l.Options()
 }
+
+func checkManagerConfig(configPath *string) error {
+	if _, err := os.Stat(*configPath); err == nil {
```
cpheps, line 111: I think this if statement might need a restructure a bit. I'd like to avoid the `os.ErrInvalid` at the end. Let's sync up on this.

cmd/collector/main.go
```diff
@@ -95,3 +107,57 @@ func logOptions(loggingConfigPath *string) ([]zap.Option, error) {
...
+func checkManagerConfig(configPath *string) error {
+	_, statErr := os.Stat(*configPath)
+
+	switch {
+	case statErr == nil:
+		// We found the file just return nil
+		return nil
+	case errors.Is(statErr, os.ErrNotExist):
+		// Endpoint is only required env
+		if ep, ok := os.LookupEnv(endpoint); ok {
```
BinaryFissionGames, line 120: This is super nipicky; But would it be better to do something like:
```go
ep, ok := os.LookupEnv(endpoint)
if !ok {
    return statErr
}
// stuff that was previously nested in the if statment
...
return nil
```

To avoid this nesting here?

cmd/collector/main.go
```diff
@@ -95,3 +107,57 @@ func logOptions(loggingConfigPath *string) ([]zap.Option, error) {
...
+			if label, ok := os.LookupEnv(labelsENV); ok {
+				newConfig.Labels = &label
+			}
+
+			data, err := yaml.Marshal(newConfig)
+			if err != nil {
+				return fmt.Errorf("failed to marshal config: %w", err)
+			}
+
+			// write data to a manager.yaml file, with 0777 file permission
```
BinaryFissionGames, line 149:
```suggestion
			// write data to a manager.yaml file, with 0600 file permission
```

cmd/collector/main_test.go
```diff
@@ -0,0 +1,77 @@
...
+	require.ErrorIs(t, err, os.ErrNotExist)
+
+	tmp := "\000"
+	err = checkManagerConfig(&tmp)
+	require.Error(t, err)
+}
+
+func TestCheckManagerConfigNoFile(t *testing.T) {
+	exec.Command("rm", "-r", "./manager.yaml").Run()
+	manager := "./manager.yaml"
```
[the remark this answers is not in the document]
cpheps: Same comment on the other tests that have path of `./manager.yaml`

cmd/collector/main.go
cpheps: Yeah I would agree with that. The very first `os.LookupEnv` for `endpointENV` should probably be what @BinaryFissionGames suggested. This is more idiomatic Go with avoiding nesting by returning early.

cpheps (approved):

cmd/collector/main.go
```diff
@@ -21,12 +21,24 @@ import (
...
 	"github.com/observiq/observiq-otel-collector/internal/version"
+	"github.com/observiq/observiq-otel-collector/opamp"
 	"github.com/spf13/pflag"
 	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	// env variable name constants
+	endpointENV  = "OPAMP_ENDPOINT"
```
StefanKurek, line 37: nit: should this not be uppercase?

StefanKurek (approved):

cmd/collector/main.go
cpheps: For Go if there's an abbreviation in a variable name it should be fully capitalized. ENV could probably go either way so if we want to change it to `Env` that's fine.

cmd/collector/main_test.go
```diff
@@ -0,0 +1,76 @@
...
+	dat, err := os.ReadFile(manager)
+	out := &opamp.Config{}
+	err = yaml.Unmarshal(dat, out)
+	require.Equal(t,
+		&opamp.Config{
+			Endpoint: "0.0.0.0",
+		},
+		&opamp.Config{
+			Endpoint: out.Endpoint,
+		})
```
BinaryFissionGames, line 65: Can you test that the other fields are set as well here? Including the labels, agent name, secret key, and agent ID.

PR #537, opened 2022-06-30
deps: Updated OpAMP lib to v0.2.0

cpheps:
### Proposed Change
- Updated `github.com/open-telemetry/opamp-go` to v0.2.0
- Updated Callbacks to use newer `OnMessageFunc`
- Updated mocks and interfaces as required

Leaving in draft until I can test with Bindplane

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/opamp/observiq/observiq_client.go
+++ b/opamp/observiq/observiq_client.go
@@ -149,17 +149,13 @@ func (c *Client) Connect(ctx context.Context) error {
 			OnConnectFunc:          c.onConnectHandler,
 			OnConnectFailedFunc:    c.onConnectFailedHandler,
 			OnErrorFunc:            c.onErrorHandler,
-			OnRemoteConfigFunc:     c.onRemoteConfigHandler,
+			OnMessageFunc:          c.onMessageFuncHandler,
 			GetEffectiveConfigFunc: c.onGetEffectiveConfigHandler,
-			// Unimplemented Handles
-			// SaveRemoteConfigStatusFunc:
+			// Unimplemented handlers
 			// OnOpampConnectionSettingsFunc
 			// OnOpampConnectionSettingsAcceptedFunc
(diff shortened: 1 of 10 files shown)


opamp/observiq/observiq_client.go
```diff
@@ -193,16 +189,45 @@ func (c *Client) onErrorHandler(errResp *protobufs.ServerErrorResponse) {
...
 
-func (c *Client) onRemoteConfigHandler(_ context.Context, remoteConfig *protobufs.AgentRemoteConfig) (*protobufs.EffectiveConfig, bool, error) {
+func (c *Client) onMessageFuncHandler(ctx context.Context, msg *types.MessageData) {
+	switch {
+	// Fill in other message types as needed
+	case msg.RemoteConfig != nil:
+		if err := c.onRemoteConfigHandler(ctx, msg.RemoteConfig); err != nil {
+			c.logger.Error("Error while processing Remote Config Change", zap.Error(err))
+		}
+	}
```
BinaryFissionGames, line 199: This doesn't really matter for this PR since we only have one message type we respond to, but should this be an if statement instead? 

Not sure if each message can only provide one "sub-message", or if it can have several at once, if that makes sense.

cpheps: That is a good question...I think we likely need to ask the BP team. If it's only one at a time then the switch makes sense if it's multiple then we need if statements. The BP team is still working on implementing this so I'll ask them.

StefanKurek (approved):

opamp/observiq/observiq_client.go
andykellr: It looks like you changed it, but it's definitely possible to receive several messages.

BinaryFissionGames (approved):

PR #538, opened 2022-06-30
feat: Add tarball download + unarchiving to updater

BinaryFissionGames:
### Proposed Change
* Add ability to download archive
* Add ability to verify the archive matches passed in content-hash
  * This can probably be simplified when this is moved to the collector, since it won't have to be converted to hex
* Add ability to unarchive the files into specified folder

This is in the updater right now, but will likely be moved into the collector with some minor adjustments.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


StefanKurek (approved):
Overall looks pretty good. Maybe I'd move some of the magic error strings to constants, I think that's about it though.

updater/cmd/updater/main.go
```diff
@@ -16,14 +16,20 @@ package main
...
 	"github.com/spf13/pflag"
 )
 
 // Unimplemented
 func main() {
-	var showVersion = pflag.BoolP("version", "v", false, "prints the version of the collector")
+	var showVersion = pflag.BoolP("version", "v", false, "Prints the version of the collector and exits, if specified.")
+	var downloadURL = pflag.String("url", "", "URL to download the update archive from.")
+	var tmpDir = pflag.String("tmpdir", "", "Temporary directory for artifacts. Parent of the 'rollback' directory.")
+	var contentHash = pflag.String("content-hash", "", "Hex encoded hash of the content at the specified URL.")
```
cpheps, line 32: From another comment this is a hex encoded hash of a sha256? Does it come like that from BP or are we hex encoding it?

BinaryFissionGames: We would hex encode it before passing it to the updater binary.

That being said, when this is moved to the collector, I assume it'll be received as a `[]byte`, and we can remove this encoding/decoding step.

cpheps: Ok, I realize this is a moot question because we're moving it to the collector but curious why we decided to hex encode it?

cpheps (approved):

updater/cmd/updater/main.go
BinaryFissionGames: Testability, basically. I can run `sha256sum` and input the output of that tool into the cli, and have it work. 

I also wasn't sure if passing arbitrary bytes into args could run into any issues, but maybe that isn't a concern. Curious if it actually does cause issues.

PR #539, opened 2022-06-30
feat(plugin) Cassandra Logs

jsirianni:
### Proposed Change

Ported Stanza Cassandra plugin. Tested system, gc, and debug logs.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


cpheps:
@jsirianni Do you have sample output you could post either from the logging exporter or GCP?

cpheps (approved):

plugins/apache_cassandra_logs.yaml
```diff
@@ -0,0 +1,110 @@
...
+      attributes:
+        log_type: 'cassandra.debug'
+      operators:
+        - type: regex_parser
+          regex: '(?P<severity>[A-Z]+)\s+\[(?P<type>[^\]]+)\]\s+(?P<timestamp>\d{4}-\d{2}-\d{2}\s+\d{2}:\d{2}:\d{2},\d{3})\s+(?P<message>.*\S)'
+          severity:
+            parse_from: attributes.severity
+          timestamp:
+            parse_from: attributes.timestamp
+            layout: '%F %T,%L'
```
BinaryFissionGames, line 73: It doesn't look like these logs and the system logs specify a time offset; We should probably add a timezone parameter for these timestamp parsers

jsirianni: Done

BinaryFissionGames (approved):

PR #550, opened 2022-07-07
feat: Added OpAMP PackageStatuses functionality & basic response to PackagesAvailable

StefanKurek:
DRAFT: 98% ready. Need to nail down where the package statuses yaml will live. Also need to test against BindPlane environment.

### Proposed Change
Added OpAMP PackageStatuses functionality & basic response to PackagesAvailable. Should respond to OpAMP server's request for Package Status info and also will send back correct "initial" responses for PackagesAvailable (without actually starting installation).

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


opamp/observiq/observiq_client.go
```diff
@@ -229,6 +238,91 @@ func (c *Client) onRemoteConfigHandler(ctx context.Context, remoteConfig *protob
...
+	var lastPkgStatusMap map[string]*protobufs.PackageStatus
+	if lastPackageStatuses != nil {
+		lastPkgStatusMap = lastPackageStatuses.GetPackages()
+	}
+
+	curPackages := c.createPackageStatusMap(packagesAvailable.GetPackages(), lastPkgStatusMap)
+	curPackageStatuses.Packages = curPackages
+
+	if err = c.packagesStateProvider.SetLastReportedStatuses(curPackageStatuses); err != nil {
+		c.logger.Debug("Failed to save last reported package statuses", zap.Error(err))
```
cpheps, line 268: Should this return an error? Also, this should probably be a warning log as it would be good to know if this is happening.

opamp/observiq/observiq_client.go
```diff
@@ -229,6 +238,91 @@ func (c *Client) onRemoteConfigHandler(ctx context.Context, remoteConfig *protob
...
+		Packages:                      map[string]*protobufs.PackageStatus{},
+	}
+
+	// Retrieve last known status (this should return with minimal info even on first time)
+	lastPackageStatuses, err := c.packagesStateProvider.LastReportedStatuses()
+
+	// If there is a problem retrieving the last saved PackageStatuses, we will log the error
+	// but continue on as the only thing missing will be the agent package hash.
+	if err != nil {
+		c.logger.Debug("Failed to retrieve last reported package statuses", zap.Error(err))
```
cpheps, line 256: This should be a warning log as it would be useful to know why this is happening.

opamp/observiq/observiq_client.go
```diff
@@ -229,6 +238,91 @@ func (c *Client) onRemoteConfigHandler(ctx context.Context, remoteConfig *protob
...
+		switch name {
+		// If it's an expected package, return an installing status
+		case "observiq-otel-collector":
+			var agentHash []byte
+			if lastPkgStatusMap != nil && lastPkgStatusMap[name] != nil {
+				if lastPkgStatusMap[name].GetAgentHasVersion() != version.Version() {
+					c.logger.Debug(fmt.Sprintf(
+						"Version: %s and last reported package status version: %s differ",
+						version.Version(),
+						lastPkgStatusMap[name].GetAgentHasVersion()))
```
cpheps, line 294: Is this a place holder to where we would kick off update?

opamp/observiq/observiq_client.go
```diff
@@ -229,6 +238,91 @@ func (c *Client) onRemoteConfigHandler(ctx context.Context, remoteConfig *protob
...
+				pkgStatusMap[name].Status = protobufs.PackageStatus_Installing
+			}
+		// If it's not an expected package, return a failed status
+		default:
+			pkgStatusMap[name] = &protobufs.PackageStatus{
+				Name:                 name,
+				ServerOfferedVersion: availPkg.GetVersion(),
+				ServerOfferedHash:    availPkg.GetHash(),
+				Status:               protobufs.PackageStatus_InstallFailed,
+				ErrorMessage:         fmt.Sprintf("Package %s not allowed", name),
```
cpheps, line 318: I would maybe change this message to `Package %s not supported`. `Allowed` make it seem like a permissions error.

opamp/observiq/observiq_packages_state_provider.go
```diff
@@ -0,0 +1,174 @@
...
+
+	return errors.New("method not implemented: PackageStateProvider DeletePackage")
+}
+
+// LastReportedStatuses retrieves the PackagesStatuses from a saved yaml file
+func (p *PackagesStateProvider) LastReportedStatuses() (*protobufs.PackageStatuses, error) {
+	p.logger.Debug("Retrieve last reported package statuses")
+
+	packages := map[string]*protobufs.PackageStatus{
+		"observiq-otel-collector": {
```
cpheps, line 122: Should this string `observiq-otel-collector` be a const somewhere? I thought I saw we use this in the client on lookup.

opamp/observiq/observiq_packages_state_provider.go
```diff
@@ -0,0 +1,174 @@
...
+			AgentHasVersion: version.Version(),
+			Status:          protobufs.PackageStatus_Installed,
+		},
+	}
+	packageStatuses := &protobufs.PackageStatuses{
+		Packages: packages,
+	}
+
+	// If there's a problem reading the file, we just return a barebones PackageStatuses.
+	if _, err := os.Stat(p.yamlPath); errors.Is(err, os.ErrNotExist) {
```
cpheps, line 133: If this errors for a different reason do we not handle it as it'll be picked up in the following `ReadFile` call?

opamp/observiq/observiq_packages_state_provider.go
```diff
@@ -0,0 +1,174 @@
...
+		p.logger.Debug("Package statuses yaml doesn't exist")
```
cpheps, line 134: Is this ok if it doesn't exist? Wondering if this should be a warning message if it shouldn't happen.

opamp/observiq/observiq_packages_state_provider.go
```diff
@@ -0,0 +1,174 @@
...
+		return packageStatuses, nil
+	}
+
+	statusesBytes, err := os.ReadFile(p.yamlPath)
+	if err != nil {
+		return packageStatuses, fmt.Errorf("problem reading from package statuses yaml: %w", err)
```
cpheps, line 140: nit: every other error starts with `failed to X` might be worth keeping this error consistent.

opamp/observiq/observiq_packages_state_provider_test.go
```diff
@@ -0,0 +1,550 @@
...
+				assert.Equal(t, "", actual.Packages[pkgName].GetServerOfferedVersion())
+				assert.Nil(t, actual.Packages[pkgName].GetServerOfferedHash())
+				assert.Equal(t, protobufs.PackageStatus_Installed, actual.Packages[pkgName].GetStatus())
+				assert.Equal(t, "", actual.Packages[pkgName].GetErrorMessage())
+			},
+		},
+		{
+			desc: "Problem reading file",
+			testFunc: func(*testing.T) {
+				cantReadYaml := "../testdata/noread.yaml"
```
cpheps, line 320: So you don't have to remove the written file you can create a temp dir that gets cleaned up at the end of the test by using [t.TempDir](https://pkg.go.dev/testing#T.TempDir).

Then creating the path by using `filepath.Join` so it'll format the path based on the OS running the tests

```go
tmpDir := t.TempDir()
can'tReadYaml := filepath.Join(tmpDir, "noread.yaml")
```

cpheps:
@StefanKurek for

> Need to nail down where the package statuses yaml will live.

I would create it next to the binary so just `./package_status.yaml`. I don't think the user needs to be aware of this file and we should have the permissions to add/modify files within our working directory which is almost always our install directory.

opamp/observiq/observiq_client.go
```diff
@@ -229,6 +238,91 @@ func (c *Client) onRemoteConfigHandler(ctx context.Context, remoteConfig *protob
...
+		return fmt.Errorf("failed to set package statuses: %w", err)
+	}
+
+	return nil
+}
+
+func (c *Client) createPackageStatusMap(
+	pkgAvailMap map[string]*protobufs.PackageAvailable,
+	lastPkgStatusMap map[string]*protobufs.PackageStatus) map[string]*protobufs.PackageStatus {
+	pkgStatusMap := map[string]*protobufs.PackageStatus{}
```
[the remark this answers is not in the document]
StefanKurek: @cpheps unless you feel pretty strongly about this, i think i'll keep them as is. they're equivalent, there's 4 fewer characters and...it looks nicer to me this way haha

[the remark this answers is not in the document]
cpheps: Yeah I don't feel super strong just more of an FYI. `make` could be useful too if you want to preset the size of the map but I don't think we need that here.

opamp/observiq/observiq_client.go
StefanKurek: @cpheps yeah to some degree. this built status implies that an update would actually be started (but i'm not implementing any of that with this story. just handling the PackageStatuses messages...which i know ultimately won't fully make sense without the update actually occurring)

opamp/observiq/observiq_client.go
StefanKurek: 👍

opamp/observiq/observiq_packages_state_provider_test.go
StefanKurek: @cpheps awesome. good to know

opamp/observiq/observiq_packages_state_provider.go
```diff
@@ -0,0 +1,174 @@
...
+
+// SetLastReportedStatuses saves the given PackageStatuses into a yaml file
+func (p *PackagesStateProvider) SetLastReportedStatuses(statuses *protobufs.PackageStatuses) error {
+	p.logger.Debug("Set last reported package statuses")
+
+	data, err := yaml.Marshal(statuses)
+	// If there is any problem saving the new package statuses, make sure that we delete any existing file
+	// in order to not have outdated data as its better to start fresh (is this the best solution?)
+	if err != nil {
+		if err := os.Remove(p.yamlPath); err != nil {
```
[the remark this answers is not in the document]
StefanKurek: @cpheps good thinking. i'll make the change

opamp/observiq/observiq_packages_state_provider.go
```diff
@@ -0,0 +1,174 @@
...
+		return packageStatuses, nil
+	}
+
+	statusesBytes, err := os.ReadFile(p.yamlPath)
+	if err != nil {
+		return packageStatuses, fmt.Errorf("problem reading from package statuses yaml: %w", err)
+	}
+
+	var readPackageStatuses *protobufs.PackageStatuses
+	if err := yaml.Unmarshal(statusesBytes, &readPackageStatuses); err != nil {
```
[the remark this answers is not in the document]
StefanKurek: @cpheps nope, PackageStatuses is a struct

opamp/observiq/observiq_packages_state_provider.go
StefanKurek: @cpheps yeah, this should be a constant. i'll do that

opamp/observiq/observiq_packages_state_provider.go
StefanKurek: @cpheps We could make it a warning, but it is guaranteed to happen at least once. I could go either way. If you feel strongly I'll change it.

opamp/observiq/observiq_packages_state_provider.go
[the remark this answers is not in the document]
cpheps: Ok in that case you can either remove the `*` and not make `readPackageStatuses` a pointer or remove the `&` in marshal. Right now you have a pointer to a pointer to the struct (which might actually work in the yaml package depending on if it recurses through the reflection).

opamp/observiq/observiq_packages_state_provider.go
cpheps: Ok if it'll happen at least once then that's fine to keep it as debug.

opamp/observiq/observiq_packages_state_provider.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	return packageStateProvider
+}
+
+// AllPackagesHash not implemented so returns an error with this info
+func (p *PackagesStateProvider) AllPackagesHash() ([]byte, error) {
+	p.logger.Debug("Retrieve all packages hash")
+
+	return nil, errors.New("method not implemented: PackageStateProvider AllPackagesHash")
```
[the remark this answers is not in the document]
cpheps: Talked about this outside in slack.

opamp/observiq/observiq_packages_state_provider.go
```diff
@@ -0,0 +1,172 @@
...
+
+	return readPackageStatuses, nil
+}
+
+// SetLastReportedStatuses saves the given PackageStatuses into a yaml file
+func (p *PackagesStateProvider) SetLastReportedStatuses(statuses *protobufs.PackageStatuses) error {
+	p.logger.Debug("Set last reported package statuses")
+
+	// If there is any problem saving the new package statuses, make sure that we delete any existing file
+	// in order to not have outdated data as its better to start fresh (is this the best solution?)
```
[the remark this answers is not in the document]
StefanKurek: @cpheps yeah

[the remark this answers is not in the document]
cpheps: Is there any negative effect to removing this file? 

If we remove it and fail to set it then we don't have a file and we'll just write one on the next call right?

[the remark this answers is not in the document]
StefanKurek: @cpheps shouldn't be. yeah in theory we just write a new one

opamp/observiq/observiq_packages_state_provider.go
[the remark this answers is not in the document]
StefanKurek: @cpheps this was confusing to me as i copied this from code elsewhere.

so interestingly enough if i remove the * and make it not a pointer, things are fine. if i keep it a pointer and try to remove the & inside `Unmarshal`, then my test give me a weird failure.

opamp/observiq/observiq_client.go
```diff
@@ -195,6 +200,10 @@ func (c *Client) onMessageFuncHandler(ctx context.Context, msg *types.MessageDat
 		if err := c.onRemoteConfigHandler(ctx, msg.RemoteConfig); err != nil {
 			c.logger.Error("Error while processing Remote Config Change", zap.Error(err))
 		}
+	} else if msg.PackagesAvailable != nil {
```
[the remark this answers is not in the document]
StefanKurek: @cpheps yeah, i knew that. i'm just being an idiot here.

opamp/observiq/observiq_packages_state_provider.go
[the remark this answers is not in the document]
cpheps: Ok, I think it's safe to remove this question unless you think there's some part of it we've missed.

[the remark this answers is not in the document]
StefanKurek: yeah that's fine. i'll remove it.

opamp/observiq/observiq_packages_state_provider.go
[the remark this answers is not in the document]
StefanKurek: We discussed this and I'm just going to make the variable a non pointer

cpheps (approved):

opamp/testdata/package_statuses_good.yaml
```diff
@@ -0,0 +1,25 @@
+packages:
```
BinaryFissionGames, line 1: Would it make more sense to put these under a `observiq` package (e.g. `opmap/observiq/testdata` instead of `opamp/testdata`)? Since it's only being used in that package.

opamp/observiq/observiq_client.go
```diff
@@ -39,13 +39,17 @@ var (
 // Ensure interface is satisfied
 var _ opamp.Client = (*Client)(nil)
 
+// MainPackageName is the name for the top level packages for this collector
+const MainPackageName = "observiq-otel-collector"
```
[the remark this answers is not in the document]
StefanKurek: that was just a brain fart thinking that constants should be capitalized in go. changing this.

opamp/observiq/observiq_packages_state_provider.go
```diff
@@ -0,0 +1,170 @@
...
+var _ types.PackagesStateProvider = (*PackagesStateProvider)(nil)
+
+// PackagesStateProvider represents a PackagesStateProvider which uses a yaml file to persist PackageStatuses
+type PackagesStateProvider struct {
+	yamlPath string
+	logger   *zap.Logger
+}
+
+// NewPackagesStateProvider creates a new OpAmp PackagesStateProvider
+func NewPackagesStateProvider(logger *zap.Logger, yamlPath string) types.PackagesStateProvider {
```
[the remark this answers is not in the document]
StefanKurek: couldn't hurt. we can always make it public later if need be. i'm fine changing it

opamp/testdata/package_statuses_good.yaml
StefanKurek: i'm a little on the fence for this one. i'm ok with having test files in one nice spot unless that spot gets unwieldy and needs to be split up. if you feel strongly about it though, let me know

BinaryFissionGames: I don't feel super strongly about it, it's just where I'd put it. Should be fine in that spot.

opamp/observiq/observiq_packages_state_provider_test.go
```diff
@@ -0,0 +1,499 @@
...
+)
+
+func TestNewPackagesStateProvider(t *testing.T) {
+	testCases := []struct {
+		desc     string
+		testFunc func(*testing.T)
+	}{
+		{
+			desc: "New PackagesStateProvider",
+			testFunc: func(*testing.T) {
```
BinaryFissionGames, line 40:
```suggestion
			testFunc: func(t *testing.T) {
```
I think this is a problem that's present in a lot (all?) of these tests; We should shadow the parent test, otherwise when it fails you get: 
```
testing.go:1169: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

StefanKurek: @BinaryFissionGames changed these

BinaryFissionGames (approved):

armstrmi (approved):
My understanding is somewhat limited, but from what I've seen it make sense.

opamp/observiq/observiq_packages_state_provider.go
```diff
@@ -156,7 +172,9 @@ func (p *packagesStateProvider) SetLastReportedStatuses(statuses *protobufs.Pack
 		p.logger.Debug("Failed to delete package statuses yaml", zap.Error(err))
 	}
 
-	data, err := yaml.Marshal(statuses)
+	states := packageStatusesToStates(statuses)
+
+	data, err := yaml.Marshal(*states)
```
cpheps, line 177: Do you need to dereference the pointer here? If so maybe we shouldn't return a pointer from this method.

BinaryFissionGames (approved):

PR #565, opened 2022-07-14
feat: Moved code to download, verify, and extract OpAMP package file from updater to collector

StefanKurek:
### Proposed Change
Moved and refactored the code to download, verify, and extract an OpAMP PackagesAvailable file from the Updater to the collector in the PackagesAvailable handler after sending the initial "Installing" status message.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


go.mod
```diff
@@ -373,13 +377,15 @@ require (
...
 	github.com/uber/jaeger-lib v2.4.1+incompatible // indirect
+	github.com/ulikunitz/xz v0.5.9 // indirect
 	github.com/vishvananda/netlink v1.1.1-0.20210330154013-f5de75959ad5 // indirect
 	github.com/vishvananda/netns v0.0.0-20210104183010-2eb08e3e575f // indirect
 	github.com/vmware/govmomi v0.28.0 // indirect
 	github.com/vultr/govultr/v2 v2.17.0 // indirect
 	github.com/xdg-go/pbkdf2 v1.0.0 // indirect
 	github.com/xdg-go/scram v1.1.1 // indirect
 	github.com/xdg-go/stringprep v1.0.3 // indirect
+	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
```
cpheps, line 388: This library is causing a licensing issue. Guessing it's a dependency of `github.com/mholt/archiver`.

BinaryFissionGames (approved):

go.mod
StefanKurek: @cpheps added an exception as we have used this dependency in the past and the license says we can do whatever we want with the code.

PR #566, opened 2022-07-14
fix: Resolved overwriting hostnames

jmwilliams89:
### Proposed Change
- Removes the host detection processor
- Manually appends hostname to telemetry if not already present
- Adds a configuration option for disabling hostname discovery
- Updates documentation

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/googlecloudexporter/exporter.go
+++ b/exporter/googlecloudexporter/exporter.go
@@ -17,16 +17,23 @@ package googlecloudexporter
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/pdata/plog"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/collector/pdata/ptrace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
 )
+
+// hostname is the name of the current host
+var hostname, _ = os.Hostname()
 
 // exporter is a google cloud exporter wrapped with additional functionality
 type exporter struct {
+	appendHost bool
+
 	metricsProcessors []component.MetricsProcessor
 	metricsExporter   component.MetricsExporter
 	metricsConsumer   consumer.Metrics
@@ -42,6 +49,7 @@ type exporter struct {
 
 // ConsumeMetrics consumes metrics
 func (e *exporter) ConsumeMetrics(ctx context.Context, md pmetric.Metrics) error {
+	e.appendMetricAttrs(&md)
 	if e.metricsConsumer == nil {
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


exporter/googlecloudexporter/exporter.go
```diff
@@ -17,16 +17,23 @@ package googlecloudexporter
...
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/pdata/plog"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/collector/pdata/ptrace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
 )
 
+// hostname is the name of the current host
+var hostname, _ = os.Hostname()
```
cpheps, line 31: We're ignoring the error here. Are we ok with an empty string if this fails? This shouldn't fail but it's possible.

cpheps (approved):

exporter/googlecloudexporter/exporter.go
```diff
@@ -150,3 +160,39 @@ func (e *exporter) Shutdown(ctx context.Context) error {
...
+// appendLogAttrs appends log attributes if not present
+func (e *exporter) appendLogAttrs(ld *plog.Logs) {
+	for i := 0; i < ld.ResourceLogs().Len(); i++ {
+		resourceAttrs := ld.ResourceLogs().At(i).Resource().Attributes()
+		_, hostNameExists := resourceAttrs.Get(string(semconv.HostNameKey))
+		_, hostIDExists := resourceAttrs.Get(string(semconv.HostIDKey))
+		if !hostNameExists && !hostIDExists && e.appendHost {
+			resourceAttrs.InsertString(string(semconv.HostNameKey), hostname)
+		}
+	}
```
BinaryFissionGames, line 185: if e.appendHost is false, I think we should bail out before the loop, since we won't ever add the attribute.

exporter/googlecloudexporter/factory.go
```diff
@@ -106,12 +105,10 @@ func createLogsExporter(ctx context.Context, set component.ExporterCreateSetting
 	processors := []component.LogsProcessor{}
 	processorConfigs := []config.Processor{
 		exporterConfig.BatchConfig,
-		exporterConfig.DetectConfig,
 	}
 
 	processorFactories := []component.ProcessorFactory{
 		batchprocessor.NewFactory(),
-		resourcedetectionprocessor.NewFactory(),
```
StefanKurek, line 114: @jmwilliams89 can you help explain to me why this is removed from metrics and logs but not traces?

jmwilliams89: Awesome catch. That needs to be removed from all of them.

BinaryFissionGames (approved):

PR #568, opened 2022-07-18
feat: Kubelet and Container Logs Plugins

cpheps:
### Proposed Change
- Created kubelet and container logs plugins

Kubelet Parsed Logs:
kubelet_logs_plugin

Container Parsed Logs:
continer_logs_plugin

This has been tested by @jsirianni in k8s.

##### Checklist
- [X] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/container_logs.yaml
+++ b/plugins/container_logs.yaml
@@ -1,0 +1,115 @@
+version: 0.0.1
+title: Kubernetes Container Logs
+description: Log parser for Container logs in a Kubernetes pod
+parameters:
+  - name: log_paths
+    type: "[]string"
+    description: A list of file glob patterns that match the file paths to be read
+    default:
+      - "/var/log/containers/*.log"
+  - name: exclude_file_log_path
+    type: "[]string"
+    description: A list of file glob patterns to exclude from reading
+    default:
+      - "/var/log/containers/observiq-*-collector-*"
+  - name: start_at
+    type: string
+    description: At startup, where to start reading logs from the file.
+    supported:
+      - beginning
+      - end
+    default: end
+
+template: |
+  receivers:
+    filelog:
+        include:
+          {{ range $fp := .log_paths }}
+            - '{{ $fp }}'
+          {{end}}
+        start_at: {{ .start_at }}
+        exclude:
+          {{ range $fp := .exclude_log_paths }}
+            - '{{ $fp }}'
+          {{end}}
+        poll_interval: 500ms
+        operators:
+          # Support docker and containerd runtimes, which have different
+          # logging formats.
+          - type: router
+            routes:
+              - expr: 'body matches "^[^\\s]+ \\w+ .*"'
+                output: containerd_parser
+            default: docker_parser
+
+          # The raw message looks like this:
+          # {"log":"I0618 14:30:29.641678       1 logs.go:59] http: TLS handshake error from 192.168.49.2:56222: EOF\n","stream":"stderr","time":"2022-06-18T14:30:29.641732743Z"}
+          - type: json_parser
+            id: docker_parser
+            timestamp:
+              parse_from: attributes.time
+              layout: '%Y-%m-%dT%H:%M:%S.%sZ'
+            output: log-to-body
+
+          # The raw message looks like this:
+          # 2022-06-18T16:52:59.639114537Z stdout F {"message":"registered Stackdriver tracing","severity":"info","timestamp":"2022-06-18T16:52:59.639034532Z"}
+          - id: containerd_parser
+            type: regex_parser
+            regex: '^(?P<time>[^\s]+) (?P<stream>\w+) (?P<partial>\w)?(?P<log>.*)'
+          - type: recombine
+            source_identifier: attributes["log.file.name"]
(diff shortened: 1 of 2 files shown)


plugins/kubelet_logs.yaml
```diff
@@ -0,0 +1,60 @@
+version: 0.0.1
+title: Kubernetes Kubelet
+description: Log parser for Kubelet journald logs
+parameters:
+  - name: journald_directory
+    description: Directory containing journal files to read entries from
```
jsirianni, line 6: I think we should mention that this is optional, and the receiver will attempt to detect the correct directory.

plugins/kubelet_logs.yaml
```diff
@@ -0,0 +1,60 @@
...
+          - type: regex_parser
+            regex: '(?P<severity>\w)(?P<timestamp>\d{4} \d{2}:\d{2}:\d{2}.\d+)\s+(?P<pid>\d+)\s+(?P<src>[^:]*):(?P<src_line>[^\]]*)\] (?P<message>.*)'
+            severity:
+              parse_from: attributes.severity
+              mapping:
+                debug: d
+                info: i
+                warning: w
+                error: e
+                critical: c
```
BinaryFissionGames, line 44: Looks like some of these don't line up with aliases recognized by OTel, maybe it should be something like this?
```suggestion
                debug: d
                info: i
                warn: w
                error: e
                fatal: c
```

plugins/kubelet_logs.yaml
```diff
@@ -0,0 +1,60 @@
...
+            timestamp:
+              parse_from: attributes.timestamp
+              layout: '%m%d %H:%M:%S.%s'
```
BinaryFissionGames, line 47: This timestamp doesn't have any time offset, I think we might want to add a timezone param (unless we think it's fixed to UTC, in which case we should hardcode to that).

plugins/container_logs.yaml
```diff
@@ -0,0 +1,115 @@
+version: 0.0.1
+title: Kubernetes Container Logs
+description: Log parser for Container logs in a Kubernetes pod
```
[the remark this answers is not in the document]
cpheps: Cool I'll make it more generalized.

plugins/container_logs.yaml
```diff
@@ -0,0 +1,115 @@
...
+  - name: start_at
+    type: string
+    description: At startup, where to start reading logs from the file.
+    supported:
+      - beginning
+      - end
+    default: end
+
+template: |
+  receivers:
```
BinaryFissionGames, line 24: Looks like we are missing a `log_type` attribute on this receiver

plugins/container_logs.yaml
```diff
@@ -0,0 +1,115 @@
...
+  - name: log_paths
+    type: "[]string"
+    description: A list of file glob patterns that match the file paths to be read
+    default:
+      - "/var/log/containers/*.log"
+  - name: exclude_file_log_path
+    type: "[]string"
+    description: A list of file glob patterns to exclude from reading
+    default:
+      - "/var/log/containers/observiq-*-collector-*"
```
BinaryFissionGames, line 14: Will this pattern always match when our collector is deployed, or should we call out that this might vary in some way (e.g. different container or pod name)? I forget how this path is constructed.

cpheps: I'm not sure @jsirianni do you happen to know?

plugins/kubelet_logs.yaml
cpheps: @jsirianni Do you happen to know if it always uses UTC?

jsirianni: Hmm yeah. Generally it will be UTC but we have seen other timezones in use. This would be the timezone set on the host operating system.

cpheps: I can add it, defaulting to UTC

plugins/container_logs.yaml
BinaryFissionGames: Cool, can we add into the description that the default is for deploying with the operator?

plugins/container_logs.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+template: |
+  receivers:
+    filelog:
+        include:
+          {{ range $fp := .log_paths }}
+            - '{{ $fp }}'
+          {{end}}
+        start_at: {{ .start_at }}
+        log_type: "k8s.container_logs"
```
BinaryFissionGames, line 31: I think this needs to be nested under attributes (https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver)
```suggestion
        attributes:
            log_type: "k8s.container_logs"
```
Something we need to figure out with our plugin receiver is how to make it more robust against these types of configs; It just drops fields if they aren't valid options instead of giving an error.

cpheps: Yeah you're right. went too fast.

BinaryFissionGames (approved):

jsirianni (approved):

PR #570, opened 2022-07-18
feat: Add mutex for updatingClient flag in client

StefanKurek:
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


armstrmi (approved):
I have a somewhat limited understanding of what exactly is going on, but it LGTM for the most part and changes are minor

PR #578, opened 2022-07-19
feat: Added timezone parameter type

dmikolay:
### Proposed Change
- Added timezone parameter type for plugins 
- Made appropriate changes to plugin.go tests
- Changed tomcat plugin to reflect changes and tested it

##### Checklist
- [X] Changes are tested
- [ ] CI has passed


plugins/tomcat_logs.yaml
```diff
@@ -23,9 +23,8 @@ parameters:
...
+    type: timezone
```
cpheps, line 26: Can we bump the version of this file to v0.1.0 since we made changes?

receiver/pluginreceiver/plugin.go
```diff
@@ -173,12 +173,11 @@ func (p *Plugin) checkType(values map[string]interface{}) error {
...
 			raw, ok := value.(string)
 			if !ok {
 				return fmt.Errorf("parameter %s must be a string", parameter.Name)
 			}
-			for _, tz := range tzlist {
+			for tz := range tzlist {
```
cpheps, line 180: This should be a lookup not a range

So this:

```go
if _, ok := tzlist[raw]; ok {
  return nil
} 
```

BinaryFissionGames:
Can we add this as a support type to the docs?
This line here: https://github.com/observIQ/observiq-otel-collector/blob/65eaa693536187c78a3a68aac84cfdc5c328f967/receiver/pluginreceiver/README.md?plain=1#L69

I think ideally we'd have a separate markdown document with a list of all available locations, but we should at least document that this is `timezone` supported here.

cpheps (approved):

BinaryFissionGames (approved):

PR #580, opened 2022-07-20
feat: Changed instances of timezone as a string to the new timezone type

dmikolay:
### Proposed Change
- Converted all plugins that had the timezone parameter specified as a string to the new timezone type
- Changed those plugins to version 0.1.0
- Tested on a few of the plugins that were changed, all of which worked

##### Checklist
- [X] Changes are tested
- [X] CI has passed


armstrmi (approved):
LGTM

PR #581, opened 2022-07-21
Logic for Updater to monitor Collector Status

cpheps:
### Proposed Change
- Set default package name as const in `packagestate` module
- Added logic for Updater to monitor installing collector.

Updater will monitor the `package_status.json` file in the install directory with the following logic:
- File does not exist on updater startup -> exit with error
- Installation fails on updater side -> mark file with failed install and exit with error
- Status is set to failed after new installation started -> rollback and exit with error
- Status is not set for 10s after new installations started -> rollback, mark file with failed, and exit with error
- Status is set to successful -> exit

**Note:** Monitoring logic has been unit tested but not in the `main` function yet as the rollback logic is still under work in the branch being merged to.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


cpheps:
Linting may fail as there are some linting errors in the `updater-rollback` branch that this was based off of. I didn't want to fix those as that branch is still under development.

updater/internal/state/monitor.go
```diff
@@ -0,0 +1,139 @@
...
+			packageStatus, err := c.stateManager.LoadStatuses()
+			switch {
+			case errors.Is(err, os.ErrNotExist):
+				// File may not exist for a short period of time while being written so just continue
+				continue
+			case err != nil:
+				return fmt.Errorf("error while loaded status: %w", err)
+			default:
+				targetPackage, ok := packageStatus.GetPackages()[packageName]
+				// Target package might not exist yet so continue
```
StefanKurek, line 121: @cpheps how could the target package not exist yet in the status file?

updater/cmd/updater/main.go
```diff
@@ -60,11 +74,35 @@ func main() {
 
 	if err := installer.Install(rb); err != nil {
 		log.Default().Printf("Failed to install: %s", err)
-		// TODO: Change packagestatuses to have error message
+
+		// Set the state to failed before rollback so collector knows it failed
+		if setErr := monitor.SetState(packagestate.DefaultFileName, protobufs.PackageStatus_InstallFailed, err); setErr != nil {
```
StefanKurek, line 79: @cpheps it would be great if you could fill out the error message too so the collector has that to send to the Server as well.

updater/cmd/updater/main.go
```diff
@@ -60,11 +74,35 @@ func main() {
...
+	defer cancel()
+
+	// Monitor the install state
+	if err := monitor.MonitorForSuccess(checkCtx, packagestate.DefaultFileName); err != nil {
+		log.Println("Failed to install:", err)
+
+		// If this is not an error due to the collector setting a failed status we need to set a failed status
+		if !errors.Is(err, state.ErrFailedStatus) {
+			// Set the state to failed before rollback so collector knows it failed
+			if setErr := monitor.SetState(packagestate.DefaultFileName, protobufs.PackageStatus_InstallFailed, err); setErr != nil {
```
StefanKurek, line 97: @cpheps again, would be great if we set the error message here as well

updater/internal/state/monitor.go
cpheps: This is kind of future proofing. This shouldn't happen but I'm protecting against it just in case.

updater/cmd/updater/main.go
cpheps: This does fill out the error message because the error passed in is not nil. Unless I'm missing what you're saying.

updater/cmd/updater/main.go
cpheps: Same as above it sets the err message because the passed in error is not nil.

updater/cmd/updater/main.go
StefanKurek: nope, i'm just dumb and totally missed that

updater/internal/state/monitor.go
StefanKurek: i mean, i guess it doesn't hurt anything right now so ok

cpheps: Yeah, mostly I had to do a map lookup and I didn't want to assume it would just be ok. I though retrying then hitting a timeout was more flexible than just failing.

updater/internal/state/monitor.go
```diff
@@ -0,0 +1,139 @@
...
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			packageStatus, err := c.stateManager.LoadStatuses()
+			switch {
+			case errors.Is(err, os.ErrNotExist):
+				// File may not exist for a short period of time while being written so just continue
+				continue
+			case err != nil:
+				return fmt.Errorf("error while loaded status: %w", err)
```
StefanKurek, line 118: could we get an error because of a partially written JSON file? should we always retry at least once or something?

BinaryFissionGames, line 118: This is where I'm really paranoid; Some scenarios that we need to consider:

1. (This code already handles this one)
  * Collector deletes status file (it does this on every write IIRC)
  * Updater attempts to read status file
  * File does not exist, updater gets error

2. Read while write
  * Collector deletes status file, gets halfway through writing the status file
  * Updater attempts to read status file
  * Updater gets unmarshal error
3. Write while read
  * Updater opens file for reading
  * Updater partially reads file
  * Collector completely writes file
  * Updater reads whole file
  * This could be an unmarshal error, OR it could be completely valid depending on what was written, ending up being with the incorrect data being read.

I'm not sure if this is overly paranoid, but I think we should 
a. Basically ignore any error (probably log it), because unmarshal errors could happen, and IDK if we'll get any other weird errors contesting the file on windows.
b. Make sure the returned packages are static for two cycles, to make sure we've read what was actually supposed to be written, and not some partial of old and new data.

Those are my thoughts here, let me know if that sounds overly paranoid.

cpheps: What do you mean partially written? We can just keep retrying I guess until timeout. There's no way to know what the cause is if we hit this state.

StefanKurek: @BinaryFissionGames i think this is all fair. 
i don't think the collector would ever write twice in a row during this process, so maybe just a single retry on error would be enough to handle 2. 
for 3. i think this would only happen if the collector detects a timeout on updater, and starts to modify status, but then the updater is trying to read at that point. this could be solved by simply making sure the collector kills the updater before writing in this case

updater/internal/state/monitor.go
```diff
@@ -111,11 +110,11 @@ func (c *CollectorMonitor) MonitorForSuccess(ctx context.Context, packageName st
...
 			switch {
-			case errors.Is(err, os.ErrNotExist):
-				// File may not exist for a short period of time while being written so just continue
-				continue
+			// If there is any error we'll just continue. Some valid reasons we could error and should retry:
+			// - File was deleted by new collector before it's rewritten
+			// - File is being written to while we're reading it so we'll get invalid JSON
 			case err != nil:
-				return fmt.Errorf("error while loaded status: %w", err)
+				continue
```
StefanKurek, line 117: should we still log the error?

StefanKurek (approved):

updater/internal/state/monitor.go
cpheps: There currently isn't a logger in this thing. Probably should add it though.

cpheps: Actually, the underlying state manager logs this error so we won't need to. That's why I didn't add a logger.

BinaryFissionGames (approved):

PR #589, opened 2022-07-26
feat: Updater logging

BinaryFissionGames:
### Proposed Change
* Change logging to be a zap.Logger
* Log to file
* Add flag to control log level (defaults to debug logging; I think updating is small and contained enough that this is OK for now)

The log automatically chooses ${INSTALL_DIR}/log/updater.log as the location to log, and isn't configurable. I figure this is OK for now, and something we can tack on later if we decide this needs to be configured.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


updater/cmd/updater/main.go
```diff
@@ -32,10 +34,10 @@ import (
 	"go.uber.org/zap"
 )
 
-// Unimplemented
 func main() {
-	var showVersion = pflag.BoolP("version", "v", false, "Prints the version of the collector and exits, if specified.")
+	var showVersion = pflag.BoolP("version", "v", false, "Prints the version of the updater and exits, if specified.")
 	var tmpDir = pflag.String("tmpdir", "", "Temporary directory for artifacts. Parent of the 'rollback' directory.")
+	var logLevelStr = pflag.String("loglevel", "debug", "Level with which to log")
```
cpheps, line 40: If the default is `debug` is there any real reason to specify this? Users can't hook into this and it's not easy to inject a value for ourselves. I think just having `debug` logging is fine.

updater/internal/logging/logging.go
```diff
@@ -0,0 +1,64 @@
...
+
+	return prodLogger, nil
+}
+
+func core(installDir string, level zapcore.Level) zapcore.Core {
+	logger := &lumberjack.Logger{
+		Filename:   path.LogFile(installDir),
+		MaxSize:    10,
+		MaxBackups: 3,
+	}
```
cpheps, line 47: We don't need log rotating in here can we remove this and reduce dependencies?

updater/internal/action/service_action.go
```diff
@@ -38,6 +39,10 @@ func (s ServiceStopAction) Rollback() error {
 	return s.svc.Start()
 }
 
+func (s ServiceStopAction) String() string {
+	return "ServiceStopAction{}"
+}
```
cpheps, line 44: I don't think we need this if we're trying to satisfy the `Stringer` interface. Without this i'll likely print something like  `&{}`. This isn't super helpful if we need to know the type of the action from the string but if we don't really care I would omit this.

updater/internal/install/install.go
```diff
@@ -60,31 +65,35 @@ func (i Installer) Install(rb rollback.ActionAppender) error {
...
 	rb.AppendAction(action.NewServiceStartAction(i.svc))
+	i.logger.Debug("Service started")
 
 	return nil
 }
 
 // copyFiles moves the file tree rooted at latestDirPath to installDirPath,
 // skipping configuration files. Appends CopyFileAction-s to the Rollbacker as it copies file.
-func copyFiles(inputPath, outputPath, tmpDir string, rb rollback.ActionAppender) error {
+func (i Installer) copyFiles(inputPath, outputPath, tmpDir string, rb rollback.ActionAppender) error {
```
cpheps, line 96: This is fine making it a receiver on the object but seems like if this is just for access to the logger could we just pass the logger in?

updater/internal/rollback/rollback.go
```diff
@@ -86,18 +91,20 @@ func (r Rollbacker) Backup() error {
...
 		if err := action.Rollback(); err != nil {
-			log.Default().Printf("Failed to run rollback option: %s", err)
+			r.logger.Error("Failed to run rollback action", zap.Error(err))
 		}
 	}
 }
 
 // copyFiles copies files from inputPath to output path, skipping tmpDir.
-func copyFiles(inputPath, outputPath, tmpDir string) error {
+func (r Rollbacker) copyFiles(inputPath, outputPath, tmpDir string) error {
```
cpheps, line 107: Same here if we don't need this to be a receiver function can we just pass in the logger?

updater/internal/action/service_action.go
BinaryFissionGames: I want to see what this looks like when manually testing; I've removed it for now.

updater/internal/logging/logging.go
BinaryFissionGames: Removed this, but won't the log file for the updater just end up growing without bound? I guess it would take quite a few updates to be significant.

updater/cmd/updater/main.go
BinaryFissionGames: I was thinking that maybe the collector could at some point dynamically change this based on the collector log level, but you're right, we should just hard code for now.

updater/internal/logging/logging.go
cpheps: Yes, the updater should probably clean up and previous logs when it starts up. So updates over time don't continuously append on that file.

BinaryFissionGames: Makes sense, the updater now deletes the previous log file when it creates the logger.

updater/internal/action/service_action.go
BinaryFissionGames: I'm going to keep these removed for the service actions, and keep it for the file action; It seems it prints the type, but not any of the contained information, which I think would be helpful for the CopyFileAction.

cpheps (approved):

StefanKurek (approved):

PR #590, opened 2022-07-26
feat: Collector starts up Updater

StefanKurek:
### Proposed Change
Collector should start up updater on a PackagesAvailable message from OpAMP server when there is a new main package to install.
Collector should ignore multiple PackagesAvailable messages while an install is currently in progress.
Fixes log files in downloadable file manager.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


opamp/observiq/updater_manager_others.go
```diff
@@ -0,0 +1,80 @@
...
+// While waiting for Updater, it should kill the collector and we should never execute any code past running it
+func (m OthersUpdaterManager) StartAndMonitorUpdater() error {
+	updaterPath := filepath.Join(m.tmpPath, updaterDir, updaterName)
+	absTmpPath, err := filepath.Abs(m.tmpPath)
+	if err != nil {
+		m.logger.Warn("Failed to get absolute path of tmp dir", zap.Error(err))
+	}
+	cmd := exec.Command(updaterPath, "--tmpdir", absTmpPath)
+
+	cmd.SysProcAttr = &syscall.SysProcAttr{
```
cpheps, line 60: Can we comment why we do this?

opamp/observiq/updater_manager_others.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+// StartAndMonitorUpdater will start the Updater binary and wait to see if it finishes unexpectedly.
+// While waiting for Updater, it should kill the collector and we should never execute any code past running it
+func (m OthersUpdaterManager) StartAndMonitorUpdater() error {
+	updaterPath := filepath.Join(m.tmpPath, updaterDir, updaterName)
+	absTmpPath, err := filepath.Abs(m.tmpPath)
+	if err != nil {
+		m.logger.Warn("Failed to get absolute path of tmp dir", zap.Error(err))
```
cpheps, line 56: Should this return the error? Is this a recoverable failure?

opamp/observiq/updater_manager_others.go
```diff
@@ -0,0 +1,80 @@
...
+	time.Sleep(5 * time.Second)
+
+	// Updater might already be killed
+	if err := cmd.Process.Kill(); err != nil {
+		m.logger.Debug("Failed to kill failed Updater", zap.Error(err))
+	}
+
+	// Ideally we should not get here as we will be killed by the updater.
+	// Updater should either exit before us with error or we die before it does.
+	return fmt.Errorf("updater failed to update collector")
```
cpheps, line 79: should be an `errors.New` since there's no string interpolation.

opamp/observiq/updater_manager_others.go
```diff
@@ -0,0 +1,80 @@
...
+	"path/filepath"
+	"syscall"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+const updaterDir = "latest"
+
+var updaterName = "updater"
```
cpheps, line 31: These live in both windows and others version of the file. Can we put them in the `updater_manager.go` file so there's only one copy across all implementations?

opamp/observiq/updater_manager_windows.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+}
+
+// StartAndMonitorUpdater will start the Updater binary and wait to see if it finishes unexpectedly.
+// While waiting for Updater, it should kill the collector and we should never execute any code past running it
+func (m WindowsUpdaterManager) StartAndMonitorUpdater() error {
+	updaterPath := filepath.Join(m.tmpPath, updaterDir, updaterName)
+	absTmpPath, err := filepath.Abs(m.tmpPath)
+	if err != nil {
+		m.logger.Warn("Failed to get absolute path of tmp dir", zap.Error(err))
```
cpheps, line 55: Same here should we return on this error?

updater/internal/file/file.go
```diff
@@ -40,19 +41,29 @@ func CopyFile(logger *zap.Logger, pathIn, pathOut string, overwrite bool) error
 		}
 	}()
 
+	pathOutClean := filepath.Clean(pathOut)
+	var fileMode fs.FileMode
+	fileMode = 0600
```
cpheps, line 46: Can we condense this into a single line? 

```go
var fileMode fs.FileMode = 0600
```

or 

```go
fileMode := fs.FileMode(0600)
```

updater/internal/install/install_test.go
```diff
@@ -244,70 +244,6 @@ func TestInstallArtifacts(t *testing.T) {
 			action.NewServiceStopAction(svc),
 		}, actions)
 	})
-
-	t.Run("An artifact exists already as a folder", func(t *testing.T) {
```
cpheps, line 248: Just checking we meant to remove this?

opamp/observiq/observiq_downloadable_file_manager_test.go
```diff
@@ -117,13 +118,13 @@ func TestGetOutputFilePath(t *testing.T) {
 		},
 		{
 			name:        "Input url is invalid",
-			basepath:    filepath.Join("tmp", "observiq-otel-collector-update"),
+			basepath:    filepath.Join("/", "tmp", "observiq-otel-collector-update"),
```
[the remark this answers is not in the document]
StefanKurek: @cpheps yeah i just made it consistent with the two tests above (the basepath isn't the thing really being specially tested in the two tests where i changed it)

opamp/observiq/updater_manager_others.go
StefanKurek: @cpheps good catch. i think i was planning to do something different here and forgot about it. i should definitely be returning an error

StefanKurek: @cpheps this is going to go away quickly, because we're going to get rid of passing in the tmp dir anyways

updater/internal/install/install_test.go
StefanKurek: @cpheps yes. this is the test that i mentioned in slack last night. because now i'm deleting old files before they get copied over (to prevent the weirdness of the collector not being able to run), this test is no longer valid

updater/internal/file/file.go
StefanKurek: @cpheps yep that would definitely be better. i meant to do this, but i must have forgot

opamp/observiq/testdata/latest/quickupdater
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sleep 1
```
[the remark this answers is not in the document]
StefanKurek: fixed

opamp/observiq/testdata/latest/slowupdater
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sleep 10
```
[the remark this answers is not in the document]
StefanKurek: fixed

opamp/observiq/updater_manager_others.go
StefanKurek: added comment

opamp/observiq/updater_manager_others.go
StefanKurek: fixed

opamp/observiq/updater_manager_others.go
StefanKurek: fixed

opamp/observiq/updater_manager_others.go
StefanKurek: fixed

opamp/observiq/updater_manager_windows.go
StefanKurek: fixed

updater/internal/file/file.go
StefanKurek: fixed

opamp/observiq/observiq_client.go
```diff
@@ -429,11 +441,30 @@ func (c *Client) installPackageFromFile(file *protobufs.DownloadableFile, curPac
...
-		c.safeSetUpdatingPackage(false)
 		return
 	}
 
-	// TODO: Start Updater
+	if monitorErr := c.updaterManager.StartAndMonitorUpdater(); monitorErr != nil {
+		// Reread package statuses in case Updater changed anything
+		newPackageStatuses, err := c.packagesStateProvider.LastReportedStatuses()
+		if err != nil {
+			c.logger.Error("Failed to read last reported package statuses", zap.Error(err))
```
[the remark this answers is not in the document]
StefanKurek: @cpheps another good catch. i don't know what i was thinking at the time. i'll rework this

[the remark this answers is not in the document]
StefanKurek: @cpheps actually going back and looking at this i think it was all intentional. this will end up kind of being what bindplane would rather have i think? so no matter what, LastReportedStatuses will return a statuses even if there is an error (barebones with the package name and version). I was setting the status/error message on this, even though there is no "all packages hash" to reference back to. I figured (maybe) it was better to give Bindplane something rather than nothing?

opamp/observiq/updater_manager_windows.go
```diff
@@ -0,0 +1,71 @@
...
+
+// StartAndMonitorUpdater will start the Updater binary and wait to see if it finishes unexpectedly.
+// While waiting for Updater, it should kill the collector and we should never execute any code past running it
+func (m WindowsUpdaterManager) StartAndMonitorUpdater() error {
+	updaterPath := filepath.Join(m.tmpPath, updaterDir, updaterName)
+	absTmpPath, err := filepath.Abs(m.tmpPath)
+	if err != nil {
+		return fmt.Errorf("failed to get absolute path of tmp dir: %w", err)
+	}
+	cmd := exec.Command(updaterPath, "--tmpdir", absTmpPath)
```
BinaryFissionGames, line 53: I think we only do gosec in CI on linux right now, but may as well add the nosec here too:
```suggestion
        //#nosec G204 -- paths are not determined via user input
	cmd := exec.Command(updaterPath, "--tmpdir", absTmpPath)
```

opamp/observiq/testdata/latest/quickupdater
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
BinaryFissionGames, line 1: For the windows binaries, do we have the code that generated the binaries? I think it would be helpful to have the committed to know what generated them if it isn't too much trouble, even if it's dead simple (I assume they are just the same as these bash scripts).

opamp/observiq/observiq_client.go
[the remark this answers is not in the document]
cpheps: That makes sense just want to make sure that `newPackageStatuses` will never be `nil`.

updater/internal/file/file.go
```diff
@@ -40,19 +41,29 @@ func CopyFile(logger *zap.Logger, pathIn, pathOut string, overwrite bool) error
...
 	if overwrite {
 		// If we are OK to overwrite, we will truncate the file on open
 		flags |= os.O_TRUNC
+
+		// Save old file's permissions and delete it first if it exists
+		fileOutInfo, _ := os.Stat(pathOutClean)
+		if fileOutInfo != nil {
+			fileMode = fileOutInfo.Mode()
+		}
+		os.Remove(pathOutClean)
```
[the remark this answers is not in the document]
StefanKurek: i have not thought through those things, so it's a good thing you're bringing it up.

opamp/observiq/testdata/latest/quickupdater
StefanKurek: @BinaryFissionGames that's fair. so you're thinking just putting the go files in the testdata/latest directory?
```
package main

import "time"

func main() {
	time.Sleep(1 * time.Second)
}
```

BinaryFissionGames: Yeah, I'm thinking just to commit the go files.

service/observiq-otel-collector.service
```diff
@@ -17,5 +17,6 @@ TimeoutSec=120
 StandardOutput=journal
 Restart=on-failure
 RestartSec=5s
+KillMode=process
```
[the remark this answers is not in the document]
cpheps: It will not orphan the JMX I don't think based on the way OTel kicks it off and shuts it down. It might orphan it if we need to SIGKILL the collector. I can try `mixed` mode which sends a SIGINT to the main process but SIGKILL to all child processes.

opamp/observiq/testdata/latest/quickupdater
cpheps: Is it possible to do this with a batch script or powershell script and avoid a go binary? Do we need it to be an `.exe`?

updater/internal/file/file.go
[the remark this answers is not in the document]
cpheps: We should be able to get the `user:group` data from the stat info I thought. On non windows systems we could re apply them.

I have no idea how window's works. I think ownership is either Admin or the current user.

opamp/observiq/testdata/latest/quickupdater
StefanKurek: @cpheps I initially tried a powershell script. but i couldn't execute the scripts directly like with a shell script. i have to execute powershell.exe instead. maybe there is some straightforward way to do this and get a script to run, but it was easier for me to just make a couple of small windows exe's instead...

opamp/observiq/updater_manager.go
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+package observiq
+
+const updaterDir = "latest"
+
+var updaterName = "updater"
+
+// UpdaterManager handles working with the Updater binary
+type UpdaterManager interface {
```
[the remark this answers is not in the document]
StefanKurek: @BinaryFissionGames changed

opamp/observiq/updater_manager_others.go
```diff
@@ -0,0 +1,79 @@
...
+	"time"
+
+	"go.uber.org/zap"
+)
+
+// Ensure interface is satisfied
+var _ UpdaterManager = (*OthersUpdaterManager)(nil)
+
+// OthersUpdaterManager handles starting a Updater binary and watching it for failure with a timeout
+type OthersUpdaterManager struct {
```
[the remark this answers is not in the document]
StefanKurek: @BinaryFissionGames changed

opamp/observiq/updater_manager_windows.go
StefanKurek: @BinaryFissionGames changed

opamp/observiq/updater_manager_others.go
```diff
@@ -0,0 +1,79 @@
...
+
+// StartAndMonitorUpdater will start the Updater binary and wait to see if it finishes unexpectedly.
+// While waiting for Updater, it should kill the collector and we should never execute any code past running it
+func (m OthersUpdaterManager) StartAndMonitorUpdater() error {
+	updaterPath := filepath.Join(m.tmpPath, updaterDir, updaterName)
+	absTmpPath, err := filepath.Abs(m.tmpPath)
+	if err != nil {
+		return fmt.Errorf("failed to get absolute path of tmp dir: %w", err)
+	}
+	cmd := exec.Command(updaterPath, "--tmpdir", absTmpPath)
```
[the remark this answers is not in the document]
StefanKurek: @BinaryFissionGames changed

updater/internal/file/file.go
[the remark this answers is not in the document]
StefanKurek: @cpheps this is what we can get from stat info
```
type FileInfo interface {
	Name() string       // base name of the file
	Size() int64        // length in bytes for regular files; system-dependent for others
	Mode() FileMode     // file mode bits
	ModTime() time.Time // modification time
	IsDir() bool        // abbreviation for Mode().IsDir()
	Sys() any           // underlying data source (can return nil)
}
```
should i start looking for a workaround to the problem that doesn't involve deleting the old file first?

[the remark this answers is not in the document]
cpheps: Huh, thought it had owner group. There are go functions to get that info but not sure if they work on windows. Looks like from [this](https://stackoverflow.com/questions/58179647/getting-uid-and-gid-of-a-file) article you have to cast the stat info.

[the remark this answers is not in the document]
BinaryFissionGames: So, ideally we wouldn't have to delete the original files to get this to work, but it sounds like we needed it for mac. I'm wondering if it makes sense to have this logic ONLY for mac, where it has the least impact in terms of permissions, or if we need it for linux and windows too.

As far as owner/group on windows, from my cursory understanding it doesn't really have a concept of a "sole-owner/group" the way that unix does, it has a list of groups/users and what permissions each has, so it's a lot more complicated.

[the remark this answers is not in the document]
cpheps: There's also no native go logic for getting that info on windows we'd have to write our own or find a library.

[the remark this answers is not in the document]
BinaryFissionGames: It seems like we don't really have to worry about this all that much, thinking on it some more. In managed mode, the whole point is not to have to muck around with the files, and this won't effect the config files, which are the ones where permissions are actually important.

Seems good as is.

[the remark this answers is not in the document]
cpheps: Talked in slack sounds like we are ok leaving this as is for now.

service/observiq-otel-collector.service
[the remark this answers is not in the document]
cpheps: `KillMode` has to be `process` looks like. It won't orphan the updater otherwise. Looking at the [code](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/jmxreceiver/internal/subprocess/subprocess.go#L134-L155) for how the JMX jar is shutdown it can already be zombied during normal shutdown if it takes to long. I think this is probably an unlikely case that we shouldn't optimize for right now.

cpheps (approved):
LGTM just need to fix the gosec issue

BinaryFissionGames (approved):

PR #591, opened 2022-07-27
feat: Remove tmpdir from updater

BinaryFissionGames:
### Proposed Change
* Remove tmpdir command line parameter
* Refactor to have the installDir be the base of all paths, as opposed to the mix of relative paths and paths based on tmpDir.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


.gitignore
```diff
@@ -20,7 +20,7 @@ windows/plugins
 opentelemetry-java-contrib-jmx-metrics.jar
 VERSION.txt
 release_deps
-tmp
+/tmp
```
cpheps, line 23: curious why we need the leading `/`?

cpheps (approved):

.gitignore
BinaryFissionGames: I have a testdata directory that has a tmp directory (it's the in the folder structure we are using for updating), so I needed to change something to make sure git picks it up.

BinaryFissionGames:
@StefanKurek Ripped it out now, I think it should be good.

updater/internal/install/install.go
```diff
@@ -93,7 +85,7 @@ func (i Installer) Install(rb rollback.ActionAppender) error {
 
 // copyFiles moves the file tree rooted at latestDirPath to installDirPath,
 // skipping configuration files. Appends CopyFileAction-s to the Rollbacker as it copies file.
-func copyFiles(logger *zap.Logger, inputPath, outputPath, tmpDir string, rb rollback.ActionAppender) error {
+func copyFiles(logger *zap.Logger, inputPath, outputPath, installDir string, rb rollback.ActionAppender) error {
```
StefanKurek, line 88: This is only used in this file right? Doesn't seem necessary to have both outputPath and installDir since they ultimately will be the same thing?

updater/internal/rollback/rollback.go
```diff
@@ -104,8 +102,8 @@ func (r Rollbacker) Rollback() {
 }
 
 // copyFiles copies files from inputPath to output path, skipping tmpDir.
-func copyFiles(logger *zap.Logger, inputPath, outputPath, tmpDir string) error {
-	absTmpDir, err := filepath.Abs(tmpDir)
+func copyFiles(logger *zap.Logger, inputPath, outputPath, installDir string) error {
```
StefanKurek, line 105: similar question to the other file. it seems a little odd to be passing in the same variable twice when it's the only use case of the function. could we get way with condensing the arguments?

updater/internal/install/install.go
BinaryFissionGames: Good catch, refactored this.
I also gave it a better suited name that seems less general.

updater/internal/rollback/rollback.go
BinaryFissionGames: done!

StefanKurek (approved):

PR #594, opened 2022-07-28
feat: Copy updater executable to CWD of collector before executing

BinaryFissionGames:
### Proposed Change
* Copy the updater to CWD, and execute it there.
* Tweak tests for updaterManager to run in parallel (this knocks like 4-5 seconds, from 11-12 seconds to ~6-7 seconds)

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/opamp/observiq/updater_manager.go
+++ b/opamp/observiq/updater_manager.go
@@ -25,8 +25,6 @@ import (
 
 const updaterDir = "latest"
 
-var updaterName = "updater"
-
 // updaterManager handles working with the Updater binary
 type updaterManager interface {
 	// StartAndMonitorUpdater starts the Updater binary and monitors it for failure
--- a/opamp/observiq/updater_manager.go
+++ b/opamp/observiq/updater_manager.go
@@ -49,7 +49,7 @@ func copyExecutable(logger *zap.Logger, inputPath, cwd string) (string, error) {
 	// Output path is just whatever the actual file name is (e.g. updater.exe),
 	// on top of the CWD. We take the absolute path, because it is needed to actually ensure you can
 	// exec a file not on your PATH.
-	outputPath, err := filepath.Abs(filepath.Base(inputPath))
+	outputPath, err := filepath.Abs(filepath.Join(cwd, filepath.Base(inputPath)))
 	if err != nil {
 		return "", fmt.Errorf("failed to get absolute path for output: %w", err)
 	}
--- a/opamp/observiq/updater_manager.go
+++ b/opamp/observiq/updater_manager.go
@@ -34,8 +34,8 @@ type updaterManager interface {
 }
 
 // copyExecutable copies the executable at the input file path to the cwd.
-// Returns the output path (which is just the filepath.Base of the input path)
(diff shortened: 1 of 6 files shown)


opamp/observiq/updater_manager.go
```diff
@@ -14,12 +14,67 @@
...
+	// on top of the CWD. We take the absolute path, because it is needed to actually ensure you can
+	// exec a file not on your PATH.
+	outputPath, err := filepath.Abs(filepath.Join(cwd, filepath.Base(inputPath)))
+	if err != nil {
+		return "", fmt.Errorf("failed to get absolute path for output: %w", err)
+	}
+
+	outputPathClean := filepath.Clean(outputPath)
+
+	// Remove the file if it already exists, need this for macOS
```
StefanKurek, line 59: @BinaryFissionGames just a note that we "may" not need this if you're just assuming because of that code I added. the problem I was seeing when copying over old files existed on executing the collector in the install directory, but not the updater for whatever reason.

opamp/observiq/updater_manager_others.go
```diff
@@ -30,24 +31,39 @@ import (
...
 var _ updaterManager = (*othersUpdaterManager)(nil)
 
+const defaultOthersUpdaterName = "updater"
+
 // othersUpdaterManager handles starting a Updater binary and watching it for failure with a timeout
 type othersUpdaterManager struct {
-	tmpPath string
-	logger  *zap.Logger
+	tmpPath     string
+	cwd         string
```
StefanKurek, line 39: @BinaryFissionGames do we really need to save this? (seems a little odd specifically because it's the current working directory and we only use it once it looks like here)

opamp/observiq/updater_manager_windows.go
```diff
@@ -19,35 +19,50 @@ package observiq
...
 // Ensure interface is satisfied
 var _ updaterManager = (*windowsUpdaterManager)(nil)
 
 // windowsUpdaterManager handles starting a Updater binary and watching it for failure with a timeout
 type windowsUpdaterManager struct {
-	tmpPath string
-	logger  *zap.Logger
+	tmpPath     string
+	updaterName string
+	cwd         string
```
StefanKurek, line 39: @BinaryFissionGames same question here

cpheps (approved):

opamp/observiq/updater_manager_others.go
BinaryFissionGames: The purpose of the variable is so we can replace it in tests, I'm open to other ideas of injecting the CWD.

StefanKurek: @BinaryFissionGames ok fair enough, I didn't catch that.

opamp/observiq/updater_manager.go
BinaryFissionGames: Yeah, I kinda assumed from your findings with the collector.

What'll end up happening is the updater will be copied over the previous updater in a similar fashion, so I thought this would also be effected.

PR #595, opened 2022-07-28
Windows updater log fix

BinaryFissionGames:
### Proposed Change
* fix the logger not being created on windows due to https://github.com/uber-go/zap/issues/621

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


updater/internal/logging/logging_windows.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/observiq/observiq-otel-collector/updater/internal/path"
+	"go.uber.org/zap"
+)
+
+var registerSinkOnce = &sync.Once{}
+
+// NewLogger returns a new logger, that logs to the log directory relative to installDir.
+// It deletes the previous log file, as well.
+// NewLogger must only be called once, at the start of the program.
+func NewLogger(installDir string) (*zap.Logger, error) {
```
cpheps, line 32: Is there a danger this is called more than once?

BinaryFissionGames: No, just a note for the future. The issue is that we delete the previous file in this function. Actually, we could fix that with a sync.Once, if we think it's a good idea.

cpheps (approved):

StefanKurek (approved):

updater/internal/logging/logging_windows.go
cpheps: This is fine as is just more curious.

PR #596, opened 2022-07-28
feat: Updater cleans up temporary directory

BinaryFissionGames:
### Proposed Change
* On error, call "fail" function, which removes the temporary directory before exiting
  * Before we were calling logger.Fatal in these scenarios to exit
* On success, remove temporary directory.
* Remove error from signature of NewRollbacker and NewInstaller
  * Neither of these need to return an error

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/updater/cmd/updater/main.go
+++ b/updater/cmd/updater/main.go
@@ -56,6 +56,7 @@ func main() {
 	logger, err := logging.NewLogger(installDir)
 	if err != nil {
 		log.Default().Printf("Failed to create logger: %s\n", err)
+		// No logger, can't really log if we fail to remove tmpdir
 		fail(zap.NewNop(), installDir)
 	}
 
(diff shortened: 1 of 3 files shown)


updater/cmd/updater/main.go
```diff
@@ -106,9 +108,26 @@ func main() {
 		}
 
 		rb.Rollback()
-		logger.Fatal("Rollback complete")
+		removeTmpDir(logger, installDir)
+
+		logger.Error("Rollback complete")
+		fail(logger, installDir)
```
cpheps, line 114: This seems redundant. Calling `removeTmpDir` then call `fail` which calls `removeTmpDir`

BinaryFissionGames: Good catch, this is redundant, removed the redundant calls.

cpheps (approved):

StefanKurek (approved):

PR #597, opened 2022-07-28
feat: Harden collector shutdown while updating

BinaryFissionGames:
### Proposed Change
* Change service timeouts to be 20 seconds
  * This is the default for macOS, but I added it in just to be sure
  * Windows default is 20s and isn't easy to set, so I left it where it was
* Change collector wait timeout for updater to 30 seconds
* Change updater to stop the collector before performing the backup, in order to reduce time to service shutdown

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/updater/cmd/updater/main.go
+++ b/updater/cmd/updater/main.go
@@ -23,10 +23,12 @@ import (
 	"time"
 
 	"github.com/observiq/observiq-otel-collector/packagestate"
+	"github.com/observiq/observiq-otel-collector/updater/internal/action"
 	"github.com/observiq/observiq-otel-collector/updater/internal/install"
 	"github.com/observiq/observiq-otel-collector/updater/internal/logging"
 	"github.com/observiq/observiq-otel-collector/updater/internal/path"
 	"github.com/observiq/observiq-otel-collector/updater/internal/rollback"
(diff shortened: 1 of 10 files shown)


cpheps (approved):

cpheps (approved):

updater/cmd/updater/main.go
```diff
@@ -68,12 +70,34 @@ func main() {
 	}
 
 	rb := rollback.NewRollbacker(logger, installDir)
+	// Stop the service before backing up the install directory;
+	// We want to stop as early as possible so that we don't hit the collector's timeout
+	// while it waits to be shutdown.
+	service := service.NewService(logger, installDir)
```
StefanKurek, line 76: i'm not going to complain too much, but maybe to something to think about when we give this whole thing a once over. it would have been nice if we could have just had a method to call on the installer to do the stopping, instead of copying "a bunch" (i know its really not that much) of logic from it to here

StefanKurek (approved):

updater/cmd/updater/main.go
BinaryFissionGames: Yeah, good call out, definitely should be a part of the larger refactor.

PR #600, opened 2022-07-28
fix: Makes sure JMX jar is installed and rolled back properly

StefanKurek:
### Proposed Change
Ensures JMX jar is both installed and rolled back properly
Also making sure that we use the backed up file permissions when
rolling back a file that no longer exists in the install directory
Updated some tests that were extremely hard to debug.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/updater/internal/file/file.go
+++ b/updater/internal/file/file.go
@@ -26,7 +26,7 @@ import (
 
 // CopyFile copies the file from pathIn to pathOut.
 // If the file does not exist, it is created. If the file does exist, it is truncated before writing.
-func CopyFile(logger *zap.Logger, pathIn, pathOut string, overwrite bool) error {
+func CopyFile(logger *zap.Logger, pathIn, pathOut string, overwrite bool, useInFilePermBackup bool) error {
 	pathInClean := filepath.Clean(pathIn)
 
 	// Open the input file for reading.
@@ -43,21 +43,41 @@ func CopyFile(logger *zap.Logger, pathIn, pathOut string, overwrite bool) error 
 
 	pathOutClean := filepath.Clean(pathOut)
 	fileMode := fs.FileMode(0600)
-
 	flags := os.O_CREATE | os.O_WRONLY
 	if overwrite {
 		// If we are OK to overwrite, we will truncate the file on open
 		flags |= os.O_TRUNC
 
-		// Save old file's permissions and delete it first if it exists
-		fileOutInfo, _ := os.Stat(pathOutClean)
-		if fileOutInfo != nil {
-			fileMode = fileOutInfo.Mode()
+		// Try to save old file's permissions
+		outFileInfo, _ := os.Stat(pathOutClean)
+		if outFileInfo != nil {
+			fileMode = outFileInfo.Mode()
+		} else if useInFilePermBackup {
+			// Use the new file's permissions as a backup and don't fail on error (best chance for rollback)
+			inFileInfo, err := inFile.Stat()
+			switch {
+			case err != nil:
+				logger.Error("failed to retrieve fileinfo for input file", zap.Error(err))
+			case inFileInfo != nil:
+				fileMode = inFileInfo.Mode()
+			}
 		}
-		_ = os.Remove(pathOutClean)
+
+		// Remove old file to prevent issues with mac
+		if err = os.Remove(pathOutClean); err != nil {
+			logger.Debug("Failed to remove output file", zap.Error(err))
+		}
 	} else {
 		// This flag will make OpenFile error if the file already exists
 		flags |= os.O_EXCL
+
+		// Use the new file's permissions and fail if there's an issue (want to fail for backup)
+		inFileInfo, err := inFile.Stat()
+		if err != nil {
+			return fmt.Errorf("failed to retrive fileinfo for input file: %w", err)
+		}
+
+		fileMode = inFileInfo.Mode()
 	}
 
 	// Open the output file, creating it if it does not exist and truncating it.
--- a/updater/internal/install/install.go
+++ b/updater/internal/install/install.go
@@ -15,6 +15,7 @@
 package install
 
 import (
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
@@ -33,6 +34,7 @@ import (
 type Installer struct {
 	latestDir  string
 	installDir string
+	backupDir  string
 	svc        service.Service
 	logger     *zap.Logger
 }
@@ -44,6 +46,7 @@ func NewInstaller(logger *zap.Logger, installDir string) *Installer {
 		latestDir:  path.LatestDir(installDir),
 		svc:        service.NewService(namedLogger, installDir),
 		installDir: installDir,
+		backupDir:  path.BackupDir(installDir),
 		logger:     namedLogger,
 	}
 }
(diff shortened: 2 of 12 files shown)


updater/internal/file/file.go
```diff
@@ -43,21 +43,42 @@ func CopyFile(logger *zap.Logger, pathIn, pathOut string, overwrite bool) error
...
+			fileMode = outFileInfo.Mode()
+		} else if useInFilePermBackup {
+			// Use the new file's permissions as a backup and don't fail on error (best chance for rollback)
+			inFileInfo, err := inFile.Stat()
+			if err != nil {
+				logger.Error("failed to retrieve fileinfo for input file", zap.Error(err))
+			}
+			if inFileInfo != nil {
+				fileMode = inFileInfo.Mode()
+			}
```
cpheps, line 63: This check seems odd. If `err != nil` likely `inFileInfo == nil` These two cases are mutually exclusive so only one can happen.

It might read better as a switch statement:

```go
switch {
  case err != nil:
    logger.Error("")
  case inFileInfo != nil:
    fileMode = inFileInfo.Mode()
}
```

updater/internal/file/file.go
```diff
@@ -43,21 +43,42 @@ func CopyFile(logger *zap.Logger, pathIn, pathOut string, overwrite bool) error
...
 		flags |= os.O_EXCL
+
+		// Use the new file's permissions and fail if there's an issue (want to fail for backup)
+		inFileInfo, err := inFile.Stat()
+		if err != nil {
+			return fmt.Errorf("failed to retrive fileinfo for input file: %w", err)
+		}
+		if inFileInfo != nil {
+			fileMode = inFileInfo.Mode()
+		}
```
cpheps, line 81: Since the err is returning you can just change this to

```go
inFileInfo, err := inFile.Stat()
if err != nil {
    return fmt.Errorf("failed to retrive fileinfo for input file: %w", err)
}
fileMode = inFileInfo.Mode()
		
```

updater/internal/install/install.go
```diff
@@ -58,9 +61,27 @@ func (i Installer) Install(rb rollback.ActionAppender) error {
...
+		if err := installFile(i.logger, latestJarPath, jarDirPath, i.backupDir, rb); err != nil {
+			return fmt.Errorf("failed to install JMX jar from latest directory: %w", err)
+		}
+		// Just log this error as the worst case is that there will be two jars copied over
+		if err = os.Remove(latestJarPath); err != nil {
+			i.logger.Warn("Failed to remove JMX jar from latest directory", zap.Error(err))
+		}
+	case !errors.Is(err, os.ErrNotExist):
+		return fmt.Errorf("failed determine where currently installed JMX jar is: %w", err)
+	}
```
cpheps, line 80: Any chance we can break this logic out into it's own function since it's specific to the JMX jar?

cpheps (commented):
A lot of this is over my head. I left comments from a code standpoint of what made sense. I'm not quite sure about all the file ops.

updater/internal/file/file.go
StefanKurek: @cpheps i just wasn't sure if something weird could happen like an error occurred but there was still file info. but yeah, it probably doesn't make sense. i can change this.

updater/internal/file/file.go
StefanKurek: same comment as above. i'll change it.

cpheps (approved):
Tested these changes on all 3 OSes and they work.

updater/internal/file/file.go
```diff
@@ -43,21 +43,41 @@ func CopyFile(logger *zap.Logger, pathIn, pathOut string, overwrite bool) error
...
+		} else if useInFilePermBackup {
+			// Use the new file's permissions as a backup and don't fail on error (best chance for rollback)
+			inFileInfo, err := inFile.Stat()
+			switch {
+			case err != nil:
+				logger.Error("failed to retrieve fileinfo for input file", zap.Error(err))
+			case inFileInfo != nil:
+				fileMode = inFileInfo.Mode()
+			}
+		}
```
BinaryFissionGames, line 64: I think this is right, but does it make sense to always prefer the destination file's permissions, even when `useInFilePermBackup` is true? The logic just feels off to me for some reason.

updater/internal/file/file.go
```diff
@@ -43,21 +43,41 @@ func CopyFile(logger *zap.Logger, pathIn, pathOut string, overwrite bool) error
...
 		// This flag will make OpenFile error if the file already exists
 		flags |= os.O_EXCL
+
+		// Use the new file's permissions and fail if there's an issue (want to fail for backup)
+		inFileInfo, err := inFile.Stat()
+		if err != nil {
+			return fmt.Errorf("failed to retrive fileinfo for input file: %w", err)
+		}
+
+		fileMode = inFileInfo.Mode()
```
BinaryFissionGames, line 80: Do we always want to use the input files's mode in this case, or only when the `useInFilePermBackup` flag is true?
I guess we only use this on backup, so I guess `overwrite` = false currently implies `useInFilePermBackup` = true.

updater/internal/rollback/rollback.go
```diff
@@ -140,7 +153,7 @@ func backupFiles(logger *zap.Logger, installDir, outputPath string) error {
 		}
 
 		// Fail if copying the input file to the output file would fail
-		if err := file.CopyFile(logger.Named("copy-file"), inPath, outPath, false); err != nil {
+		if err := file.CopyFile(logger.Named("copy-file"), inPath, outPath, false, false); err != nil {
 			return fmt.Errorf("failed to copy file: %w", err)
 		}
```
BinaryFissionGames, line 158: I know this doesn't matter (see above comment about `useInFilePermBackup` = true and `overwrite` = false), but I'd expect `useInFilePermBackup` to be set to true here when backing up a file.

updater/internal/rollback/rollback.go
```diff
@@ -153,3 +166,24 @@ func backupFiles(logger *zap.Logger, installDir, outputPath string) error {
...
+	// to get the out directory (which we will create if it does not exist).
+	outPath := filepath.Join(outputDirPath, baseInPath)
+	outDir := filepath.Dir(outPath)
+
+	if err := os.MkdirAll(outDir, 0750); err != nil {
+		return fmt.Errorf("failed to create dir: %w", err)
+	}
+
+	// Fail if copying the input file to the output file would fail
+	if err := file.CopyFile(logger.Named("copy-file"), inPath, outPath, false, false); err != nil {
```
BinaryFissionGames, line 184: Same here, i'd assume the last parameter to be true.

updater/internal/service/service_darwin.go
```diff
@@ -140,7 +140,7 @@ func (d darwinService) Update() error {
 }
 
 func (d darwinService) Backup() error {
-	if err := file.CopyFile(d.logger.Named("copy-file"), d.installedServiceFilePath, path.BackupServiceFile(d.installDir), false); err != nil {
+	if err := file.CopyFile(d.logger.Named("copy-file"), d.installedServiceFilePath, path.BackupServiceFile(d.installDir), false, false); err != nil {
```
BinaryFissionGames, line 143: Same here, I'd expect the last parameter to be true.

updater/internal/service/service_linux.go
```diff
@@ -165,7 +165,7 @@ func (l linuxService) Update() error {
 }
 
 func (l linuxService) Backup() error {
-	if err := file.CopyFile(l.logger.Named("copy-file"), l.installedServiceFilePath, path.BackupServiceFile(l.installDir), false); err != nil {
+	if err := file.CopyFile(l.logger.Named("copy-file"), l.installedServiceFilePath, path.BackupServiceFile(l.installDir), false, false); err != nil {
```
BinaryFissionGames, line 168: Same here, I'd expect the last parameter to be true.

updater/internal/file/file.go
cpheps: yeah...that's a good point. If the old file exists there's really no reason we won't get it's permissions.

StefanKurek: yeah. it's all a little confusing for sure.  ultimately we have 3 cases where we use this copy method.
1. Backup - We want to preserve the original file's permissions on the backed up file
2. Install - We want use the existing file's permissions. If there is no existing file, we are going to use 0600 and hope for the best (this might cause issues if we have executables that we add in the future).
3. Rollback - This is a little more ambiguous. If the file currently exists, it probably doesn't matter that much if we use the existing file's permissions or the backed up file's permissions. If the file does NOT exist, then we definitely want to use the backed up file's permissions (and not the default 0600).

BinaryFissionGames (approved):
Talked through the logic, makes sense, might want to come back for a refactor on this later.

PR #601, opened 2022-07-29
fix: If the collector detects an error updating, clean temporary directory

BinaryFissionGames:
### Proposed Change
* If the collector fails extracting, or the updater fails in a way that leaves the collector running, it will delete the tmp directory.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cpheps (approved):

opamp/observiq/observiq_client.go
```diff
@@ -430,6 +430,7 @@ func (c *Client) createPackageMaps(
 func (c *Client) installPackageFromFile(file *protobufs.DownloadableFile, curPackageStatuses *protobufs.PackageStatuses) {
 	// There should be no reason for us to exit this function unless we detected a problem with the Updater's installation
 	defer c.safeSetUpdatingPackage(false)
+	defer c.downloadableFileManager.CleanupArtifacts()
```
cpheps, line 433: This seems like this will always trigger do we want that?

BinaryFissionGames: reverted this change, thought it made sense because of the comment above but it apparently breaks things.

StefanKurek (approved):

PR #605, opened 2022-08-01
fix(updater): Do Update in place for windows service

BinaryFissionGames:
### Proposed Change
* Instead of uninstalling and re-installing, do the Update in-place for windows.
  * This means we remove both the `install` and `uninstall` functions

Previously, we relied on a wait loop to wait for the service to actually uninstall after being marked for removal.

In this PR, we change how we update the service to not need an uninstall, so we don't need to wait for the service manager to remove the service. Instead, we just apply an update to the existing configuration.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/updater/internal/service/service_windows.go
+++ b/updater/internal/service/service_windows.go
@@ -19,11 +19,9 @@ package service
 import (
 	"encoding/json"
 	"fmt"
-	"log"
 	"os"
 	"path/filepath"
 	"strings"
-	"time"
 
 	"go.uber.org/zap"
 	"golang.org/x/sys/windows/svc"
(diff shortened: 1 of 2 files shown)


cpheps (approved):

StefanKurek (approved):
Looks good as far as I can tell. I do have a question on if we should be manually testing that an update still works with any new code we try to merge in?

BinaryFissionGames:
@StefanKurek My opinion is that we should hold off on manual testing (just because it takes a significant amount of time) until we are done with refactors, then we should test manually again.

cpheps:
> @StefanKurek My opinion is that we should hold off on manual testing (just because it takes a significant amount of time) until we are done with refactors, then we should test manually again.

@BinaryFissionGames my only hesitation with this is the BP OP team is using the `opamp-upgrade` branch to test against. Since we reached MVP we are treating that branch as stable. Small changes are fine but anything we're not sure of or is large we should do a manual test.

StefanKurek:
@BinaryFissionGames and if you need me to help with the testing load for any PRs, just let me know.

BinaryFissionGames:
@cpheps Yeah, I hadn't considered the BP-OP team were currently using the branch for testing;

I tested this change and it worked as expected.

PR #608, opened 2022-08-02
feat: Refactor updater main

BinaryFissionGames:
### Proposed Change
* Refactor the updater main function into a separate struct to make it more testable, and reduce the size of main.
* Made Installer & Rollbacker into interfaces

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


StefanKurek (approved):

cpheps:
Have we tested an update with this? Seems fine but it's a big refactor of the updated code.

updater/cmd/updater/main.go
```diff
@@ -57,100 +47,17 @@ func main() {
...
-	os.Exit(1)
-}
-
-// removeTmpDir removes the temporary directory and any files in it.
-func removeTmpDir(logger *zap.Logger, installDir string) {
-	err := os.RemoveAll(path.TempDir(installDir))
-	if err != nil {
-		logger.Error("failed to remove temporary directory", zap.Error(err))
-	}
+	logger.Debug("Updater finished successfully")
```
cpheps, line 62: This should probably be `Info` level. I don't think much gets logged in the happy case so having some final affirmation always would be good. I know we always pass in debug flag but if we ever change that I don't want us to lose informative logs.

updater/internal/install/install.go
```diff
@@ -29,9 +29,16 @@ import (
...
+//go:generate mockery --name Installer --filename installer.go
+type Installer interface {
+	// Install installs new artifacts over the old ones.
+	Install(rollback.Rollbacker) error
+}
+
+// installer allows you to install files from latestDir into installDir,
 // as well as update the service configuration using the "Install" method.
-type Installer struct {
+type installer struct {
```
cpheps, line 41: nit: I'm not a huge fan of having the implementation of the interface be lower case. Feels like this could lead to confusion. Maybe even just calling this `packageInstaller` to differentiate it from the interface a bit more.

updater/internal/rollback/rollback.go
```diff
@@ -29,15 +29,20 @@ import (
...
+	Backup() error
+	// Rollback undoes the actions recorded by AppendAction.
+	Rollback()
 }
 
-// Rollbacker is a struct that records rollback information,
+// rollbacker is a struct that records rollback information,
 // and can use that information to perform a rollback.
-type Rollbacker struct {
+type rollbacker struct {
```
cpheps, line 45: Same here having this just be a lowercase name of the interface could lead to confusion.

cpheps (commented):
Overall I like this changed the testability it enables for us. A few nit picky things. I also want to do a real update test with this before we merge if we haven't yet.

cpheps (approved):

BinaryFissionGames:
Tested locally on my Mac, works as expected.

PR #611, opened 2022-08-02
feat: Refactor Updater's file package

BinaryFissionGames:
### Proposed Change
* Refactor CopyFile into 3 separate exported functions and 1 unexported function.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/updater/internal/file/file.go
+++ b/updater/internal/file/file.go
@@ -79,10 +79,6 @@ func CopyFileRollback(logger *zap.Logger, pathIn, pathOut string) error {
 	case errors.Is(err, os.ErrNotExist):
 		return fmt.Errorf("input file does not exist: %w", err)
 	case err != nil:
-		// TODO: Should this actually error out?
-		// If we can't STAT the input file, what are the chances we can open for reading?
-		// we are going to delete the file, so we might be better off bailing here and leaving the
-		// newer file instead of deleting it and failing to copy the input file
 		logger.Error("failed to retrieve fileinfo for input file", zap.Error(err))
 	default:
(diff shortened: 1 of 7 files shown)


updater/internal/file/file.go
```diff
@@ -24,9 +25,80 @@ import (
...
+	inFileInfo, err := os.Stat(pathInClean)
+	switch {
+	case errors.Is(err, os.ErrNotExist):
+		return fmt.Errorf("input file does not exist: %w", err)
+	case err != nil:
+		// TODO: Should this actually error out?
+		// If we can't STAT the input file, what are the chances we can open for reading?
+		// we are going to delete the file, so we might be better off bailing here and leaving the
+		// newer file instead of deleting it and failing to copy the input file
+		logger.Error("failed to retrieve fileinfo for input file", zap.Error(err))
```
BinaryFissionGames, line 86: I'm not sure if continuing in this case is actually OK; I'm thinking it might make sense to just error out here instead for the reasons I have in this comment.

Any thoughts?

StefanKurek: you're most likely right? but i don't know about just erroring out completely. we should let things continue just in case and give it the best chance to roll back as completely as possible

StefanKurek: overall this PR is an improvement though. 👍

BinaryFissionGames: Added comment about why we continue here.

BinaryFissionGames:
Tested happy path + rollback locally on my mac, both paths worked as expected.

PR #614, opened 2022-08-04
Added better logging/messaging around collector package updating

StefanKurek:
### Proposed Change
Added more logging around updater process in collector.
Fixed version mismatch while installing not having a package status error message

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/opamp/observiq/observiq_client.go
+++ b/opamp/observiq/observiq_client.go
@@ -17,6 +17,7 @@ package observiq
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


opamp/observiq/observiq_client.go
```diff
@@ -240,7 +242,15 @@ func (c *Client) onConnectHandler() {
 			c.logger.Error("OpAMP client failed to set package statuses", zap.Error(err))
 		}
 	} else {
+		c.logger.Error(
```
cpheps, line 245: For zap logging we can include attributes by using `zap.String("package_name", "packagestate.CollectorPackageName")` and it'll add them to the Json. I think this might be better than doing string interpolation. The main body message can just say `Package update failed because...`

cpheps: Same comment with all the logs added here.

StefanKurek: @cpheps ah cool. i'll do that

BinaryFissionGames (approved):
LGTM after Corbin's comment is resolved.

cpheps (approved):

PR #627, opened 2022-08-08
fix: Generate Agent ID if `manager.yaml` exists and does not supply the field

schmikei:
### Proposed Change
For users that specify a `manager.yaml` and do not supply an agent ID, we should generate a valid UUIDv4 and rewrite the config before establishing any opamp connections. This should not impede standalone collector usage as the manager 

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/cmd/collector/main_test.go
+++ b/cmd/collector/main_test.go
@@ -147,8 +147,8 @@ func TestManagerConfigCheckFileModes(t *testing.T) {
 			expectedErr: errors.New("failed to rewrite manager config with identifying fields"),
 		},
 		{
-			name:        "execute_only",
-			fileMode:    0100,
+			name:        "write_only",
+			fileMode:    0222,
 			expectedErr: errors.New("unable to read file"),
 		},
 		{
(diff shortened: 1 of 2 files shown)


cmd/collector/main_test.go
```diff
@@ -69,12 +74,94 @@ func TestCheckManagerConfigNoFile(t *testing.T) {
 	require.Equal(t, expected, actual)
 }
 
-func TestCheckManagerConfig(t *testing.T) {
-	tmpdir := t.TempDir()
-	manager := filepath.Join(tmpdir, "manager.yaml")
+func TestManagerConfigNoAgentIDWillSet(t *testing.T) {
+	tmpDir := t.TempDir()
+	defer os.RemoveAll(tmpDir)
```
cpheps, line 79: This isn't needed as [t.TempDir()](https://pkg.go.dev/testing#T.TempDir) automatically cleans up.

StefanKurek (approved):

cpheps (approved):

PR #633, opened 2022-08-09
feat: Added description of timezone parameter type to README

dmikolay:
### Proposed Change
- Added description of timezone parameter type to README
- Put under a warning in parameter section
- Linked timezone.go so user can view options

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/pluginreceiver/README.md
```diff
@@ -75,6 +75,8 @@ The following keys are used when defining a parameter.
 
 **Warning**: Parameters must adhere to their definition. Invalid parameters will return an error during configuration.
 
+**Warning**: The value given for the timezone parameter type must be a standard UTC timezone, shown in this [timezone list](https://github.com/observIQ/observiq-otel-collector/blob/main/receiver/pluginreceiver/timezone.go).
```
cpheps, line 78: Can you actually list this out in a separate markdown file? Sending users to a Go file is not super user friendly.

Something like this:

```markdown
- UTC
- Africa/Abidjan
....
```

Also I think this should be more of a `Note` rather than a `Warning`

docs/timezone.md
```diff
@@ -0,0 +1,426 @@
+- "UTC"
```
cpheps, line 1: Can we move this file into the `pluginreceiver` directory? I'd like to keep high level docs in the `docs` directory. Also, can we add a title and brief summary?

Something like:

```markdown
# Timezone

Below is a list of possible values for the `timezone` plugin parameter:
```

dmikolay: My mistake, just saw all the md files in docs and assumed it was supposed to be there

StefanKurek (approved):

receiver/pluginreceiver/README.md
```diff
@@ -75,6 +75,8 @@ The following keys are used when defining a parameter.
 
 **Warning**: Parameters must adhere to their definition. Invalid parameters will return an error during configuration.
 
+**Note**: The value given for the timezone parameter type must be a standard UTC timezone, shown in this [timezone list](https://github.com/observIQ/observiq-otel-collector/blob/main/receiver/pluginreceiver/timezone.md).
```
cpheps, line 78: This link I think can just be

```markdown
[timezone list](./timzone.md)
```

Right now it includes the branch name so it won't render the correct version on the current branch.

cpheps (approved):

PR #634, opened 2022-08-09
Refactored Package Installed related functions in observiq_client

StefanKurek:
### Proposed Change
Refactored a bunch of stuff in the observiq client regarding Update functionality. None of this is big architectural refactoring (not implementing some system where we can easily handle installation of multiple types of packages). Most of this is just to make it easier to read/understand. Fixes a few small issues along the way.

I still need to manually test this, but I don't expect any issues.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/opamp/observiq/observiq_client.go
+++ b/opamp/observiq/observiq_client.go
@@ -146,7 +146,7 @@ func (c *Client) Connect(ctx context.Context) error {
 		c.logger.Error("Error while setting agent description", zap.Error(err))
 
 		// Set package status file for error (for Updater to pick up), but do not force send to Server
-		c.attemptFailedInstall(fmt.Sprintf("Error while setting agent description: %s", err.Error()))
+		c.tryToFailPackageInstall(fmt.Sprintf("Failed setting agent description: %s", err.Error()), false)
 
 		return err
 	}
@@ -154,7 +154,7 @@ func (c *Client) Connect(ctx context.Context) error {
 	tlsCfg, err := c.currentConfig.ToTLS()
 	if err != nil {
 		// Set package status file for error (for Updater to pick up), but do not force send to Server
-		c.attemptFailedInstall(fmt.Sprintf("Failed creating TLS config: %s", err.Error()))
+		c.tryToFailPackageInstall(fmt.Sprintf("Failed creating TLS config: %s", err.Error()), false)
 
 		return fmt.Errorf("failed creating TLS config: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


opamp/observiq/observiq_client.go
```diff
@@ -334,230 +325,274 @@ func (c *Client) onPackagesAvailableHandler(packagesAvailable *protobufs.Package
...
+	for pkgName, availablePkg := range availablePkgs.Packages {
+		lastPkgStatus := lastPkgStatusMap[pkgName]
+		curPkgStatuses.Packages[pkgName] = c.buildInitialPackageStatus(pkgName, availablePkg, lastPkgStatus)
+	}
 
 	// This is an error because we need this file for communication during the update
-	if err = c.packagesStateProvider.SetLastReportedStatuses(curPackageStatuses); err != nil {
+	if err = c.packagesStateProvider.SetLastReportedStatuses(curPkgStatuses); err != nil {
 		return fmt.Errorf("failed to save last reported package statuses: %w", err)
 	}
```
BinaryFissionGames, line 360: This is kinda unrelated to the refactor, BUT I noticed that we fail here without sending package statuses back to bindplane; Should we try to send a failure message back in this case, or is this intentional?

BinaryFissionGames (approved):

opamp/observiq/observiq_client_test.go
```diff
@@ -173,11 +197,30 @@ func TestClientConnect(t *testing.T) {
...
+				}
 
 				mockOpAmpClient := new(mocks.MockOpAMPClient)
 				mockOpAmpClient.On("SetAgentDescription", mock.Anything).Return(nil)
 				mockStateProvider := new(mocks.MockPackagesStateProvider)
-				mockStateProvider.On("LastReportedStatuses").Return(nil, nil)
+				mockStateProvider.On("LastReportedStatuses").Return(packageStatuses, nil)
+				mockStateProvider.On("SetLastReportedStatuses", mock.Anything).Return(nil).Run(func(args mock.Arguments) {
+					status := args.Get(0).(*protobufs.PackageStatuses)
+					assert.Equal(t, "Failed creating TLS config: failed to read CA file: open bad-ca.cert: no such file or directory", status.Packages[packagestate.CollectorPackageName].ErrorMessage)
```
BinaryFissionGames, line 221: Also, this is failing because the specific error message for a missing file differs between Windows and linux, I think we should just assert a prefix here.

cpheps (approved):
Cleanup LGTM. With the diff it made it a bit hard to tell if any logic had been change. Can we verify this works with latest production BP OP before merging?

opamp/observiq/observiq_client.go
StefanKurek: I think yes and no? If we can't write to package statuses but we send status (success or failure) to bindplane, if we were to shut down and start up we'd have a bad last status. Maybe it's not the worst thing in the world and would possibly self correct? I don't know if there is a "right" answer here.

StefanKurek:
> Cleanup LGTM. With the diff it made it a bit hard to tell if any logic had been change. Can we verify this works with latest production BP OP before merging?

@cpheps I'll test before I merge. Thanks!

PR #636, opened 2022-08-10
fix: Plugin Receiver Panics when Telemetry Set

cpheps:
### Proposed Change
- Fixed issue when config contains a `telemetry` section for the `service`. Hardcoded in the `telemetry` to off for internal collector created by plugin receiver.
- Renamed `test` folder to `testdata` to fit pattern in the rest of the project.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


receiver/pluginreceiver/rendered_config_test.go
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      http://www.apache.org/licenses/LICENSE-2.0
+//	http://www.apache.org/licenses/LICENSE-2.0
```
cpheps, line 7: I think my IDE may have replaced tabs with spaces here. I'll remove this file from the PR.

receiver/pluginreceiver/rendered_config.go
```diff
@@ -48,6 +49,16 @@ type PipelineConfig struct {
 	Exporters  []string `yaml:"exporters,omitempty"`
 }
 
+// TelemetryConfig is a representation of collector telemetry settings
+type TelemetryConfig struct {
```
StefanKurek, line 53: @cpheps what is this config meant for exactly? Sounds like something to do with recording metrics?

cpheps: This is to mimic the [Telemetry Config](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/telemetry/config.go#L24). We can't use the configs in OTel as they don't have `yaml` struct tags. We need these tags as we "write out" the config before passing it into the new collector. This hardcoded telemetry config will be used to turn off telemetry in the new collector this receiver creates.

We actually did this in #379 but I think this change was lost in a later refactor.

JonathanWamsley (approved):

StefanKurek (approved):

PR #639, opened 2022-08-10
fix: Error out of linux script if package install fails

BinaryFissionGames:
### Proposed Change
* Add error_exit if the linux script fails to unpack the package

Tested this using the URL flag pointing to a local tarball instead of an actual package.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


schmikei (approved):
Was able to test on a focal VM and worked as expected 👍

cpheps (approved):

PR #642, opened 2022-08-11
fix(logging): Write default logging yaml if not specified

BinaryFissionGames:
### Proposed Change
* When the `--logging` flag is not specified, write out a default stdout config
  * This fixes an issue where the logging.yaml would not exist on the system, so managed mode would fail when attempting to report the config to BindPlane OP
* Make specifying a logging.yaml fail if the specified yaml does not exist

Tested with a local build to ensure the logging.yaml was created locally on startup without specifying its location.

The default logging.yaml looks like this:
```yaml
output: stdout
level: info
```

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -121,7 +121,7 @@ func getDefaultLoggingConfigPath() string {
 	if ok {
 		return lp
 	}
-	return "./logging.yaml"
+	return logging.DefaultConfigPath
 }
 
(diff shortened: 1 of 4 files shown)


cpheps (approved):

schmikei (approved):
LGTM, worked great when I was testing it as well

StefanKurek (approved):

PR #643, opened 2022-08-12
fix(plugin): Make start_at description consistent for common_event_format_logs plugin

BinaryFissionGames:
### Proposed Change
* Make start_at description consistent with other plugin's start_at descriptions for common_event_format_logs plugin

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/common_event_format_logs.yaml
+++ b/plugins/common_event_format_logs.yaml
@@ -20,7 +20,7 @@ parameters:
     default: UTC
   - name: start_at
     type: string
-    description: Start reading file from 'beginning' or 'end'
+    description: At startup, where to start reading logs from the file (`beginning` or `end`)
     supported:
       - beginning
       - end


cpheps (approved):

schmikei (approved):
Approving because this makes sense, but do any of the plugins we have ingest logs via a stream or non-file based? I don't think currently but something I was just pondering.

BinaryFissionGames:
> Approving because this makes sense, but do any of the plugins we have ingest logs via a stream or non-file based? I don't think currently but something I was just pondering.

We do have some non-file based ones ones (tcp & udp plugins, syslog, vcenter, the cisco ones, etc).

schmikei:
> > Approving because this makes sense, but do any of the plugins we have ingest logs via a stream or non-file based? I don't think currently but something I was just pondering.
> 
> We do have some non-file based ones ones (tcp & udp plugins, syslog, vcenter, the cisco ones, etc).

Ah yeah they just don't have a `start_at` parameter, forgot about that in the moment. Thanks for bearing with me on that!

PR #644, opened 2022-08-12
feat: Set ENV to path of collector config.yaml

dmikolay:
### Proposed Change
- Added function for getting default collector config paths
- Will use an ENV variable if it is set
- If not will use default (./config.yaml)

##### Checklist
- [X] Changes are tested
- [ ] CI has passed

### DIFF
--- a/cmd/collector/main_test.go
+++ b/cmd/collector/main_test.go
@@ -31,8 +31,7 @@ import (
 func TestGetDefaultCollectorConfigPathENV(t *testing.T) {
 	fakeConfigPath := "./fake/path/config.yaml"
 
-	os.Setenv(configPathENV, fakeConfigPath)
-	defer os.Unsetenv(configPathENV)
+	t.Setenv(configPathENV, fakeConfigPath)
 
 	expected := []string{fakeConfigPath}
 	actual := getDefaultCollectorConfigPaths()
--- a/cmd/collector/main_test.go
+++ b/cmd/collector/main_test.go
@@ -29,10 +29,12 @@ import (
 )
 
 func TestGetDefaultCollectorConfigPathENV(t *testing.T) {
-	os.Setenv(configPathENV, "./fake/path/config.yaml")
+	fakeConfigPath := "./fake/path/config.yaml"
+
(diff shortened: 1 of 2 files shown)


cmd/collector/main_test.go
```diff
@@ -28,6 +28,23 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+func TestGetDefaultCollectorConfigPathENV(t *testing.T) {
+	os.Setenv(configPathENV, "./fake/path/config.yaml")
```
cpheps, line 32: nit: This string literal `"./fake/path/config.yaml"` is used twice. We can just store it in an variable so it's harder to mess up if we ever change this.

cpheps (approved):
LGTM besides linter errors

cmd/collector/main_test.go
dmikolay: Good call, I went in and changed it

cmd/collector/main_test.go
```diff
@@ -28,6 +28,25 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+func TestGetDefaultCollectorConfigPathENV(t *testing.T) {
+	fakeConfigPath := "./fake/path/config.yaml"
+
+	os.Setenv(configPathENV, fakeConfigPath)
+	defer os.Unsetenv(configPathENV)
```
BinaryFissionGames, line 35: Nit: Can we use https://pkg.go.dev/testing#T.Setenv instead? Does basically the same thing as these two lines, just built into the testing.T.

dmikolay: I just used this because Mitch did too later in the test file, should I change those as well?

cpheps: I didn't know this was a thing and it does the `Unsetenv` for you. Looks like it was new in 1.17

cpheps: Outside the scope of this but we might want to update other ENV tests to use this if they aren't already.

BinaryFissionGames (approved):

PR #646, opened 2022-08-12
feat(dev): Scripts to help automate otel updates

BinaryFissionGames:
### Proposed Change
* Add scripts + make target for updating OTel core and contrib dependencies
  * example usage: `make update-otel OTEL_VERSION=v0.58.0`

I wrote these while updating OTel this time around to help make it simpler in the future. I would like to also enhance these to update operations-collector too in the future.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Makefile
```diff
@@ -165,6 +165,16 @@ add-license:
...
 
+# update-otel attempts to update otel dependencies in go.mods,
+# and update the otel versions in the docs.
+# Currently, does not handle operations-collector updates.
+# Usage: make update-otel OTEL_VERSION=vx.x.x
+.PHONY: update-otel
+update-otel:
+	./buildscripts/update-otel.sh "$(OTEL_VERSION)"
+	./buildscripts/update-docs.sh "$(OTEL_VERSION)"
+	$(MAKE) tidy
```
cpheps, line 176: Looks like your `update-otel.sh` runs tidy. We should either only do it here so it gets all modules (and change to `compat=1.18`).

BinaryFissionGames: Yeah, good callout. I was a little zealous with it.

I did run into an issue where the edits to the go.mod didn't seem to work because the go mod wasn't tidied before hand? Not sure if that was necessarily the root cause, or if I just misunderstood something there.

cpheps (approved):

schmikei (approved):

PR #649, opened 2022-08-15
feat: Set ENV to path of collector manager.yaml

dmikolay:
### Proposed Change
- Added function for getting default collector manager path
- Will use an ENV variable if it is set
- If not will use default (./manager.yaml)

##### Checklist
- [X] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):

schmikei (approved):
LGTM

PR #658, opened 2022-08-24
fix: Container image default logging configuration

jsirianni:
### Proposed Change

The logging configuration is created on collector startup, if it does not exist. This presents an issue with the container image, as the collector process does not have permission to write to `./logging.yaml` in its working directory.

There are several ways to resolve this, but I think the best solution is to just bake a logging.yaml into /etc/otel and point to it with the new environment variable.

This changes resolves issues that I am observing with 1.7.0 in Kubernetes and OpenShift. 

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,9 @@ RUN adduser \
     --uid 10005 \
     otel
 
+RUN echo "output: stdout\nlevel: info\n" > /etc/otel/logging.yaml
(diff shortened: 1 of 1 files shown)


schmikei (approved):
I'm in favor of this approach

cpheps (approved):

PR #680, opened 2022-09-09
feat(plugin): Add Journald log driver support to container log plugin

jsirianni:
### Proposed Change

Some docker based Kubernetes clusters will use Journald as the log driver (default is `json-file`).  This change adds a parameter, `log_source`, for selecting `Journald` instead of the default `file`. 

This is not a breaking change, as users will need to explicitly enable Journald. Existing users will use the default value `file`.

The Journald directory is defaulted to `/var/log/journal` as that was the Journald directory in use when testing this change against an OpenShift cluster using Journald for the log driver. 

The Journald source maps correctly to Open Telemetry semantic conventions, and shows up in Google mapped to the `k8s_container` monitored resource type.

Screen Shot 2022-09-09 at 11 00 20 AM

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/plugins/container_logs.yaml
+++ b/plugins/container_logs.yaml
@@ -2,11 +2,22 @@ version: 0.0.1
 title: Kubernetes Container Logs
 description: Log parser for Kubernetes Container logs. This plugin is meant to be used with the OpenTelemetry Operator for Kubernetes (https://github.com/open-telemetry/opentelemetry-operator).
 parameters:
+  - name: log_source
+    description: Where to read container logs from
+    type: string
+    supported:
+      - file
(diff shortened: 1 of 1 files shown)


plugins/container_logs.yaml
```diff
@@ -2,11 +2,22 @@ version: 0.0.1
 title: Kubernetes Container Logs
```
cpheps, line 2: Can you bump the version number to `0.1.0` please?

cpheps (approved):

BinaryFissionGames (approved):

PR #684, opened 2022-09-12
feat(plugin): Support parsing container application logs as json

jsirianni:
### Proposed Change

Some Kubernetes container application logs are json, and look like this in their unparsed form:
```json
{
    "log":"{\"http.req.id\":\"d2ff7faf-db0b-4a5b-86c6-344eda2c5b8c\",\"http.req.method\":\"GET\",\"http.req.path\":\"/cart\",\"http.resp.bytes\":6650,\"http.resp.status\":200,\"http.resp.took_ms\":5,\"message\":\"request complete\",\"session\":\"bad3edc3-0974-41e0-9e5e-1427b11219d2\",\"severity\":\"debug\",\"timestamp\":\"2022-09-12T20:06:38.079934409Z\"}\n",
    "stream":"stdout",
    "time":"2022-09-12T20:06:38.080008097Z"
}
```

The output entry has `body` as a string, like this:

Screenshot from 2022-09-12 16-09-16

This change will parse body as json, and look like this:

Screenshot from 2022-09-12 16-10-03

Applications logs that are no json are skipped. This change applies to both input options (file and journald).

This change is enabled by default, but I would like some opinions on this.

Also fixed filelog receiver's `exclude` option.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/container_logs.yaml
+++ b/plugins/container_logs.yaml
@@ -24,6 +24,10 @@ parameters:
     default:
       # Excludes logs for the collector
       - "/var/log/containers/observiq-*-collector-*"
+  - name: body_json_parsing
+    type: bool
+    description: If the application log is detected as json, parse the values into the log entry's body.
+    default: true
   - name: start_at
     description: At startup, where to start reading logs from the file (`beginning` or `end`)
     type: string
@@ -44,7 +48,7 @@ template: |
         attributes:
           log_type: "k8s.container_logs"
         exclude:
-          {{ range $fp := .exclude_log_paths }}
+          {{ range $fp := .exclude_file_log_path }}
             - '{{ $fp }}'
           {{end}}
         poll_interval: 500ms
@@ -105,6 +109,15 @@ template: |
             id: log-to-body
             from: attributes.log
             to: body
+
+          # Parse body as json if the body appears to be json
+          {{ if .body_json_parsing }}
+          - type: json_parser
+            if: body matches "^{.*}\\s*$"
+            parse_from: body
+            parse_to: body
+          {{ end }}
+
           # Detect pod, namespace, and container names from the file name.
           - type: regex_parser
             regex: '^(?P<pod>[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)_(?P<namespace>[^_]+)_(?P<container>.+)-'
@@ -161,6 +174,14 @@ template: |
         - type: move
           from: body.MESSAGE
           to: body
+
+        # Parse body as json if the body appears to be json
+        {{ if .body_json_parsing }}
+        - type: json_parser
+          if: body matches "^{.*}\\s*$"
+          parse_from: body
+          parse_to: body
+        {{ end }}
 
         - type: add
           field: attributes.log_type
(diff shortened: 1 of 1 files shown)


plugins/container_logs.yaml
```diff
@@ -105,6 +109,15 @@ template: |
             id: log-to-body
             from: attributes.log
             to: body
+
+          # Parse body as json if the body appears to be json
+          {{ if .body_json_parsing }}
```
schmikei, line 114: I've been struggling with the idea of boolean flags for JSON vs Regex parsing for a while now. On one hand it probably is more performant than checking format of the log entry to see if it is JSON vs a regex-able log entry, however I think from a usability perspective it suffers as its another barrier for users to start using the plugin. Also if you have some pods use JSON logging and some use regex then the container logs are already incompatible and would need to be another plugin configuration etc. 

I'd like other perspectives but I'm of the camp that we emulate something like the mongodb plugin or we don't have the boolean parameter. 
https://github.com/observIQ/observiq-otel-collector/blob/fb99ec8ecf684a45a8898319c811e491bbcba21c/plugins/mongodb_logs.yaml#L41-L45

I don't want to be the only person to dictate that as a choice though, so I'm open to having a larger discussion.

schmikei: I don't want to diminish the performance impact but it is something that I've been thinking about quite a bit as far as headaches to start using plugins. For this case in a containerized environment it may be more complex to manually modify the plugins to optimize performance, so I can see that as an argument to keep the boolean flag. 

TLDR; think this is might be fine but just thought I'd share some of my internal conflictions/consistency with some of the boolean flag design 🙃

BinaryFissionGames: I think this case is a little odd/different.

This isn't really a regex parsing vs JSON parsing thing; There doesn't seem to be any further regex parsing, this is more just a toggle for optional extra parsing of the log message, if the log message happens to be JSON (IDK if this is often the case or not). It is also conditional similar to your example, when enabled (`if: body matches "^{.*}\\s*$"`).

I like the idea of being able to disable this if you want to squeak out extra performance and you know that your logs aren't JSON.

jsirianni: The json_parser's `if` condition will allow the operator to be skipped in the event that a log is not json. The only downside here is the clear performance penalty of checking every line before parsing or skipping.

From a usability perspective, there is no prerequisite for the user. They can have a mixed environment without running into issues.

Users that do not have json logs (or do not wish to parse the json logs) can simply disable this option. 

While testing this change, I used Google's microservice demo, which has a handful of services with differing log formats. I did not observe parsing errors, but I did observe all json logs being parsed to body.

schmikei: > I like the idea of being able to disable this if you want to squeak out extra performance and you know that your logs aren't JSON.

I get that so I'll resolve this. Just in general I was thinking of the concept of whether nested JSON or in general how we should approach it. I see now its a slippery spiral and this is a tad niche of a use case so I'll go ahead and resolve this though

cpheps (approved):

schmikei:
@BinaryFissionGames should this be a minor version bump? New functionality right?

BinaryFissionGames:
Yeah, @jsirianni could you bump this plugin to 0.2.0?

jsirianni:
Done :+1:

PR #694, opened 2022-09-21
deps: Update hostmetrics to latest

BinaryFissionGames:
### Proposed Change
* Update hostmetrics receiver to https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/8b183beccbf7ef11776ea676a98881cf1e28b876 - this solves a bug where the hostmetrics receiver would fail to start on some windows systems (this meant our installer would fail)
* Update operations collector to https://github.com/observIQ/opentelemetry-operations-collector/tree/update-otel-to-0.61.0
* Update influx2otel package to latest commit
* Fix instances of removed methods on pdata
 
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cpheps (commented):
Looks good can we update the docs that point the forks we have. I realize they'll likely be the same as the current ones but just so when users go to documentation they see it for the version we're running.

BinaryFissionGames:
@cpheps whoops, totally missed that step. Updated now.

cpheps (approved):

StefanKurek (approved):
I don't completely follow why we are now doing this in the processor_metrics.go
```
if _, ok := dp.Attributes().Get(attrName); !ok {
				value.CopyTo(dp.Attributes().PutEmpty(attrName))
			}
```
but overall seems ok to me

BinaryFissionGames:
> I don't completely follow why we are now doing this in the processor_metrics.go
> 
> ```
> if _, ok := dp.Attributes().Get(attrName); !ok {
> 				value.CopyTo(dp.Attributes().PutEmpty(attrName))
> 			}
> ```
> 
> but overall seems ok to me

The old `Insert` function was removed, so the CopyTo stuff is done because this seems like the only way to do it now.

The reason why I do that Get before-hand is because insert wouldn't overwrite values if they were already set, but doing this CopyTo method makes it so that the original attribute would get overwritten if it already existed in the attributes, so we don't want to overwrite if the attribute already exists.

There are tests for this that fail otherwise, since this is the documented behavior of the processor.

PR #701, opened 2022-09-23
feat: Added forward porting of legacy agent IDs to ULIDs

cpheps:
### Proposed Change
Added logic so that if an agent currently has an `agent_id` in the `manager.yaml` that is not a ULID the collector will generate a ULID and save that as its new `agent_id`.

This ID change does make BPOP think the collector is a new agent. All labels and config files are preserved so the agent will behave the same in BPOP it will just be a new entity. I plan to discuss with the BPOP team to see if they have any plan to auto clean up this legacy porting.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -199,8 +199,8 @@ func ensureIdentity(configPath string) error {
 		return fmt.Errorf("unable to interpret config file: %w", err)
 	}
 
-	// if they already have an AgentID, don't generate it
-	if candidateConfig.AgentID != "" {
+	// If the AgentID is not a ULID (legacy ID or empty) then we need to generate a ULID as the AgentID.
+	if _, err := ulid.Parse(candidateConfig.AgentID); err == nil {
 		return nil
 	}
 
--- a/cmd/collector/main_test.go
+++ b/cmd/collector/main_test.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


cmd/collector/main.go
```diff
@@ -199,8 +199,8 @@ func ensureIdentity(configPath string) error {
 		return fmt.Errorf("unable to interpret config file: %w", err)
 	}
 
-	// if they already have an AgentID, don't generate it
-	if candidateConfig.AgentID != "" {
+	// If the AgentID is not a ULID (legacy ID or empty) then we need to generate a ULID as the AgentID.
+	if _, err := ulid.Parse(candidateConfig.AgentID); err == nil {
 		return nil
 	}
```
BinaryFissionGames, line 205: Kind of a ridiculous question here, but have we considered marshalling the (binary) UUID into a ULID? I know it's not technically correct, but it would create a one-to-one mapping of the legacy ID to the new ULID, which bindplane-op could use to correlate during the update here.

BinaryFissionGames (approved):

JonathanWamsley (approved):

cmd/collector/main.go
cpheps: @BinaryFissionGames talked with BPOP team on this and they are ok with this creating a new agent. Since it'll likely only be a release or two it didn't seem worth having special code in both BPOP and the collector to mitigate this.

PR #704, opened 2022-09-29
fix: Pass through TLS settings to Snapshot Client

cpheps:
### Proposed Change
Fixed TLS issues when using Snapshot feature as TLS config wasn't being passed down to the Client.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/opamp/observiq/observiq_client.go
+++ b/opamp/observiq/observiq_client.go
@@ -85,8 +85,14 @@ func NewClient(args *NewClientArgs) (opamp.Client, error) {
 		return nil, fmt.Errorf("failed to create updaterManager: %w", err)
 	}
 
+	// Propagate TLS config to repoteManager agent
+	tlsCfg, err := args.Config.ToTLS()
+	if err != nil {
+		return nil, fmt.Errorf("failed creating TLS config: %w", err)
+	}
+
 	reportManager := report.GetManager()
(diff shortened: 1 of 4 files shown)


BinaryFissionGames (approved):

opamp/observiq/observiq_client.go
```diff
@@ -85,8 +85,14 @@ func NewClient(args *NewClientArgs) (opamp.Client, error) {
 		return nil, fmt.Errorf("failed to create updaterManager: %w", err)
 	}
 
+	// Propagate TLS config to repoteManager agent
```
andykellr, line 88:
```suggestion
	// Propagate TLS config to reportManager agent
```

opamp/observiq/observiq_client.go
```diff
@@ -85,8 +85,14 @@ func NewClient(args *NewClientArgs) (opamp.Client, error) {
...
+	tlsCfg, err := args.Config.ToTLS()
```
StefanKurek, line 89: does this just return nil if there's no tls config? (Edit: it must because the unit tests work)

StefanKurek (approved):

opamp/observiq/observiq_client.go
cpheps: Yes, nil is an acceptable TLS config so we just set it accordingly.

PR #708, opened 2022-10-04
feat: Add a new default resource filter to the Google exporter

jmwilliams89:
### Proposed Change
- Updates the Google exporter to use a new default resource filter.
  - Prior to this change, the exporter used an empty array as the default. If the user did not configure this field, the exporter would ignore all resource attributes.
  - After this change, the exporter will use an array of one item as the default. This item has an empty string as the prefix. If the user does not configure this field, the exporter will now include all resource attributes by default.
  - Users can still restrict all resource attributes by setting the field manually to an empty array. Or they can still configure the field as before.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/exporter/googlecloudexporter/config.go
+++ b/exporter/googlecloudexporter/config.go
@@ -17,6 +17,7 @@ package googlecloudexporter
 import (
 	"os"
 
+	"github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector"
 	gcp "github.com/open-telemetry/opentelemetry-collector-contrib/exporter/googlecloudexporter"
 	"go.opentelemetry.io/collector/config"
 	"go.opentelemetry.io/collector/processor/batchprocessor"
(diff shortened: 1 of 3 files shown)


cpheps (approved):

exporter/googlecloudexporter/config.go
```diff
@@ -86,6 +87,10 @@ func createDefaultGCPConfig() *gcp.Config {
 	config.UserAgent = defaultUserAgent
 	config.MetricConfig.Prefix = defaultMetricPrefix
 	config.LogConfig.DefaultLogName, _ = os.Hostname()
+
+	// Overwrites the default resource filter to match all resource attributes
+	defaultResourceFilter := collector.ResourceFilter{Prefix: ""}
+	config.MetricConfig.ResourceFilters = []collector.ResourceFilter{defaultResourceFilter}
```
BinaryFissionGames, line 93: Should we document that we do this somewhere?

jmwilliams89: I've updated the PR to include documentation that calls out this behavior.

BinaryFissionGames (approved):

PR #709, opened 2022-10-04
fix: Unique storage extension support to plugin receiver

armstrmi:
Signed-off-by: armstrmi <mitchell.armstrong@bluemedora.com>

### Proposed Change

Fix has been added to support unique storage extension directories. Previously, when multiple instances of the same plugin receivers utilized `filelog` receivers and the `file_storage` extension, it would cause the collector to fail and crash upon start up. 
The fix is appending the `file_storage` directory with the unique component ID upon render of the plugin. This prevents the receivers from attempting to create the same file, saving the collector from crashing. 

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


receiver/pluginreceiver/plugin.go
```diff
@@ -91,6 +100,32 @@ func (p *Plugin) ApplyDefaults(values map[string]any) map[string]any {
 	return result
 }
 
+// CheckExtensions checks to see if there exists any filestorage extensions
+// within the plugin config. If one exists, make a unique storage extension
+// directory based on the receiver component ID.
+func CheckExtensions(values map[string]any, id string) error {
```
cpheps, line 106: This function doesn't need to be exported.

receiver/pluginreceiver/plugin.go
```diff
@@ -91,6 +100,32 @@ func (p *Plugin) ApplyDefaults(values map[string]any) map[string]any {
...
+	for i, ext := range values {
+		if i == "file_storage" {
```
cpheps, line 108: This should be a [strings.HasPrefix](https://pkg.go.dev/strings#HasPrefix) as it's valid to have something like `file_storage/one`.

receiver/pluginreceiver/plugin.go
```diff
@@ -91,6 +100,32 @@ func (p *Plugin) ApplyDefaults(values map[string]any) map[string]any {
...
+			var cfg filestorage.Config
+			err := mapstructure.Decode(ext, &cfg)
+			if err != nil {
+				return err
+			}
+
+			name := strings.Replace(id, "/", "_", -1)
```
cpheps, line 115: Should use [strings.ReplaceAll](https://pkg.go.dev/strings#ReplaceAll) as it's the same as `-1`.

receiver/pluginreceiver/plugin.go
```diff
@@ -91,6 +100,32 @@ func (p *Plugin) ApplyDefaults(values map[string]any) map[string]any {
...
+			cfg.Directory = filepath.Join(cfg.Directory, name)
+			values[i] = map[string]any{"directory": cfg.Directory}
+			if _, err := os.Stat(cfg.Directory); os.IsNotExist(err) {
+				err = os.Mkdir(cfg.Directory, 0750)
+				if err != nil {
+					return err
+				}
+			}
```
cpheps, line 123: I don't think you need to stat I think you can just do a [os.MkdirAll](https://pkg.go.dev/os#MkdirAll) here and if it already exist it's a noop.

receiver/pluginreceiver/plugin_test.go
```diff
@@ -186,7 +186,7 @@ service:
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			result, err := tc.plugin.Render(tc.values)
+			result, err := tc.plugin.Render(tc.values, config.NewComponentID(config.LogsDataType))
```
cpheps, line 189: Can we add a test case for the `CheckExtensions` function? Is it possible to just mock out a config that we can verify the directory was created with the path expected?

receiver/pluginreceiver/plugin.go
BinaryFissionGames: I think the *IDEAL* thing to do here is to use https://github.com/open-telemetry/opentelemetry-collector/blob/b1e161415ab7838b82fb5336a6f52b61d914bd4f/config/identifiable.go#L57 and use the `Type` method to get the type; This way we know that it matches OTel parsing exactly.

plugins/sap_hana_logs.yaml
```diff
@@ -59,8 +60,11 @@ template: |
                 warn: w
                 error: e
                 fatal: f
-
+  extensions:
+    file_storage:
+      directory: /tmp/storage
```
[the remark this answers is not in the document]
cpheps: Yeah I missed this at first because it was SAP HANA but this should be a parameter with a default value of `OIQ_OTEL_COLLECTOR_STORAGE`. We do want to give customers the ability to set this without having to change that ENV if they aren't aware of it or don't want to.

receiver/pluginreceiver/plugin_test.go
```diff
@@ -492,3 +494,44 @@ func TestCheckParameters(t *testing.T) {
 		})
 	}
 }
+
+func TestCheckExtensions(t *testing.T) {
```
cpheps, line 498: nit: There's a bit of an unofficial Go standard when testing unexported functions to structure the test name as such `Test_{internal_function_name}`.

receiver/pluginreceiver/plugin_test.go
```diff
@@ -492,3 +494,44 @@ func TestCheckParameters(t *testing.T) {
...
+	tmpDir := "/tmp/dir"
```
cpheps, line 499: Can we use `t.TmpDir` as these tests won't work on Windows. That way the test structure will take care of cleaning it up too.

receiver/pluginreceiver/plugin.go
```diff
@@ -91,6 +100,29 @@ func (p *Plugin) ApplyDefaults(values map[string]any) map[string]any {
 	return result
 }
 
+func checkExtensions(values map[string]any, id string) error {
+	for i, ext := range values {
+		c, _ := config.NewComponentIDFromString(i)
+		if c.Type() == "file_storage" {
```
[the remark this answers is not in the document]
armstrmi: I see what you're saying, but in this case, the casting isn't coming from the parameter `id` its coming from the loop variable `i`. Since were looping through the map of extensions (of type any),  in which it was originally stored in, I pass in the map of extensions and index through each one to see if that key matches the component type. `id` stands for the plugin name/id that is going to be appended to the end of the directory. 

Although, now that you bring it up, it may be a bit confusing with parameter/variable names and may be worth reworking those to something else.

receiver/pluginreceiver/plugin.go
```diff
@@ -91,6 +100,29 @@ func (p *Plugin) ApplyDefaults(values map[string]any) map[string]any {
...
+			var cfg filestorage.Config
+			err := mapstructure.Decode(ext, &cfg)
+			if err != nil {
+				return err
+			}
+
+			name := strings.ReplaceAll(pluginID, "/", "_")
+			cfg.Directory = filepath.Join(cfg.Directory, name)
+			extensions[i] = map[string]any{"directory": cfg.Directory}
+			err = os.MkdirAll(cfg.Directory, 0750)
```
BinaryFissionGames, line 116: I tried running the tests locally, noticed I ended up with a literal `$OIQ_OTEL_COLLECTOR_HOME` directory in the test dir (it think the Validate tests run this for the default file_logs plugin). It looks like it's because we don't do environment variable substitution, so it just makes the directory literally.

We could do something like:
```go
err = os.MkdirAll(return os.Expand(s, func(s string) string {
    // Allow escaping with $$
    if s == "$" {
        return "$"
    }
    return os.Getenv(str)
}), 0750)
```
This is what the collector does to expand environment variables. I think this will work fine; If the user specifies a directory the parameter will end up expanding twice, unfortunately, but I think that's generally true of all plugin parameters.

receiver/pluginreceiver/plugin_test.go
```diff
@@ -492,3 +494,44 @@ func TestCheckParameters(t *testing.T) {
...
+		{
+			name:        "Invalid Extensions Decoding",
+			extenstions: map[string]any{"file_storage": "hello"},
+			expectedErr: errors.New("'' expected a map, got 'string'"),
+		},
+		{
+			name:        "Invalid Extensions Directory",
+			extenstions: map[string]any{"file_storage": map[string]any{"directory": "/invalid/directory"}},
+			expectedErr: errors.New(""),
+		},
```
BinaryFissionGames, line 520: This is failing on Windows; I think this is actually a valid directory, it just requires root permissions on mac/linux.

Not sure what do use here as an invalid directory, I'd be OK with just removing this test case.

BinaryFissionGames (approved):

PR #712, opened 2022-10-06
feat: OracleDB Metrics Plugin

cpheps:
### Proposed Change
Added a plugin for OracleDB Metrics.

Here is a small sample of a metric resulting from the plugin:

```
ResourceMetrics #0
Resource SchemaURL: 
ScopeMetrics #0
ScopeMetrics SchemaURL: 
InstrumentationScope  
Metric #0
Descriptor:
     -> Name: oracle.tablespace.size
     -> Description: The size of tablespaces in the database.
     -> Unit: by
     -> DataType: Sum
     -> IsMonotonic: false
     -> AggregationTemporality: AGGREGATION_TEMPORALITY_CUMULATIVE
NumberDataPoints #0
Data point attributes:
     -> db.system: STRING(oracle)
     -> state: STRING(free)
     -> database_id: STRING(1643340059)
     -> global_name: STRING(ORCL.C.OTEL-AGENT-DEV.INTERNAL)
     -> tablespace_name: STRING(SYSTEM)
     -> contents: STRING(PERMANENT)
StartTimestamp: 1970-01-01 00:00:00 +0000 UTC
Timestamp: 2022-10-06 17:50:47.960831095 +0000 UTC
Value: 6225920
Metric #1
Descriptor:
     -> Name: oracle.tablespace.size
     -> Description: The size of tablespaces in the database.
     -> Unit: by
     -> DataType: Sum
     -> IsMonotonic: false
     -> AggregationTemporality: AGGREGATION_TEMPORALITY_CUMULATIVE
NumberDataPoints #0
Data point attributes:
     -> db.system: STRING(oracle)
     -> state: STRING(free)
     -> database_id: STRING(1643340059)
     -> global_name: STRING(ORCL.C.OTEL-AGENT-DEV.INTERNAL)
     -> tablespace_name: STRING(SYSAUX)
     -> contents: STRING(PERMANENT)
StartTimestamp: 1970-01-01 00:00:00 +0000 UTC
Timestamp: 2022-10-06 17:50:47.960831095 +0000 UTC
Value: 69140480
```

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/oracledb_metrics.yaml
+++ b/plugins/oracledb_metrics.yaml
@@ -647,6 +647,7 @@ template: |
     metricstransform:
       transforms:
         - include: .*
+          match_type: regexp
           action: update
           operations:
             - action: update_label
--- a/plugins/oracledb_metrics.yaml
+++ b/plugins/oracledb_metrics.yaml
@@ -643,7 +643,38 @@ template: |
                     )
                   )
                   GROUP BY DATABASE_ID, GLOBAL_NAME, INST_ID
+  processors:
+    metricstransform:
+      transforms:
+        - include: .*
+          action: update
+          operations:
+            - action: update_label
+              label: 'DATABASE_ID'
+              new_label: 'database_id'
+            - action: update_label
+              label: 'GLOBAL_NAME'
+              new_label: 'global_name'
+            - action: update_label
+              label: 'INSTANCE_ID'
+              new_label: 'instance_id'
+            - action: update_label
+              label: 'TABLESPACE_NAME'
+              new_label: 'tablespace_name'
+            - action: update_label
+              label: 'CONTENTS'
+              new_label: 'contents'
+            - action: update_label
+              label: 'STATUS'
+              new_label: 'status'
+            - action: update_label
+              label: 'PROGRAM'
+              new_label: 'program'
+            - action: update_label
+              label: 'WAIT_CLASS'
+              new_label: 'wait_class'
   service:
     pipelines:
       metrics/oracledb:
         receivers: [sqlquery/oracledb]
+        processors: [metricstransform]
--- a/plugins/oracledb_metrics.yaml
+++ b/plugins/oracledb_metrics.yaml
@@ -1,0 +1,650 @@
+version: 0.0.1
+title: OracleDB Metrics
+description: Metrics receiver for OracleDB
+parameters:
+  - name: endpoint
+    description: Address to scrape metrics from
+    type: string
+    default: localhost:1521
+  - name: username
+    description: Username to run metric queries with
+    type: string
+    required: true
+  - name: password
+    description: Password for user
+    type: string
+    required: true
+  - name: scrape_interval
+    description: Time in between every scrape request
+    type: string
+    default: 60s
+  - name: sid
+    description: Site Identifier. sid or service_name must be specified both can be.
+    type: string
+  - name: service_name
+    description: OracleDB Service Name. sid or service_name must be specified both can be.
+    type: string
+  - name: wallet
+    description: OracleDB Wallet file location
+    type: string
+template: |
+  # construct datasource endpoint based on config
+  {{$datasource := "oracle://"}}
+  {{ $datasource = printf "%s%s" $datasource .username }}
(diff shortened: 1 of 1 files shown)


plugins/oracledb_metrics.yaml
```diff
@@ -0,0 +1,649 @@
...
+  {{$datasource := "oracle://"}}
+  {{ $datasource = printf "%s%s" $datasource .username }}
+  {{ if .password }}
+    {{ $datasource = printf "%s:%s" $datasource .password }}
+  {{end}}
+  {{ $datasource = printf "%s@%s" $datasource .endpoint }}
+  {{ if .service_name }}
+    {{ $datasource = printf "%s/%s" $datasource .service_name }}
+  {{end}}
+        {{ if and .sid .wallet }}
```
BinaryFissionGames, line 40:
```suggestion
{{ if and .sid .wallet }}
```

plugins/oracledb_metrics.yaml
```diff
@@ -0,0 +1,649 @@
...
+    {{ $datasource = printf "%s?SID%%3D%s&WALLET%3D%s" $datasource .sid .wallet}}
```
BinaryFissionGames, line 41: I think this is missing an escape?
```suggestion
    {{ $datasource = printf "%s?SID%%3D%s&WALLET%%3D%s" $datasource .sid .wallet}}
```

plugins/oracledb_metrics.yaml
```diff
@@ -0,0 +1,649 @@
...
+              value_type: int
+          sql: SELECT (SELECT DBID FROM SYS.GV_$$DATABASE) DATABASE_ID, (SELECT GLOBAL_NAME FROM sys.GLOBAL_NAME) GLOBAL_NAME, CONTENTS, STATUS, COUNT(*) COUNT FROM sys.dba_tablespaces GROUP BY STATUS, CONTENTS
+        - metrics:
+            - attribute_columns:
+                - DATABASE_ID
+                - GLOBAL_NAME
+              data_type: sum
+              description: The number of seconds since the last RMAN backup.
+              metric_name: oracle.backup.latest
+              monotonic: "true"
```
BinaryFissionGames, line 111: Does this map to the bool true? I know that OTel does surprising things sometimes when they map to the config structs, so I'm just curious if this string is actually mapped to the bool value or not.

plugins/oracledb_metrics.yaml
```diff
@@ -0,0 +1,649 @@
...
+                        'user commits' USER_COMMITS,
+                        'user rollbacks' USER_ROLLBACKS
+                      )
+                    )
+                  )
+                  GROUP BY DATABASE_ID, GLOBAL_NAME, INST_ID
+  service:
+    pipelines:
+      metrics/oracledb:
+        receivers: [sqlquery/oracledb]
```
BinaryFissionGames, line 649: ops-agent adds a processor at the end to rename the uppercase labels to lowercase; Do we want that behavior in our collector or are we OK with the all-caps?

BinaryFissionGames (approved):
This looks functionally correct to me, just had some questions on whether we wanted to match the ops-agent output; If not, LGTM.

plugins/oracledb_metrics.yaml
cpheps: It's a bool in the sqlqueryreceiver config [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6dc71bd739b4eaa407a9a596dc57ca4884e6313e/receiver/sqlqueryreceiver/config.go#L77). I'm guessing map structure attempts a string to bool conversion.

plugins/oracledb_metrics.yaml
cpheps: Yeah I omitted that on purpose because it was ops-agent specific behavior. I am wondering though, probably is worth adding that processor

cpheps: @BinaryFissionGames added mind just doing a double check on the new processor addition?

BinaryFissionGames (approved):

PR #713, opened 2022-10-06
fix: Update plugin receiver file storage extension

armstrmi:
### Proposed Change

Updated plugin receivers that use filelog receiver to utilize `file_storage` extension.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


plugins/scratch.yaml
```diff
@@ -0,0 +1,40 @@
+receivers:
```
BinaryFissionGames, line 1: I think this file was added by accident

BinaryFissionGames (approved):
LGTM

PR #716, opened 2022-10-11
fix: Fix plugins that had invalid stanza component configs

cpheps:
### Proposed Change
Since OTel v0.61.0 there is strict parsing of the Stanza based receivers. This caused a failure of some of the plugins using those receivers. Fixed those receivers and added tests to render plugin configs and run through through component parsing.

##### Checklist
- [X] Changes are tested
- [ ] CI has passed


plugins/common_event_format_logs.yaml
```diff
@@ -50,7 +50,6 @@ template: |
         - type: csv_parser
           parse_from: attributes.cef_headers
           header: 'version|device_vendor|device_product|device_version|signature_id|name|severity|extensions'
```
cpheps, line 52: I didn't bump the version of this plugin as the field I removed wasn't being used anyway so the way this plugin functioned didn't change.

receiver/pluginreceiver/supplied_plugins_test.go
```diff
@@ -41,17 +63,46 @@ func TestValidateSuppliedPlugins(t *testing.T) {
...
+
+				switch componentID.Type() {
+				case windowseventlogreceiver.NewFactory().Type():
+					if runtime.GOOS != "windows" {
+						return
+					}
+				case journaldreceiver.NewFactory().Type():
+					if runtime.GOOS != "linux" {
+						return
+					}
```
JonathanWamsley, line 92: Should there be `darwin` case or a default case triggering an error?

cpheps: I don't think so. The two components that are called out on work on specific operating systems. There is no receiver that only works on mac. Though I think syslog won't work on windows, will need to add that in.

JonathanWamsley (approved):

PR #717, opened 2022-10-12
feat(plugin):  Add log_driver option to container log plugin

jsirianni:
### Proposed Change

- Added `log_driver` parameter with `auto`, `docker-json-file`, and `containerd-cri` options
  - `auto` is the default value and retains the same behavior (detects log format on every log)
  - By selecting a log driver, the pipeline will skip a regex match on every single log that enters the pipeline
- Fixed the regex used when checking if containerd cri format
  - Previous regex was working for the most part, but I had some application logs match despite coming from docker.
  - This [example](https://regex101.com/r/glMYc3/1) shows the previous containerd incorrectly regex matching a docker log.
- Renamed the parsers to better reflect what format they are processing
  - docker_parser --> docker_json_file_parser
  - containerd_parser --> containerd_cri_parser

These changes are backwards compatible, and do not change behavior unless the `log_driver` option is set to something other than the default `auto` value.

I have tested against the following log_driver options, outputing to Google Cloud Logging:

**docker**
- auto (no parsing failures)
- docker-json-file (no parsing failures)
- containerd-cri (parse failures on every entry, expected)

**containerd**
- auto (no parsing failures)
- docker-json-file (parse failures on every entry, expected)
- containerd-cri (no parsing failures)

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/container_logs.yaml
+++ b/plugins/container_logs.yaml
@@ -49,6 +49,7 @@ parameters:
       - docker-json-file
       # This format is not well documented: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/logs/logs.go#L125-L169
       - containerd-cri
+    default: auto
 
 
 template: |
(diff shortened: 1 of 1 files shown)


cpheps (approved):

schmikei:
@jsirianni We should probably bump the version to be v0.3.0

jsirianni:
> @jsirianni We should probably bump the version to be v0.3.0

Thank you, good catch.

schmikei (approved):

PR #718, opened 2022-10-12
fix(plugin): Hadoop filter expression

jsirianni:
### Proposed Change

I stumbled upon this in the wild. At some point between v1.9.0 and v1.9.2, the `expr` language changed. We now need to quote `true` when doing `expr: true`. I did verify that this is backwards compatible and is working as expected with v1.9.0 and v1.9.2.

The hadoop plugin is the only place we do this.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/hadoop_logs.yaml
+++ b/plugins/hadoop_logs.yaml
@@ -1,4 +1,4 @@
-version: 0.0.1
+version: 0.0.2
 title: Apache Hadoop
 description: Log parser for Apache Hadoop
 parameters:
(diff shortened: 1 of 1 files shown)


cpheps (approved):
Did you happen to do a quick find all in the receivers to see if this was anywhere else?

schmikei (approved):

jsirianni:
> 

I searched for `expr: true' and `expr: 'true'` and found one hit. I have used `expr: true` in other repos and have PRs out for them.

PR #724, opened 2022-10-14
deps: Changes collector dependency version to 1.10.0

StefanKurek:
### Proposed Change
Changes collector dependency version to 1.10.0

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):

cpheps (commented):
Looks like we've missed in a few releases but can we update the packagestate dependency inside the updater/go.mod?

StefanKurek:
> Looks like we've missed in a few releases but can we update the packagestate dependency inside the updater/go.mod?

@cpheps good catch. i was just searching for "1.9.0". fixed now

cpheps (approved):

PR #733, opened 2022-10-20
fix: Added hotfixed for collector that cleans up on service shutdown

cpheps:
### Proposed Change
Forked the otel collector and added in a fix for the collector not cleaning up if service.Start failed. Commit [here](https://github.com/observIQ/opentelemetry-collector/commit/b69664f75273a9a3af994288b24b3280e8cbc852).

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


schmikei (approved):
Did you have good steps to validate this change? Did a bunch of stops/starts with a variety of components and appears to be fine, but curious if there was further validation that you did?

cpheps:
>Did you have good steps to validate this change? Did a bunch of stops/starts with a variety of components and appears to be fine, but curious if there was further validation that you did?

I don't, we haven't been able to reliably reproduce this. It's been more of a "without this fix an error occurs 10% of the time and with the fix it never occurs". This is a bug we need fixed but we haven't been able to isolate it reliably.

PR #734, opened 2022-10-21
feat: Added Sampling processor

cpheps:
### Proposed Change
Added a simple processor for sampling data. The processor takes in a `drop_ratio` for configuration and roughly drops that percentage of payloads passed through it.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


processor/samplingprocessor/config.go
```diff
@@ -0,0 +1,42 @@
...
+type Config struct {
+	config.ProcessorSettings `mapstructure:",squash"`
+
+	// DropRatio is the ratio of payloads that are dropped. Values between 0.0 and 1.0 are valid.
+	DropRatio float64 `mapstructure:"drop_ratio"`
+}
+
+// Validate validates the processor configuration
+func (cfg Config) Validate() error {
+	// Validate drop ration
```
schmikei, line 36:
```suggestion
	// Validate drop ratio
```

processor/samplingprocessor/processor.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (tmp *throughputMeasurementProcessor) processLogs(_ context.Context, ld plog.Logs) (plog.Logs, error) {
+	//#nosec G404 -- randomly generated number is not used for security purposes. It's ok if it's weak
+	if rand.Float64() <= tmp.dropCutOffRatio {
+		return plog.NewLogs(), nil
+	}
+
+	return ld, nil
+}
```
BinaryFissionGames, line 55: Should we be doing this for each individual log record, instead of the whole plog.Logs? I think this would end up with the distribution being less uniform over shorter periods of time. Maybe that isn't a concern, though.

The metrics and traces processors make sense to drop whole payloads, I think (metrics could maybe drop per-ResourceMetric, but it makes sense to just eat the whole output of the scraper).

processor/samplingprocessor/processor.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (tmp *throughputMeasurementProcessor) processTraces(_ context.Context, td ptrace.Traces) (ptrace.Traces, error) {
+	//#nosec G404 -- randomly generated number is not used for security purposes. It's ok if it's weak
+	if rand.Float64() <= tmp.dropCutOffRatio {
+		return ptrace.NewTraces(), nil
+	}
+
+	return td, nil
+}
```
BinaryFissionGames, line 46: Note here, there is https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/probabilisticsamplerprocessor that does this for traces in contrib, but I don't really know what the differences are. The "probabilisticsampler" processor seems to have some extra considerations for distributed tracing stuff, so I'm not sure if that's 100% needed or not.

processor/samplingprocessor/processor.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (tmp *throughputMeasurementProcessor) processLogs(_ context.Context, ld plog.Logs) (plog.Logs, error) {
+	//#nosec G404 -- randomly generated number is not used for security purposes. It's ok if it's weak
+	if rand.Float64() <= tmp.dropCutOffRatio {
```
schmikei, line 50: May be a non-concern, but just confirming its all good that this is an approximation right? Just thinking how batching works, which is something I'm not entirely familiar with when it comes from receivers => processors. 

i.e. a drop ratio of .6 
1st emission: 100 logs - rand.Float64 == .5 (drop)
2nd emission: 50 logs - rand.Float64 == .7 (keep)
3rd emission: 100 logs - rand.Float64 == .45 (drop) 

would theoretically give a sample of logs that is only 20% of the true value right(50/250)? Statistics would probably eventually get us to the correct ratio given enough time but unless these go through a batch processor it may be slightly more inaccurate on first couple samples, which might be perfectly fine.

I believe this is probably fine as long as users know that this value is an estimated approximation. Just trying to wrap my head around an example and thinking through some cases.

processor/samplingprocessor/processor.go
cpheps: Yeah good call. I'll make that change.

processor/samplingprocessor/processor.go
cpheps: Yeah we don't need anything super fancy for this first pass. Mostly as long as it does rough sampling we can always tweak it down the road.

cpheps:
@BinaryFissionGames @schmikei Fixed up to remove base level objects. Tested this to verify it works as expected.

schmikei (approved):

BinaryFissionGames (approved):

PR #735, opened 2022-10-21
feat(plugin): BindPlane OP Logs

jsirianni:
### Proposed Change

Added plugin for [BindPlane OP](https://github.com/observIQ/bindplane-op/).
- Default path is `/var/log/bindplane/bindplane.log` on a linux system, which is the only supported install path at the moment.
- BindPlane rotates the log into gzip compressed files, which should always be excluded.
- Logs are always json encoded.
- Timestamps are ISO 8601.
- Severity parsing uses the default mapping.

Screenshot from 2022-10-21 16-53-27

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/plugins/bindplane_logs.yaml
+++ b/plugins/bindplane_logs.yaml
@@ -1,6 +1,6 @@
 version: 0.1.0
 title: BindPlane OP Logs
-description: Log parser for BindPlane OP observability pipeline.
+description: Log parser for BindPlane Observability Pipeline.
 parameters:
   - name: log_paths
     type: "[]string"
--- a/plugins/bindplane_logs.yaml
+++ b/plugins/bindplane_logs.yaml
@@ -1,0 +1,61 @@
+version: 0.1.0
+title: BindPlane OP Logs
+description: Log parser for BindPlane OP observability pipeline.
+parameters:
+  - name: log_paths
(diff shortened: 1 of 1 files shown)


plugins/bindplane_logs.yaml
```diff
@@ -0,0 +1,61 @@
...
+    default:
+      - "/var/log/bindplane/bindplane-*.log.gz"
+  - name: start_at
+    description: At startup, where to start reading logs from the file (`beginning` or `end`)
+    type: string
+    supported:
+      - beginning
+      - end
+    default: end
+template: |
```
cpheps, line 22: Can we add in the storage extension to ensure offsets are kept? See example in [solr](https://github.com/observIQ/observiq-otel-collector/blob/main/plugins/solr_logs.yaml#L23-L25). It should add the extension the [storage](https://github.com/observIQ/observiq-otel-collector/blob/main/plugins/solr_logs.yaml#L28) parameter and the plugin [parameter](https://github.com/observIQ/observiq-otel-collector/blob/main/plugins/solr_logs.yaml#L18-L21)

jsirianni: Done

cpheps (approved):

PR #748, opened 2022-10-27
feat: Adds plugin readme doc generation

StefanKurek:
### Proposed Change
Now can use `make create-plugin-docs` to generate documentation for all plugins. This is done with no extra info needed for almost every plugin. The oracledb_metrics plugin needs some extra details, so there is an `extra_details.yaml` file for custom stuff to throw in the "top", "bottom" and example parameters area of the documentation.

If this PR is too much to look at with all of the generated docs, I can remove these and add those in a separate PR.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


cmd/plugindocgen/templates/readme.tmpl
```diff
@@ -0,0 +1,52 @@
...
+    {{- end -}}
+    {{- if .ExtraDetails.CustomExampleParams }} 
+      {{ .ExtraDetails.CustomExampleParams -}} 
+    {{- end }}
+```
+{{- if .ExtraDetails.Bottom }}
+
+{{ .ExtraDetails.Bottom -}}
+{{ else }}
+{{ end -}}
```
cpheps, line 52: newline

cpheps (approved):
Looks great. Linter looks a bit mad just add a dummy package comment please.

Makefile
```diff
@@ -215,3 +215,7 @@ scan-licenses:
 .PHONY: generate
 generate:
 	$(MAKE) for-all CMD="go generate ./..."
+
+.PHONY: create-plugin-docs
+create-plugin-docs:
+	cd cmd/plugindocgen; go run *.go
```
BinaryFissionGames, line 221: This is super nitpicky and I don't even know if there's an actual difference so feel free to ignore this, but usually I see this:
```suggestion
	cd cmd/plugindocgen; go run .
```

BinaryFissionGames (commented):
This is missing docs for `bindplane_logs` it looks like. The generated docs look awesome.

StefanKurek:
aware of CI errors. working on them

StefanKurek:
> This is missing docs for `bindplane_logs` it looks like. The generated docs look awesome.
> 
> Edit to add: I think as a follow up in the future, we could add a CI job that makes sure there's no diff on PRs when running this tool.

@BinaryFissionGames I forgot to rerun the tool after rebasing from main. Should be fixed in the next push.

BinaryFissionGames (approved):

PR #760, opened 2022-11-03
fix: Fix asynchronous error causing zombie collector

BinaryFissionGames:
### Proposed Change
I booted up the collector in standalone mode, and got this error:
```
{"level":"error","ts":"2022-11-03T01:11:12.776-0400","caller":"service/collector.go:194","msg":"Asynchronous error received, terminating process","error":"listen tcp :8888: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted.","stacktrace":"go.opentelemetry.io/collector/service.(*Collector).Run\n\tC:/Users/brandon/go/pkg/mod/go.opentelemetry.io/collector@v0.63.1/service/collector.go:194\ngithub.com/observiq/observiq-otel-collector/collector.(*collector).Run.func1\n\tD:/projects/observiq-otel-collector/collector/collector.go:108"}
{"level":"info","ts":"2022-11-03T01:11:12.776-0400","caller":"service/service.go:115","msg":"Starting shutdown..."}
{"level":"info","ts":"2022-11-03T01:11:12.776-0400","caller":"pipelines/pipelines.go:118","msg":"Stopping receivers..."}
{"level":"info","ts":"2022-11-03T01:11:12.776-0400","caller":"pipelines/pipelines.go:125","msg":"Stopping processors..."}
{"level":"info","ts":"2022-11-03T01:11:12.776-0400","caller":"pipelines/pipelines.go:132","msg":"Stopping exporters..."}
{"level":"info","ts":"2022-11-03T01:11:12.777-0400","caller":"extensions/extensions.go:56","msg":"Stopping extensions..."}
{"level":"info","ts":"2022-11-03T01:11:12.777-0400","caller":"service/service.go:129","msg":"Shutdown complete."}
```
The collector process was still running in a zombied state at this point.

The error occurred because I had another collector up in the background, which makes sense, but the process should definitely exit if this happens.

After investigating, I found that the collector service does not return an error in this case, and the waitForStartup function never observes the collector running, so it forever waits for startup.

To fix this, we should signal waitForStartup even if err is nil, and return an error there if it doesn't exist.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -108,9 +108,11 @@ func (c *collector) Run(ctx context.Context) error {
 		err := svc.Run(ctx)
 		c.sendStatus(false, err)
 
-		if err != nil {
-			startupErr <- err
-		}
+		// The error may be nil;
+		// We want to signal even in this case, because otherwise waitForStartup could keep waiting
+		// for the collector startup, even though the collector will never start up.
+		// This can occur if an asynchronous error occurs quickly after collector startup.
+		startupErr <- err
 	}()
 
 	// A race condition exists in the OT collector where the shutdown channel
@@ -156,6 +158,11 @@ func (c *collector) waitForStartup(ctx context.Context, startupErr chan error) e
 			c.svc.Shutdown()
 			return ctx.Err()
 		case err := <-startupErr:
+			if err == nil {
+				// We want to report an error here, even if the error is nil, because we did not observe
+				// the collector actually start.
+				return fmt.Errorf("collector failed to start, and no error was returned")
+			}
 			return err
 		}
 	}


cpheps (approved):
Good catch, is there any way to unit test this? Seems like it might be hard without mocking out the collector.

collector/collector.go
```diff
@@ -156,6 +158,11 @@ func (c *collector) waitForStartup(ctx context.Context, startupErr chan error) e
 			c.svc.Shutdown()
 			return ctx.Err()
 		case err := <-startupErr:
+			if err == nil {
```
StefanKurek, line 161: @BinaryFissionGames i'm obviously missing something as i'm trying to wrap my head around these functions and changes. doesn't this just ensure that we return an error every single time run is called now?

cpheps (commented):
Looking at this with @StefanKurek and saw a race condition in `waitForStartup`. In the OTel collector it's state is set to `Running` [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L158) before the async error is ever detected.

It's likely rare this would happen but we could detect the state is running before we check the startupErr channel. I think the chances of this happening is low enough we don't need to worry.

collector/collector.go
BinaryFissionGames: It doesn't always return an error; This function has 3 exit points. It can exit if:
1. The startupErr channel is signalled. In this case, we attempted to run the collector, but the collector is no longer running; So the collector failed to start. We should always return a non-nil error in this case.
2. The context is cancelled (returns ctx error).
3. The service state is `Running` - in this case, no error is returned (https://github.com/observIQ/observiq-otel-collector/blob/129f7b1e36114eaf014dfea2e5698506d05ae9ae/collector/collector.go#L150)

StefanKurek (approved):

BinaryFissionGames:
We talked through the above race condition and found that it's OK because the status channel is notified, and the service listens on the service channel for changes.

PR #761, opened 2022-11-07
fix: Snapshot Race Condition Panic on Config Reload

cpheps:
### Proposed Change
Fixed a possible race condition panic that could occur by concurrent accesses of the snapshot buffer maps during a reconfiguring.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/internal/report/snapshot_reporter.go
+++ b/internal/report/snapshot_reporter.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"sync"
 
 	"github.com/observiq/observiq-otel-collector/internal/report/snapshot"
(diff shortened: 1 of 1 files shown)


StefanKurek (approved):
Look ok to me. What was the circumstance that was causing an issue? (or is this purely preventative?)

schmikei (approved):

cpheps:
>Look ok to me. What was the circumstance that was causing an issue? (or is this purely preventative?)
@StefanKurek  yeah, Nico ran into this issue during a reconfiguration in BPOP. When that occurs the buffers are cleared out but it's possible the OTel collector is still collecting and saving snapshots while it's clearing buffers.

PR #772, opened 2022-11-10
feat: Add project id detection to the Google exporter

jmwilliams89:
### Proposed Change
- This change adds project id extraction to the Google exporter. The upstream exporter already supports detecting project id based on automatic credentials. This change achieves parity when the `credentials` or `credentials_file` field is configured by the user.
- The exporter makes a best attempt at extracting the project id. If it fails to find a project id in the specified credentials, it will log the error, but it will not cause the exporter to explicitly fail.
- The exporter will not attempt to extract the project id for the user if they have already specified a project id in their config.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/exporter/googlecloudexporter/config.go
+++ b/exporter/googlecloudexporter/config.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector"
 	gcp "github.com/open-telemetry/opentelemetry-collector-contrib/exporter/googlecloudexporter"
@@ -109,7 +110,7 @@ func (c *Config) updateProjectFromJSON(jsonBytes []byte) error {
 }
 
 func (c *Config) updateProjectFromFile(fileName string) error {
-	jsonBytes, err := os.ReadFile(fileName)
+	jsonBytes, err := os.ReadFile(filepath.Clean(fileName))
 	if err != nil {
 		return fmt.Errorf("failed to read file: %w", err)
 	}


exporter/googlecloudexporter/config.go
```diff
@@ -69,6 +73,51 @@ func (c *Config) getClientOptions() []option.ClientOption {
...
+	if !ok {
+		return errors.New("project id is not a string")
+	}
+
+	c.GCPConfig.ProjectID = strValue
+	return nil
+}
+
+func (c *Config) updateProjectFromFile(fileName string) error {
+	jsonBytes, err := ioutil.ReadFile(fileName)
```
BinaryFissionGames, line 113: Nit: ioutil is deprecated, the functions were split between the os and io packages
```suggestion
	jsonBytes, err := os.ReadFile(fileName)
```

exporter/googlecloudexporter/config.go
```diff
@@ -69,6 +73,51 @@ func (c *Config) getClientOptions() []option.ClientOption {
...
+	if !ok {
+		return errors.New("project id is not a string")
+	}
+
+	c.GCPConfig.ProjectID = strValue
+	return nil
+}
+
+func (c *Config) updateProjectFromFile(fileName string) error {
+	jsonBytes, err := os.ReadFile(fileName)
```
BinaryFissionGames, line 113: Looks like gosec is complaining about this, I think the filename just needs to be passed through filepath.Clean first
https://securego.io/docs/rules/g304.html

schmikei (approved):
I was curious about if the collector was deployed to GCP if there was any funky things that could happen with the default credentials loading. 

Its an edge case but something I wished to validate the behavior of, so I spun up a vm instance and I found that the override does in fact get used with precedence over the default application creds. 

TLDR; did some more extraneous validation, seems to be expected behavior 🔥

PR #773, opened 2022-11-14
feat: Add mask processor

jmwilliams89:
### Proposed Change
- This adds a new mask processor to the collector. See the processor README for description and design.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


processor/maskprocessor/README.md
```diff
@@ -0,0 +1,79 @@
...
+| rules       | `email`: `\b[a-z0-9._%\+\-—\|]+@[a-z0-9.\-—\|]+\.[a-z\|]{2,6}\b`<br /><br />`ssn`: `\b\d{3}[- ]\d{2}[- ]\d{4}\b`<br /><br />`credit_card`: `\b(?:(?:(?:\d{4}[- ]?){3}\d{4}\|\d{15,16}))\b`<<br /><br />`phone`: `\b((\+\|\b)[1l][\-\. ])?\(?\b[\dOlZSB]{3,5}([\-\. ]\|\) ?)[\dOlZSB]{3}[\-\. ][\dOlZSB]{4}\b`| A series of key value pairs that define the masking rules of the processor. The key is the name of the rule. The value is the regex to mask. The regex engine used is [standard golang](https://pkg.go.dev/regexp/syntax). |
+| exclude     | `[]`    | A list of json dot notation fields that will be excluded from masking. The prefixes `resource`, `attributes`, and `body` can be used to indicate the root of the field. |
+
+### Example Config
+The following config is an example configuration of the mask processor using default values. This configuration will receive logs through an otlp receiver. The mask processor will then search and mask any logs that match the predefined email, ssn, credit_card, or phone rules. The logs will then be sent to the logging exporter.
+```yaml
+receivers:
+    otlp:
+        protocols:
+            grpc::
```
jsirianni, line 26: quick config fix
```suggestion
            grpc:
```

jsirianni (commented):
I played with this a little. It is working for my visa card, but not my amex with the following form
```
0000 000000 00000
0000-000000-00000
```

It does seem to work without dashes or spaces.

You can grab example numbers [here](https://webfiles.blackbaud.com/support/helpfiles/intsol/q42014/us/Content/BBNCCardTestDemo.html).

processor/maskprocessor/factory.go
```diff
@@ -0,0 +1,128 @@
...
+// createTracesProcessor creates a mask processor for traces.
+func createTracesProcessor(
+	ctx context.Context,
+	set component.ProcessorCreateSettings,
+	cfg config.Processor,
+	nextConsumer consumer.Traces,
+) (component.TracesProcessor, error) {
+	maskCfg, ok := cfg.(*Config)
+	if !ok {
+		return nil, errors.New("config is not of type maskprocessor.Config")
```
cpheps, line 70: Can we make this error a variable since it's reused?

processor/maskprocessor/processor.go
```diff
@@ -0,0 +1,191 @@
...
+
+	p.rules = rules
+	return nil
+}
+
+// processLogs masks incoming logs.
+func (p *processor) processLogs(_ context.Context, ld plog.Logs) (plog.Logs, error) {
+	for i := 0; i < ld.ResourceLogs().Len(); i++ {
+		resource := ld.ResourceLogs().At(i)
+		resourceMaskFunc := p.createMaskFunc(resourceField)
```
cpheps, line 66: Do we need to do this every iteration for every log. Seems like we could create this once and save it off as it's not creating the function based on any specific log.

exporter/googlecloudexporter/go.mod
```diff
@@ -46,7 +46,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.11.1 // indirect
 	go.opentelemetry.io/otel/trace v1.11.1 // indirect
 	go.uber.org/atomic v1.10.0 // indirect
-	go.uber.org/zap v1.23.0 // indirect
```
cpheps, line 49: Was this a result of a `go mod tidy`? Curious why this changed.

jmwilliams89: Yeah, it was a result of go mod tidy. I'm guessing go mod tidy wasn't used in the previous pr that updated the google exporter.

processor/maskprocessor/README.md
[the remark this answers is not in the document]
jmwilliams89: I think your eyes might be playing tricks on you. I'm having difficulty locating any errant tabs.

jmwilliams89: Good catch. This is addressed now.

license.yaml
```diff
@@ -39,3 +39,4 @@ exceptions:
     - path: "../packagestate"
     - path: "./opamp/observiq/testdata/latest"
     - path: "./processor/throughputmeasurementprocessor"
+    - path: "./processor/maskprocessor"
```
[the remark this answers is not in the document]
jmwilliams89: This should be added now.

processor/maskprocessor/config.go
```diff
@@ -0,0 +1,67 @@
...
+func (cfg Config) Validate() error {
+	if len(cfg.Rules) == 0 {
+		return errNoRules
+	}
+
+	for key, value := range cfg.Rules {
+		if _, err := regexp.Compile(value); err != nil {
+			return fmt.Errorf("rule '%s' does not compile as valid regex", key)
+		}
+	}
```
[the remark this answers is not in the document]
jmwilliams89: I went ahead and called CompileRules from validate.

processor/maskprocessor/factory.go
jmwilliams89: This is now a variable defined at the top of the file.

processor/maskprocessor/processor.go
jmwilliams89: I changed it so the processor has the resource and attributes range functions saved as fields. These fields are created when `Start` is called.

processor/maskprocessor/README.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+## How To
+### Add custom rules
+The following configuration adds a custom rule to mask any word greater than 10 characters. In this example, matching values will be replaced with `masked_long_word`.
+```yaml
+processors:
+    mask:
+        rules:
+            long_word: "\w{10,}"
```
BinaryFissionGames, line 46: I think we should single quote all of these example regexes; GitHub highlights these as red because they are invalid escape sequences, which makes this invalid YAML as-is.

processor/maskprocessor/factory.go
```diff
@@ -0,0 +1,129 @@
...
+		ctx,
+		set,
+		cfg,
+		nextConsumer,
+		processor.processTraces,
+		processorhelper.WithCapabilities(consumerCapabilities),
+		processorhelper.WithStart(processor.start))
+}
+
+// createTracesProcessor creates a mask processor for logs.
```
BinaryFissionGames, line 85:
```suggestion
// createLogsProcessor creates a mask processor for logs.
```

processor/maskprocessor/factory.go
```diff
@@ -0,0 +1,129 @@
...
+		ctx,
+		set,
+		cfg,
+		nextConsumer,
+		processor.processLogs,
+		processorhelper.WithCapabilities(consumerCapabilities),
+		processorhelper.WithStart(processor.start))
+}
+
+// createTracesProcessor creates a mask processor for metrics.
```
BinaryFissionGames, line 108:
```suggestion
// createMetricsProcessor creates a mask processor for metrics.
```

processor/maskprocessor/processor.go
```diff
@@ -0,0 +1,187 @@
...
+			for k := 0; k < scope.Metrics().Len(); k++ {
+				metrics := scope.Metrics().At(k)
+				switch metrics.Type() {
+				case pmetric.MetricTypeSum:
+					p.processSum(metrics.Sum())
+				case pmetric.MetricTypeGauge:
+					p.processGauge(metrics.Gauge())
+				case pmetric.MetricTypeSummary:
+					p.processSummary(metrics.Summary())
+				}
```
BinaryFissionGames, line 117: Any reason we don't support histogram & exponential histograms here?

processor/maskprocessor/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Supported pipelines
+- Logs
+- Metrics
+- Traces
+
+## How It Works
+1. This processor traverses the attribute and resource fields of incoming telemetry. For log-based telemetry, it will also traverse the body.
+2. If a field matches a defined regex rule, the matching value is replaced with `[masked_value]`, where `value` is the name of the  rule. For instance, a rule that masks email addresses would result in `[masked_email]`.
+3. If a field is to be excluded from masking, it can be specified in the processor's `exclude` field. By default, the processor will mask all fields.
```
BinaryFissionGames, line 12: Might be good to call out that this only works for string values (e.g. a social security number wouldn't be masked if it was encoded as an int for some reason)

processor/maskprocessor/processor.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	switch value.Type() {
+	case pcommon.ValueTypeMap:
+		maskFunc := p.createMaskFunc(field)
+		value.Map().Range(maskFunc)
+	case pcommon.ValueTypeStr:
+		p.maskString(value)
+	}
```
BinaryFissionGames, line 160: Is it worth supporting the "bytes" type here? I don't know of any receiver that uses it, probably only relevant if a user is using OTLP right now.

Also, maybe we want to iterate slices as well?

processor/maskprocessor/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Configuration
+| Field       | Default | Description |
```
Dylan-M, line 15: One of the new documentation standards items we've decided on is to include data type for these fields. It is the only thing "missing" from this doc based on the new standards Craig and I are writing. I realize that to some this is redundant, but not everyone is capable of inferring allowed data types from the examples.

processor/maskprocessor/processor.go
cpheps: @jmwilliams89 any thoughts on this? @BinaryFissionGames are those still "legacy" types? I thought at one point there was discussion of only using them if you absolutely have to. If that's still the case I think it's ok to skip them. This processor likely isn't going to be targeted at metrics as much as logs/traces.

processor/maskprocessor/processor.go
cpheps: I stubbed out handling bytes and it's a bit nasty as you have to rewrite the buffer byte by byte. I'm also not 100% confident it's appropriate to convert the bytes to a string then regex match on it and assume everything worked as intended. A string is more much straight forward. 

I think because the majority use case here is likely logs we can omit bytes and add later if a customer asks. @jmwilliams89 would like your input on this please.

As far as slices go I believe we can at least search a slice for strings and apply masking to those.

BinaryFissionGames: >  I'm also not 100% confident it's appropriate to convert the bytes to a string then regex match on it and assume everything worked as intended.

I was thinking that you would just use the regex functions that work on bytes (e.g. [Match](https://pkg.go.dev/regexp#Regexp.Match) and [ReplaceAll](https://pkg.go.dev/regexp#Regexp.ReplaceAll)), but if it's not actually as straight-forward, then it probably isn't a big deal to omit.

cpheps: Yes we can do that. I'm not sure if blindly applying a regex to a range of bytes is always safe in this case. Maybe it's being overly cautious but I'm not sure if we can always assume the bytes can be treated as a string.

processor/maskprocessor/README.md
cpheps: @Dylan-M Applied types to the readme.

processor/maskprocessor/processor.go
BinaryFissionGames: @cpheps I think it's OK to just ignore ExponentialHistogram, but maybe we should note in our docs that it's not supported?

BinaryFissionGames (approved):
LGTM, I think we should note that this doesn't support the ExponentialHistogram metric type, just in case someone wants to use this processor and is confused why it doesn't work with some metrics.

processor/maskprocessor/processor.go
cpheps: @BinaryFissionGames decided to just add the ExponentialHistogram type.

cpheps (approved):

PR #774, opened 2022-11-14
feat(container): Red Hat Universal base image

jsirianni:
### Proposed Change

For OpenShift deployments that require a RHEL certificate image, we will need a container image built on top of Red Hat's Universal base image.

- Moved Dockerfile --> docker/Dockerfile.ubuntu
- Added docker/Dockerfile.ubi8
- Added docker build config to goreleaser, with single `Minor }}.{{ .Patch }}-ubi8` tag
- Replaced `gcr.io/observiq-container-images/stanza-base:v1.2.2` with `ubuntu:22.04`
- Added `shell` and `home` options to `adduser` command

I opted to stop using the `stanza-base` image because I wanted to keep both Dockerfiles as identical as possible. This means the ubuntu image needs the `apt` commands for installing the systemd, tzdata, and ca-certificates packages. Stanza base image is nice because it pins our package versions in a nice way, but it is additional maintenance overhead that we can probably do without, and it does not work for rhel unless we re-work some things (Add a ubi8 base image). We could always do this in the future.

The images are pretty big, 509MB and 630MB (rhel). This is mainly from systemd and java's installation. We could probably justify building a minimal image in the future, which contains only the collector binary and plugins.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -180,9 +180,10 @@ nfpms:
       preinstall: ./scripts/package/preinstall.sh
       postinstall: ./scripts/package/postinstall.sh
 
-# Build container images with docker buildx (mutli arch builds)
+# Build container images with docker buildx (mutli arch builds).
 dockers:
-  - goos: linux
+  - id: ubuntu-amd64
+    goos: linux
     goarch: amd64
     ids:
       - collector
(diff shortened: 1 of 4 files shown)


cpheps (approved):
I think you might need to update this line in the Makefile to have a `Dockerfile.*`.

```
ALL_SRC := $(shell find . -name '*.go' -o -name '*.sh' -o -name 'Dockerfile' -type f | sort)
```

BinaryFissionGames (approved):

PR #793, opened 2022-11-16
fix: Added package level functions to reset internal telemetry views

cpheps:
### Proposed Change
In managed mode during a reconfiguration the throughput internal telemetry counters would not reset their values but rather just change labels. Added logic so when factories are recreated it resets the telemetry counters.

Screen shot showing the reset on a reconfigure:
Screen Shot 2022-11-16 at 8 35 03 AM

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/collector/settings.go
+++ b/collector/settings.go
@@ -15,6 +15,7 @@
 package collector
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/observiq/observiq-otel-collector/factories"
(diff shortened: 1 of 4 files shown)


BinaryFissionGames (approved):
For whatever reason, it feels unexpected to reset the views whenever DefaultFactories is called, but it doesn't seem like anywhere else would really be more appropriate. LGTM.

schmikei (approved):

PR #799, opened 2022-11-17
feat: Rollback panic recover

cpheps:
### Proposed Change
Added in logic to check for `.rollback` files when starting in managed mode. The assumption is if a `collector.yaml.rollback` file exist the collector crashed on reconfigure and was unable to rollback. The collector will overwrite the current config with the rollback config and clean up the rollback config.

If for some reason the current config is valid and we overwrite it with the rollback BPOP will send us down the config we should have.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):
LGTM; I assume there will be another PR for the other side of this logic that places the rollback file?

cpheps:
>I assume there will be another PR for the other side of this logic that places the rollback file?

@BinaryFissionGames Not sure what you mean. Are you referring to the current reload functionality that creates the reload function [here](https://github.com/observIQ/observiq-otel-collector/blob/main/opamp/observiq/reload_funcs.go#L110)?

BinaryFissionGames:
> @BinaryFissionGames Not sure what you mean. Are you referring to the current reload functionality that creates the reload function [here](https://github.com/observIQ/observiq-otel-collector/blob/main/opamp/observiq/reload_funcs.go#L110)?

Yep, totally missed that we already place the rollback file.

PR #801, opened 2022-11-18
fix(container): RHEL UBI logging.yaml

jsirianni:
### Proposed Change

Seeing some inconsistent behavior between the ubuntu based image and the rhel ubi image. The logging.yaml file is written incorrectly. This PR creates a `config/logging.stdout.yaml` logging config file and uses `COPY` to add it to each container image.

When running the container, I can see that the logging.yaml file is correct:
```
bash-4.4$ ls -la
total 24
drwxr-xr-x 1 otel otel 4096 Nov 18 15:12 .
drwxr-xr-x 1 root root 4096 Nov 18 15:12 ..
-rw-r--r-- 1 otel otel 3184 Nov 18 15:11 config.yaml
-rw-r--r-- 1 otel otel   28 Nov 18 15:11 logging.yaml

drwxr-xr-x 2 root root 4096 Nov 18 15:12 plugins
bash-4.4$ cat logging.yaml
output: stdout
level: info
```

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -245,7 +245,7 @@ dockers:
     extra_files:
       - plugins
       - config/example.yaml
-    - config/logging.stdout.yaml
+      - config/logging.stdout.yaml
   - id: ubi8-arm64
     goos: linux
     goarch: arm64
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -203,6 +203,7 @@ dockers:
     extra_files:
       - plugins
       - config/example.yaml
+      - config/logging.stdout.yaml
   - id: ubuntu-arm64
     goos: linux
     goarch: arm64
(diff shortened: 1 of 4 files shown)


schmikei (approved):

BinaryFissionGames (approved):
This LGTM, but I'm curious what the actual issue was? Was it just a permissions issue? 

Want to be sure this isn't a bug in the collector code somewhere.

jsirianni:
> This LGTM, but I'm curious what the actual issue was? Was it just a permissions issue?
> 
> Want to be sure this isn't a bug in the collector code somewhere.

Meant to clarify, no collector bug.

The RHEL image was not writing the logging.yaml correctly. 
```
bash-4.4$ cat logging.yaml
output: stdout\nlevel: info\n
```

The Ubuntu image has always been fine, and wrote:
```
output: stdout
level: info
```

It is probably better to copy a config from the repo anyway.

PR #805, opened 2022-11-21
Removed deprecated set output and used new reference variable for tag

cpheps:
### Proposed Change
Removed deprecated `set-output` from GitHub Actions. These were only used for setting the tag. I replaced this with `GITHUB_REF_NAME` and `github.ref_name`. These are newer ENVs that contain the short ref name that triggered the workflow.

More info [here](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables)

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,11 +62,8 @@ jobs:
           curl -f -L -o wix310-binaries.zip http://wixtoolset.org/downloads/v3.10.3.3007/wix310-binaries.zip
           unzip wix310-binaries.zip
         working-directory: C:/build
-      - name: Get Tag From Environment
-        id: get-tag
-        run: printf '::set-output name=tag::%s' "$(printf '%s' "${{ github.ref }}" | sed 's/refs\/tags\///')"
       - name: "Build MSI from Tagged Release"
-        run: go-msi.exe make -m observiq-otel-collector.msi --version ${{ steps.get-tag.outputs.tag }} --arch amd64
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -139,21 +136,18 @@ jobs:
...
-        id: get-tag
-        run: printf '::set-output name=tag::%s' "$(printf '%s' "${{ github.ref }}" | sed 's/refs\/tags\///')"
       - name: Trigger Installation Testing
         uses: peter-evans/repository-dispatch@v2
         with:
           token: ${{ secrets.ORG_GORELEASER_GITHUB_TOKEN }}
           repository: observIQ/otel-collector-installer-testing
           event-type: upstream_prerelease
-          client-payload: '{ "version": "${{ steps.get-tag.outputs.tag }}" }'
+          client-payload: '{ "version": "$GITHUB_REF_NAME" }'
```
BinaryFissionGames, line 145: This doesn't look like it runs in a shell, does this need to be `${{ github.ref_name }}` instead of "$GITHUB_REF_NAME"?

Same with the `Update Amplify Version` step.

cpheps: It's funny you ask I had that before because I wasn't sure. If you're also second guessing it I'll switch it back

BinaryFissionGames (approved):

JonathanWamsley (approved):

PR #822, opened 2022-12-01
fix(oracledb): Don't encode '=' in Oracle DB datasource

antonblock:
### Proposed Change
When building `$datasource`, use plain `=` instead of `%3D`

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/oracledb_metrics.yaml
+++ b/plugins/oracledb_metrics.yaml
@@ -1,4 +1,4 @@
-version: 0.0.1
+version: 0.1.0
 title: OracleDB Metrics
 description: Metrics receiver for OracleDB
 parameters:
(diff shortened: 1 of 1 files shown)


cpheps (approved):

plugins/oracledb_metrics.yaml
```diff
@@ -1,4 +1,4 @@
-version: 0.0.1
+version: 0.1.0
```
schmikei, line 1:
```suggestion
version: 0.0.2
```

This doesn't seem like a minor version update since no new functionality was added, seems like a fix :smile

schmikei (changes requested):
LGTM except I think this should be a patch fix not a minor, in my opinion. If someone else on the @observIQ/collector-team disagrees, feel free to comment

https://www.fileformat.info/info/unicode/char/003d/index.htm

plugins/oracledb_metrics.yaml
cpheps: @schmikei I told @antonblock to bump the minor version. This is the current way we've been versioning these. The very first version is `0.0.1` then the first change after that is `0.1.0`. I realize this is clear as mud and we should probably do a better more defined way of versioning this in the future.

schmikei (approved):

PR #823, opened 2022-12-02
feat: Add log count processor

jmwilliams89:
### Proposed Change
- Adds a processor for converting log counts into metrics. See readme for full implementation details.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


docs/processors.md
```diff
@@ -12,6 +12,7 @@ Below is a list of supported processors with links to their documentation pages.
 | Group by Attributes Processor           | [groupbyattrsprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.66.0/processor/groupbyattrsprocessor/README.md) |
 | Group by Trace Processor                | [groupbytraceprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.66.0/processor/groupbytraceprocessor/README.md) |
 | Kubernetes Attributes Processor         | [k8sattributesprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.66.0/processor/k8sattributesprocessor/README.md) |
+| Log Count Processor                          | [logcountprocessor](../processor/logcountprocessor/README.md) |
```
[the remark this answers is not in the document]
jmwilliams89: I debated on this. I went with Log Count, because it's the count of a log type, where the match expression is describing the type of log being counted.

processor/logcountprocessor/processor.go
```diff
@@ -0,0 +1,176 @@
...
+			continue
+		}
+		attrs[key] = value
+	}
+	return attrs
+}
+
+// extractResource extracts the resource from the record.
+func (p *processor) extractResource(record Record) map[string]interface{} {
+	value, ok := record[resourceField].(map[string]interface{})
```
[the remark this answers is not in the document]
jmwilliams89: The idea here is that resource is expected to be a map on the record. However, there's nothing to guarantee this key will exist or that everything will have a resource. I'm going to update this to return an empty map rather than nil.

processor/logcountprocessor/processor.go
```diff
@@ -0,0 +1,176 @@
...
+		attrExprs: attrExprs,
+		counter:   NewLogCounter(),
+		consumer:  consumer,
+		logger:    logger,
+	}
+}
+
+// Start starts the processor.
+func (p *processor) Start(_ context.Context, _ component.Host) error {
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
jmwilliams89: I wasn't sure how the collector operated with the context it passes in to this function. I didn't want it to prematurely kill the processor. I'm going to switch it to use this context and we'll test locally to see if it behaves as expected or dies early.

processor/logcountprocessor/processor.go
```diff
@@ -0,0 +1,176 @@
...
+	if err := sendMetrics(ctx, p.config.ID(), metrics); err != nil {
+		p.logger.Error("Failed to send metrics", zap.Error(err))
+	}
+}
+
+// matchRecord returns true if the record matches the configured expression.
+func (p *processor) matchRecord(record Record) bool {
+	matches, err := p.matchExpr.Match(record)
+	if err != nil {
+		p.logger.Debug("Failed to evaluate match expression", zap.Error(err))
```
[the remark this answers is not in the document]
jmwilliams89: I think we should keep the debug logging for now. We can reassess this if users find it noisy. I'd rather err on the side of caution and be able to troubleshoot potential problems.

processor/logcountprocessor/processor.go
```diff
@@ -0,0 +1,176 @@
...
+	return matches
+}
+
+// extractAttributes extracts attributes from the record.
+func (p *processor) extractAttributes(record Record) map[string]interface{} {
+	attrs := map[string]interface{}{}
+	for key, expression := range p.attrExprs {
+		value, err := expression.Evaluate(record)
+		if err != nil {
+			p.logger.Debug("Failed to evaluate attribute expression", zap.Error(err))
```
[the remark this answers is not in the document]
jmwilliams89: Same as above.

processor/logcountprocessor/config.go
```diff
@@ -0,0 +1,85 @@
...
+	defaultInterval = time.Minute
+
+	// defaultMatch is the default match expression.
+	defaultMatch = "true"
+)
+
+// Config is the config of the processor.
+type Config struct {
+	config.ProcessorSettings `mapstructure:",squash"`
+	Route                    string            `mapstructure:"route"`
```
cpheps, line 44: Can/do we need to do any validation that this actually exists?

processor/logcountprocessor/processor.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Capabilities returns the consumer's capabilities.
+func (p *processor) Capabilities() consumer.Capabilities {
+	return consumer.Capabilities{MutatesData: false}
+}
+
+// Shutdown stops the processor.
+func (p *processor) Shutdown(_ context.Context) error {
+	p.cancel()
```
cpheps, line 71: Do we need to wait for the `handleMetricsInternval` groutine to finish?

receiver/routereceiver/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## How It Works
+1. The user configures this receiver in a pipeline.
+2. The user configures a supported component to route telemetry to this receiver.
+
+## Configuration
+The route receiver does not have any configuration parameters. It simply receives telemetry from other components.
+
+### Example Config
+The following config is an example configuration of the route receiver. In this example, logs are collected from a file and sent to a log count processor. After each minute, the log counts are converted to metrics and sent to the route receiver in the metrics pipeline.
```
cpheps, line 17: Is it worth linking to the log count processor from here?

receiver/routereceiver/receiver.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// Start starts the receiver.
+func (r *receiver) Start(_ context.Context, _ component.Host) error {
+	return nil
+}
+
+// Shutdown stops the receiver.
+func (r *receiver) Shutdown(_ context.Context) error {
+	r.stopOnce.Do(func() {
```
cpheps, line 65: I don't think you need this in a stopOnce. `removeRoute` wraps [delete](https://pkg.go.dev/builtin#delete) which is a noop if there is nothing to delete. I'm not sure if this could be called up to three times if it's used in logs, traces, and metrics but even so two of those calls would be noops.

processor/logcountprocessor/config.go
jmwilliams89: Unfortunately not. This config will have no knowledge of outside components or configured pipelines. Receivers are also instantiated last in the pipeline. So at this point, there is no guarantee the receiver has been instantiated.

cpheps: Ok, looks like we'll probably log an error from the processor if it tries to route somewhere that doesn't exist?

jmwilliams89: Yeah exactly. We can't validate ahead of time. But we can log and warn the user when they attempt to route to something that doesn't exist.

receiver/routereceiver/README.md
jmwilliams89: Not yet. I think once we have some more components that use the route receiver, we include this list at the bottom of the readme. Maybe with a header that says "supported components."

receiver/routereceiver/receiver.go
jmwilliams89: Sounds good. I'll remove this then. I was following this pattern from the OTLP receiver which appears to do this to avoid unregistering the receiver 3 times. But I agree. It's unnecessary.

processor/logcountprocessor/processor.go
jmwilliams89: No. I think this should be okay. The only way that goroutine doesn't exit immediately is if it's in the process of sending metrics out. At which point it will exit immediately after that.

cpheps (approved):

processor/logcountprocessor/processor.go
BinaryFissionGames, line 71: I think I agree with Corbin that we should wait for the goroutine to exit here. I think we could end up with incorrect metrics being emitted in managed mode:
- Ticker goes off, sendMetrics is called
- Shutdown is called, processor is shutdown (does not wait for goroutine)
- New collector comes up, this processor was reconfigured to have a different expr
- sendMetrics emits metric for old logs in the pipeline, despite them not matching expr

That's an unlikely sequence just because the collector coming up is a long complex process, but I think it's still a good idea to protect against scenarios like that.

processor/logcountprocessor/processor.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// sendMetrics sends metrics to the consumer.
+func (p *processor) sendMetrics(ctx context.Context) {
+	p.mux.Lock()
+	defer p.mux.Unlock()
+
+	metrics := p.createMetrics()
+	p.counter.Reset()
+
```
schmikei, line 114:
```suggestion
	if metrics.MetricCount() == 0 {
		return
	}
```

Do you have any thoughts about not sending empty metrics when there is nothing to report? This can cause unnecessary exporter/processor overhead if there is nothing left to do.

processor/logcountprocessor/processor.go
```diff
@@ -0,0 +1,177 @@
...
+		attrExprs: attrExprs,
+		counter:   NewLogCounter(),
+		consumer:  consumer,
+		logger:    logger,
+	}
+}
+
+// Start starts the processor.
+func (p *processor) Start(ctx context.Context, _ component.Host) error {
+	processorCtx, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
jmwilliams89: @cpheps for reference. We originally debated on this, because previously, it was written the other way. I didn't think we could trust the passed in context on the chance that it was cancelled after everything was started.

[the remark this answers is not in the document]
jmwilliams89: @cpheps any concerns with switching the behavior back?

processor/logcountprocessor/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Expression Language
+In order to match or extract values from logs, the following `keys` are reserved and can be used to traverse the logs data model.
+
+| Key          | Description |
+| ---          | ---   |
+| `body`       | Used to access the body of the log. |
+| `attributes` | Used to access the attributes of the log. |
+| `resource`   | Used to access the resource of the log. |
+| `severity`   | Used to access the severity text of the log. |
```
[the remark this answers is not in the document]
jmwilliams89: I think it's best to hold off on severity number until we receive a request from users to expose it.

receiver/routereceiver/factory.go
```diff
@@ -0,0 +1,62 @@
...
+		createDefaultConfig,
+		component.WithMetricsReceiver(createMetricsReceiver, stability),
+		component.WithLogsReceiver(createLogsReceiver, stability),
+		component.WithTracesReceiver(createTracesReceiver, stability),
+	)
+}
+
+// createMetricsReceiver creates a metric receiver.
+func createMetricsReceiver(_ context.Context, _ component.ReceiverCreateSettings, cfg component.ReceiverConfig, consumer consumer.Metrics) (component.MetricsReceiver, error) {
+	receiver := createOrGetRoute(cfg.ID().Name())
```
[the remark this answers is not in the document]
jmwilliams89: This was the original implementation to register on start. However, @cpheps made the suggestion to switch it the create function.

processor/logcountprocessor/README.md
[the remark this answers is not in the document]
BinaryFissionGames: That's fair; I do think we should rename this though, I would assume `severity` refers to the otel severity of the log, like the one that gets mapped in GCP.

schmikei (commented):
Probably swimming in notifications, but had one more observation I believe is unaddressed/wanted to confirm the reasoning behind behavior

<details>
<summary> Config </summary>
```yaml
receivers:
    filelog:
        include: [./apache.log]
        start_at: beginning
    route/example:
processors:
    logcount:
        route: example
        interval: 10s

exporters:
    logging:
    file:
      path: ./out.json
service:
    pipelines:
        logs:
            receivers: [filelog]
            processors: [logcount]
            exporters: [logging]
        metrics:
            receivers: [route/example]
            processors: []
            exporters: [logging, file]
```

</details>
 
<details>
<summary> output </summary>
```json
{"resourceMetrics":[{"resource":{},"scopeMetrics":[{"scope":{"name":"logcount"},"metrics":[{"name":"log.count","unit":"{logs}","gauge":{"dataPoints":[{"timeUnixNano":"1670266420536256000","asInt":"1"}]}}]}]}]}
```
</details>

processor/logcountprocessor/processor.go
jmwilliams89: I'm having a difficult time tracking this scenario. If the component is shutdown and a new one is configured, wouldn't this new component have a new counter on the struct? Also, wouldn't the collector wait for all exporters to be shutdown before restarting the collector with the newly reconfigured components?

jmwilliams89: I went ahead and added a wait group.

processor/logcountprocessor/processor.go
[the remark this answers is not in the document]
jmwilliams89: I went ahead and switched the behavior back to defining a background context.

processor/logcountprocessor/README.md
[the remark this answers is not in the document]
jmwilliams89: I just went ahead and added both as options to avoid any confusion.

receiver/routereceiver/factory.go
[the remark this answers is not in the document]
jmwilliams89: One reason I don't think we'll be able to register in the start func now is if you include the route receiver in different telemetry pipelines, we need to register before this in order to return the same instance of the component for each pipeline. I believe this is the pattern currently in the OTLP receiver.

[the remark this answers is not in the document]
BinaryFissionGames: Sounds like there isn't a good way around this issue then; Thanks for the explanation!

processor/logcountprocessor/record.go
```diff
@@ -55,9 +58,10 @@ func convertToRecords(logs plog.Logs) []Record {
...
 	return Record{
-		resourceField:   resource,
-		attributesField: log.Attributes().AsRaw(),
-		bodyField:       log.Body().AsRaw(),
-		severityField:   log.SeverityText(),
+		resourceField:       resource,
+		attributesField:     log.Attributes().AsRaw(),
+		bodyField:           log.Body().AsRaw(),
+		severityTextField:   log.SeverityText(),
+		severityNumberField: log.SeverityNumber().String(),
```
BinaryFissionGames, line 65: Just one last comment here; Should this be the actual number instead of the string, so that comparison operators can be used against it?

jmwilliams89: Let me see if it's possible. I originally tried pulling that out as a number, but I didn't see any easy functions for providing that, so I defaulted to the string.

BinaryFissionGames: @jmwilliams89 You should be able to just cast it `int32(log.SeverityNumber())` (if I understand the docs right: https://pkg.go.dev/go.opentelemetry.io/collector/pdata/plog#SeverityNumber)

jmwilliams89: Looks like it gets returned as an int32, so I think we can use that. Do you think the severity text or the string representation of severity number are more useful for the user?

BinaryFissionGames (approved):

PR #841, opened 2022-12-09
Update Google Cloud Prometheus example config to use `generic_node`

jsirianni:
### Proposed Change

The Prometheus example results in two issues:
- Metrics show up in GCP as [generic_task](https://cloud.google.com/monitoring/api/resources#tag_generic_task) when we would prefer [generic_node](https://cloud.google.com/monitoring/api/resources#tag_generic_node).
- Metric labels do not align with Prometheus server. Prometheus server would have `job` and `instance` labels.

Solution: "Rename" `service.name` and `service.instance.id` to `job` and `instance`. This will prevent the `generic_task` mapping while also emitting metrics that look closer to what you would expect when coming from a Prometheus environment.

I also re-wrote the readme to support things like multiple scrape configs and custom metric prefix.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/config/google_cloud_exporter/prometheus/README.md
+++ b/config/google_cloud_exporter/prometheus/README.md
@@ -6,23 +6,95 @@ The Prometheus Receiver can be used to send prometheus style metrics scraped fro
 
 See the [prerequisites](../README.md) doc for Google Cloud prerequisites.
 
-## Configuration
+Ensure your applications support prometheus metrics, or have prometheus exporters configured.
 
-An example configuration is located [here](./config.yaml).
+## Usage
+
+Modify the example configuration's `prometheus` receiver. Update
+the `job_name` value to reflect your environment. Also update the
+`targets` value(s) to point to your application's prometheus exporter(s).
+
+For example, if you have five Nginx systems that need to be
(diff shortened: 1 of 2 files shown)


config/google_cloud_exporter/prometheus/README.md
```diff
@@ -6,23 +6,95 @@ The Prometheus Receiver can be used to send prometheus style metrics scraped fro
...
+        - targets:
+          - 'redis-0:9900'
+```
+
+### Deploy
 
 1. Copy [config.yaml](./config.yaml) to `/opt/observiq-otel-collector/config.yaml`
-2. Restart the collector: `sudo systemctl restart observiq-otel-collector`
+2. Modify the configuration to reflect your environment (See [Usage](./README.md#usage))
+3. Restart the collector: `sudo systemctl restart observiq-otel-collector`
```
cpheps, line 59: Do we assume linux for all our config examples?

jsirianni: I think so, we probably could include Windows at some point. The config path and restart commands would need to be changed or made "generic".

schmikei (approved):

PR #843, opened 2022-12-12
feat: Add metric extract processor

jmwilliams89:
### Proposed Change
- Added the metric extract processor. See README for implementation details.
- Refactored expression and extraction logic as an internal library

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


.github/dependabot.yml
```diff
@@ -97,6 +97,17 @@ updates:
     commit-message:
       prefix: "deps"
       include: "scope"
+  - package-ecosystem: "gomod"
```
cpheps, line 100: I think we need an entry for the `processor/metricextractprocessor` module also.

processor/metricextractprocessor/processor.go
```diff
@@ -0,0 +1,196 @@
...
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.uber.org/zap"
+)
+
+// processor is a processor that extracts metrics from logs.
+type processor struct {
```
cpheps, line 33: nit: Can we change the name to `metricExtractProcessor` or something more specific. The latest version of otel introduced the following packages `receiver`, `exporter`, `extension`. Which caused collision in some of our code where we had those same names for structs. I have a feeling at some point there will be a `processor` package as well.

cpheps (approved):

internal/expr/record.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// ResourceGroup is a group of records with the same resource attributes.
+type ResourceGroup struct {
+	Resource map[string]any
+	Records  []Record
+}
+
+// ConvertToResourceGroups converts plog.Logs to a slice of resource groups.
+func ConvertToResourceGroups(logs plog.Logs) []ResourceGroup {
```
StefanKurek, line 80: this makes me a little sad that it is so similar to the other function, but i also don't know if i have any good suggestions on how to refactor it that are worth it

StefanKurek (changes requested):

internal/expr/expression_test.go
```diff
@@ -0,0 +1,369 @@
...
+package expr
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestExpressionMatch(t *testing.T) {
```
StefanKurek, line 24: can get rid of this if it's true that the method under test isn't needed

internal/expr/map.go
```diff
@@ -0,0 +1,54 @@
...
+	expressions map[string]*Expression
+}
+
+// Extract extracts a value map from the record using the expression map.
+func (e *ExpressionMap) Extract(record Record) map[string]any {
+	results := map[string]any{}
+	for key, expression := range e.expressions {
+		value, err := expression.Evaluate(record)
+		if err != nil || value == nil {
+			continue
```
StefanKurek, line 32: should errors be treated the same as non matching expressions? should we have entries for non matching expressions in the results (i haven't actually looked at the code yet where this is used...more of a general question only about this function)?

internal/expr/expression.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/antonmedv/expr/vm"
+)
+
+// Expression is an Expression used to evaluate values.
+type Expression struct {
+	*vm.Program
+}
+
+// Match checks if an expression matches the supplied environment.
+func (e *Expression) Match(env map[string]any) (bool, error) {
```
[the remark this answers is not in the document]
jmwilliams89: It's used in both the logcount and metricextract processors to determine if a log record should be evaluated.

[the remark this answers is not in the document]
StefanKurek: @jmwilliams89 that's my sloppy reviewing and not checking the full code 😄

internal/expr/map.go
jmwilliams89: There's a distinction between matching records and extracting attributes. Even if a record is matched, it doesn't mean that all of the attributes the user defines will exist on that log record. In this case, if the user defines the location of one of their attributes as `body.embedded.field` and the `embedded` field doesn't exist, the evaluate expression will return an error. This shouldn't cause the rest of the attribute extraction to fail. Instead, it just means that attribute doesn't exist in the way the user defined and should be omitted.

processor/metricextractprocessor/config.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+const (
+	// defaultMetricName is the default metric name.
+	defaultMetricName = "extracted.metric"
+
+	// defaultMetricUnit is the default metric unit.
+	defaultMetricUnit = "{units}"
+
+	// defaultInterval is the default metric interval.
```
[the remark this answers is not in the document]
cpheps: Looks like it's only for unused function parameters

processor/metricextractprocessor/README.md
```diff
@@ -0,0 +1,93 @@
...
+```yaml
+processors:
+    metricextract:
+        match: attributes.status startsWith "4"
+        extract: body.byte_count
+        metric_name: byte.count
+```
+
+### Extract metric attributes
+The following configuration extracts the status and endpoint values from the body of a log. These values are used as metric attributes. For each unique combination observed, a unique metric count is created.
```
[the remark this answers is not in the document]
jmwilliams89: This is a mistake on my part. I started with the `logcount` processor README as a starting point, since the two processors share so much in common. This sentence was missed and should be removed.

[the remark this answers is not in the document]
StefanKurek: @jmwilliams89 so what is the intended behavior if there are multiple types of matches for an attribute across the logs?

internal/expr/map.go
StefanKurek: @jmwilliams89 in case it wasn't clear, i wasn't suggesting that no matches here should error. i just noticed that errors and no match were treated the same in that they weren't added to the results and was wondering if it was worth having a distinction (even if both should not cause a problem). sounds like no so i'll resolve

internal/expr/record.go
jmwilliams89: I'm personally okay with this level of naming. We could be more verbose and call it `ConvertToResourceGroupSlice`, but I'm not losing any sleep over it. If people feel strongly enough about it, I have no problem changing it.

processor/metricextractprocessor/README.md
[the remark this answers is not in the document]
StefanKurek: @jmwilliams89 i guess that makes sense and is fair...i still think it could be helpful as a reminder that there will be multiple metrics made as you had it, but i won't block on it and leave it up to you

processor/metricextractprocessor/factory.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// createLogsProcessor creates a log processor.
+func createLogsProcessor(_ context.Context, params component.ProcessorCreateSettings, cfg component.ProcessorConfig, consumer consumer.Logs) (component.LogsProcessor, error) {
+	processorCfg, ok := cfg.(*Config)
+	if !ok {
+		return nil, fmt.Errorf("invalid config type: %+v", cfg)
+	}
+
+	match, err := expr.CreateBoolExpression(processorCfg.Match)
```
StefanKurek, line 50: @jmwilliams89 minor and non blocking question - would these make more sense in the validation of the config?

processor/metricextractprocessor/processor.go
```diff
@@ -0,0 +1,196 @@
...
+
+	return metrics
+}
+
+// extractDataPoints extracts data points from the records.
+func (e *extractProcessor) extractDataPoints(records []expr.Record) pmetric.NumberDataPointSlice {
+	dataPoints := pmetric.NewNumberDataPointSlice()
+
+	for _, record := range records {
+		if e.match.MatchRecord(record) {
```
StefanKurek, line 122: non blocking preference: would be a little nicer if we continued on the inverse of this to have a little less nesting

StefanKurek (approved):
A few non blocking comments/questions. But OK to merge as is

processor/metricextractprocessor/processor.go
jmwilliams89: That's a great idea. I'll gladly make this change.

processor/metricextractprocessor/factory.go
jmwilliams89: Sure, I think we can include that.

StefanKurek: @jmwilliams89 now that you've put them in config validation, did you want to remove them from here?

StefanKurek: talked about this offline

PR #847, opened 2022-12-13
Added make release task

cpheps:
### Proposed Change
Added a `make release` task that first tags the main module and pushes in order to kick off the release. Then tags all submodules and pushes those tags.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,6 +204,32 @@ for-all:
 	 	$${CMD} ); \
 	done
 
+# Release a new version of the collector. This will also tag all submodules
+.PHONY: release
+release:
+	@if [ -z "$(version)" ]; then \
+		echo "version was not set"; \
+		exit 1; \
+	fi
+
+	@if ! [[ "$(version)" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$$ ]]; then \
+		echo "version $(version) is invalid semver"; \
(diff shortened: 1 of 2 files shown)


BinaryFissionGames (approved):

schmikei (approved):

Makefile
```diff
@@ -204,6 +204,34 @@ for-all:
...
+		exit 1; \
+	fi
+
+	@if ! [[ "$(version)" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$$ ]]; then \
+		echo "version $(version) is invalid semver"; \
+		exit 1; \
+	fi	
+
+	@git tag $(version)
+	@git push --tags
```
schmikei, line 221: Can we get by with only one `git push --tags`? It feels like we can, I just am not familiar enough with how GA would interpret that correctly/don't want to run one to validate if its worth it.

cpheps: So we have to do this twice for a reason. If we push all the tags at once the release action won't trigger (I tried this manually before). We have to push the single tag that triggers the action then all the rest.

schmikei: Ahh gotcha, that's what I assumed just was asking out of curiosity sake. Thanks for clarifying, approval still stands!

PR #869, opened 2023-01-05
deps: Removed dependency on opentelemetry-operations-collector

cpheps:
### Proposed Change
Removing the `github.com/GoogleCloudPlatform/opentelemetry-operations-collector` as a dependency to reduce external dependencies on the collector.

This removes the following components:
- normalizesums processor
- varnish receiver

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/config/google_managed_prometheus/hostmetrics/config.yaml
+++ b/config/google_managed_prometheus/hostmetrics/config.yaml
@@ -17,7 +17,6 @@ receivers:
       #process:
       #  mute_process_name_error: true
 
-
 processors:
   # Resourcedetection is used to add a unique (host.name)
   # to the metric resource(s), allowing users to filter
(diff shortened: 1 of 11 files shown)


StefanKurek (approved):
Not sure exactly why we're doing this, but the changes look good

cpheps:
> Not sure exactly why we're doing this, but the changes look good

@StefanKurek we're doing this to remove an external dependency from the collector. Right now updating OTel requires us to update `opentelemetry-operations-collector` with the latest OTel. We don't own that code base so if for some reason the owners push back against us updating we would have to fork the repo so we don't block our own release. The components that are removed as a result are not used by any current customer so the risk is low of customer impact.

If for some reason a customer needs Varnish we can look at alternatives to adding that back in.

schmikei (approved):

JonathanWamsley (approved):

PR #872, opened 2023-01-06
deps: Update OTel v0.68.0

cpheps:
### Proposed Change
Updated to OTel [v0.68.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.68.0)

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


exporter/googlecloudexporter/config_test.go
```diff
@@ -29,7 +28,6 @@ func TestCreateDefaultConfig(t *testing.T) {
 	googleCfg, ok := cfg.(*Config)
 	require.True(t, ok)
 
-	require.Equal(t, component.NewID(typeStr), googleCfg.ID())
```
StefanKurek, line 32: i believe you that this is fine, but can you explain why we don't need this test now?

BinaryFissionGames (approved):

StefanKurek (approved):

exporter/googlecloudexporter/config_test.go
cpheps: The ID is no longer on the config but rather on the settings.

PR #884, opened 2023-01-18
deps: Removed github.com/pkg/errors in sap receiver in favor of stdlib errors

cpheps:
### Proposed Change
I think this was missed in the original SAP PR but the receiver was using the `github.com/pkg/errors` package in place of the stdlib `errors` package. I switched it to using the stdlib package to reduce the dependencies of the receiver.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):

JonathanWamsley (approved):

schmikei (approved):
Ahh good catch

PR #905, opened 2023-02-09
feat: Add aggregation processor

BinaryFissionGames:
### Proposed Change
* Adds an aggregation processor, which can be used to aggregate metrics and emit those aggregates on a fixed interval.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


processor/aggregationprocessor/README.md
```diff
@@ -0,0 +1,141 @@
...
+This processor aggregates metrics over a configurable interval, allowing for metrics to be sampled at a higher rate, or to reduce the volume of metric data from push-based sources.
+
+## Minimum collector versions
+- Introduced: [v1.18.0](https://github.com/observIQ/observiq-otel-collector/releases/tag/v1.18.0)
+
+## Supported pipelines
+- Metrics
+
+## How it works
+1. The user configures the resource attribute transposer processor in the desired metrics pipeline.
```
cpheps, line 11: This says `resource attribute transposer processor`

processor/aggregationprocessor/README.md
```diff
@@ -0,0 +1,141 @@
...
+2. Every metric that flows through the pipeline is matched against the provided `include` regex.
+3. If the metric name does not match, the metric passes through the processor.
+4. If the metric is not a gauge or cumulative sum, the metric passes through the processor.
```
cpheps, line 14: Does this mean if it matches but it's not a gauge?

BinaryFissionGames: Updated the wording here to be more clear

processor/aggregationprocessor/README.md
```diff
@@ -0,0 +1,141 @@
...
+- type: max
+  metric_name: "$$0.max"
+- type: avg
+  metric_name: "$$0.avg"
+</pre></td><td>A list of aggregations to perform on each metric. By default, min, max, and average aggregate metrics are emitted every `interval`.</td></tr>
+
+<tr><td><pre>aggregations[].type</pre></td><td>aggregate type</td><td></td><td> The type of the aggregation. Valid values are: `min`, `max`, `avg`, `first`, `last`.</td></tr>
+
+<tr><td><pre>aggregations[].metric_name</pre></td><td>string</td><td><pre>"$$0"</pre></td><td>The name of the metric emitted for this aggregation. By default, the portion of the name matched by the `include` regex is used.</td></tr>
+</table>
```
BinaryFissionGames, line 46: I made this an html table because of the multi line default for the `aggregations` section; Don't know if we'd prefer this or if we'd prefer doing  something like the mask processor: https://github.com/observIQ/observiq-otel-collector/tree/main/processor/maskprocessor#configuration - I personally would like for the default here to be copy/paste-able

factories/processors.go
```diff
@@ -72,6 +73,7 @@ var defaultProcessors = []processor.Factory{
 	snapshotprocessor.NewFactory(),
 	spanmetricsprocessor.NewFactory(),
 	spanprocessor.NewFactory(),
+	aggregationprocessor.NewFactory(),
```
cpheps, line 76: nit: can we alphabetize this?

internal/processor/snapshotprocessor/processor.go
```diff
@@ -27,7 +27,7 @@ import (
 
 // getSnapshotReporter is function for retrieving the SnapshotReporter.
 // Meant to be overridden for tests.
-var getSnapshotReporter func() *report.SnapshotReporter = report.GetSnapshotReporter
+var getSnapshotReporter = report.GetSnapshotReporter
```
cpheps, line 30: Should this have been changed?

processor/aggregationprocessor/README.md
cpheps: Personally to me the maskprocessor is easier to read without having to know the html syntax.

processor/aggregationprocessor/config.go
```diff
@@ -0,0 +1,114 @@
...
+type Config struct {
+	Interval time.Duration `mapstructure:"interval"`
+	// Include is a regex that must match the metric name for it to be sampled.
+	// Otherwise, the metric is passed through.
+	Include string `mapstructure:"include"`
+	// List of aggregations for the metric
+	Aggregations []AggregateConfig `mapstructure:"aggregations"`
+}
+
+// AggregateConfig is a config that
```
cpheps, line 38: unfinished comment?

processor/aggregationprocessor/config.go
```diff
@@ -0,0 +1,114 @@
...
+	// MetricName is the name for the re-emitted metric. Defaults to `$0` (this is what is matched by the regex)
+	MetricName string `mapstructure:"metric_name"`
+}
+
+// Validate validate the config, returning an error explaining why it isn't if the config is invalid.
+func (a AggregateConfig) Validate() error {
+	var errs error
+
+	if !a.Type.Valid() {
+		errs = multierr.Append(errs, fmt.Errorf("invalid aggregate type for `type`: %s", a.Type))
```
cpheps, line 51: I don't think we need multierr here as there seems to be only one error.

processor/aggregationprocessor/config.go
```diff
@@ -0,0 +1,114 @@
...
+	if a.MetricName != "" {
+		return a.MetricName
+	}
+
+	return "$0"
+}
+
+// Validate validates the processor configuration
+func (cfg Config) Validate() error {
+	var errs error
```
cpheps, line 68: I think in the past we've preferred to just exit on the first failure instead of building them up with multierr. This does differ from how OTel does it. I'm not to picky on this but would like to try to be consistent within our codebase.

processor/aggregationprocessor/go.mod
```diff
@@ -0,0 +1,45 @@
+module github.com/observiq/observiq-otel-collector/processor/aggregationprocessor
+
+go 1.19
```
cpheps, line 3: Should be go 1.18

cpheps (changes requested):

processor/aggregationprocessor/config.go
BinaryFissionGames: I've removed the multierr here. I do think the multierr stuff is really valuable, the less times you have to reconfigure the collector the better, but it's unrelated to this PR.

processor/aggregationprocessor/README.md
BinaryFissionGames: It's definitely annoying from a dev POV. I'll try changing it over to the MD syntax, but the problem is that the default for `aggregations` will probably not be easily copy-pasteable, which probably sucks if you just want to tweak the default config for it from a user POV.

BinaryFissionGames: Actually, I think maybe I found a good middle-ground? Not sure how I feel about it, but it is copy-pasteable valid yaml in the MD format here.

https://github.com/observIQ/observiq-otel-collector/blob/eb9f373bae9780978bade70592bd3495964f292f/processor/aggregationprocessor/README.md

processor/aggregationprocessor/processor.go
```diff
@@ -0,0 +1,340 @@
...
+				dps := datapointsFromMetric(m)
+				// We remove datapoints that we aggregate here, so we use RemoveIf to iterate the datapoints
+				dps.RemoveIf(func(dp pmetric.NumberDataPoint) bool {
+					switch dp.ValueType() {
+					case pmetric.NumberDataPointValueTypeInt:
+					case pmetric.NumberDataPointValueTypeDouble:
+					default:
+						// ignore empty datapoints
+						return false
+					}
```
cpheps, line 126: Does this mean we remove if `dp.ValueType()` is `pmetric.NumberDataPointValueTypeInt` or `pmetric.NumberDataPointValueTypeDouble`? If so an if statement might be clearer. 

```go
if dp.ValueType() != pmetric.NumberDataPointValueTypeDouble && dp.ValueType() != pmetric.NumberDataPointValueTypeInt {
  return false
}

sp.aggregateDatapoint(ma, dp)
return true
```

processor/aggregationprocessor/processor.go
```diff
@@ -0,0 +1,340 @@
...
+		dp.SetTimestamp(now)
+	}
+}
+
+func (sp *aggregationProcessor) Capabilities() consumer.Capabilities {
+	// Data is mutate, since we remove Metric payloads if they are aggregated
+	return consumer.Capabilities{MutatesData: true}
+}
+
+func (sp *aggregationProcessor) Shutdown(_ context.Context) error {
```
cpheps, line 313: Should we be using this context as a timeout incase Wait hangs?

processor/aggregationprocessor/processor.go
```diff
@@ -0,0 +1,340 @@
...
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.uber.org/multierr"
+	"go.uber.org/zap"
+)
+
+type aggregationProcessor struct {
+	logger   *zap.Logger
+	mux      *sync.Mutex
+	wg       *sync.WaitGroup
```
cpheps, line 38: nit: These don't need to be pointers we aren't passing them around in functions so having references to them doesn't help us much.

processor/aggregationprocessor/README.md
cpheps: looks good to me

cpheps (approved):

Mrod1598 (approved):

PR #915, opened 2023-02-17
feat: Remove custom naming from aggregation processor

BinaryFissionGames:
### Proposed Change
~~* Move over to using our `expr` wrapper for calculating the metric name~~
  ~~* This allows for the full metric name to always be available when creating the metric name~~

* Remove naming for aggregations
  * Naming will be a fixed template of "${metric_name}.${aggregation_type}"
##### Checklist
- [x] Changes are tested
- [ ] CI has passed


processor/aggregationprocessor/config.go
```diff
@@ -48,16 +53,45 @@ func (a AggregateConfig) Validate() error {
...
+	}
+
 	return nil
 }
 
-// MetricNameString returns the configured name for the emitted metric, or "$0" if none was specified.
-func (a AggregateConfig) MetricNameString() string {
-	if a.MetricName != "" {
-		return a.MetricName
+const metricNameKey = "metric_name"
```
cpheps, line 64: nit: any reason this is middle of the file rather than the top? Just thinking consistency wise with most of our code.

BinaryFissionGames: I put it there since it's close to actual usage, but good point. Moved to top.

processor/aggregationprocessor/README.md
```diff
@@ -13,17 +13,17 @@ This processor aggregates metrics over a configurable interval, allowing for met
...
-| `include`                    | regex          | `"^.*$"` | A regex that specifies which metrics to consider for aggregation. The default regex matches all metrics.                         |
-| `aggregations`               | []map          | `[{type: min, metric_name: "$$0.min"}, {type: max, metric_name: "$$0.max"}, {type: avg, metric_name: "$$0.avg"}]`| A list of aggregations to perform on each metric.                                                                                |
+| `include`                    | regex          | `".*"` | A regex that specifies which metrics to consider for aggregation. The default regex matches all metrics.                         |
+| `aggregations`               | []map          | `[{type: min, metric_name_expression: 'metric_name + ".min"'}, {type: max, metric_name_expression: 'metric_name + ".max"'}, {type: avg, metric_name_expression: 'metric_name + ".avg"'}]`| A list of aggregations to perform on each metric.                                                                                |
```
cpheps, line 24: I missed the discussion on this but I feel like embedding a match expression is a bit complicated. I wasn't sure at first how I was suppose to configure this. I believe the intent is to give the user flexibility to name their aggregate metrics. I'm just wondering how many we really think will need that flexibility. I almost think if you really need that you can just chain a transform processor to rename the aggregate metrics and simplify configuration here where we just always do `metric_name.min`.

BinaryFissionGames: Good point; I've reworked to remove custom templates for metric names.

cpheps (approved):

Mrod1598 (approved):

PR #937, opened 2023-02-27
feat: Add flag to Linux Install to specify local package

### LINKED ISSUE #934: Bring install script into line with the one for BPOP
This is a two part issue to create parity among our products.

1. Add support for --file flag to install from a local rpm or deb file
2. Create file name parity for the install scripts for *NIX machines. BPOP is install_linux.sh and the collector is install_unix.sh, and they should really be the same name

The first part is extremely important for doing offline installs using the script in locked down environments

cpheps:
### Proposed Change
Resolves #934 

Added the `-f`/`--file` flag to the Linux installation script to allow for specifying a file path to a local package to install.

**Screen shot using the `--file` flag:**
Screenshot 2023-02-27 at 9 34 38 AM

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/install/install_unix.sh
+++ b/scripts/install/install_unix.sh
@@ -173,6 +173,9 @@ Usage:
       If not provided, this will default to observIQ Distro for OpenTelemetry Collector\'s GitHub releases.
       Example: '-l http://my.domain.org/observiq-otel-collector' will download from there.
 
+  $(fg_yellow '-f, --file')
+      Install Collector from a local file instead of downloading from a URL.
+
   $(fg_yellow '-x, --proxy')
       Defines the proxy server to be used for communication by the install script.
       Example: $(fg_blue -x) $(fg_magenta http\(s\)://server-ip:port/).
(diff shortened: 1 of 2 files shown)


scripts/install/install_unix.sh
```diff
@@ -347,13 +357,29 @@ set_os_arch()
 # Set the package type before install
 set_package_type()
 {
-  if command -v dpkg > /dev/null 2>&1; then
-      package_type="deb"
-  elif command -v rpm > /dev/null 2>&1; then
-      package_type="rpm"
+  # if package_path is set get the file extension otherwise looks what's avabile on the system
```
BinaryFissionGames, line 360:
```suggestion
  # if package_path is set get the file extension otherwise look at what's available on the system
```

BinaryFissionGames (approved):
Tested installing from local file, installing latest, installing an older version, and installing from a URL, works great!

antonblock (approved):

PR #975, opened 2023-03-13
feat: Added base_url flag to macOS and Linux installation scripts

cpheps:
### Proposed Change
Added a `--base_url` flag to the macOS and Linux installation scripts. This allows the script to look at a host other than GitHub for release while still being able to reason about the OS, architecture, and package type needed for the target system.

##### Checklist
- [X] Changes are tested
- [ ] CI has passed


antonblock (approved):
Uses the flag when specified and works in combination with `-v`

andykellr (approved):

PR #979, opened 2023-03-14
feat(plugin): W3C Plugin: Added most of the possible fields for filelog receiver and csv_parser operator as plugin parameters

Dylan-M:
### Proposed Change
This PR exposes most of the filelog receiver and csv_parser parameters as plugin parameters to give full functionality to the users
[cdn_httpaccess-testfile.log](https://github.com/observIQ/observiq-otel-collector/files/10972800/cdn_httpaccess-testfile.log)
[collector.log](https://github.com/observIQ/observiq-otel-collector/files/10972802/collector.log)

##### Checklist
- [X] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/w3c_logs.yaml
+++ b/plugins/w3c_logs.yaml
@@ -44,6 +44,10 @@ parameters:
   - name: timestamp_layout_type
     description: Optional timestamp layout type for parsing the timestamp, suggested if timestamp_layout is set
     type: string
+    supported:
+      - strptime
+      - gotime
+      - epoch
   - name: severity
     description: Optional severity block which will parse a severity field
     type: string
--- a/plugins/w3c_logs.yaml
+++ b/plugins/w3c_logs.yaml
@@ -48,9 +48,6 @@ parameters:
       - strptime
       - gotime
       - epoch
-  - name: severity
-    description: Optional severity block which will parse a severity field
-    type: string
   - name: parse_from
     description: Where to parse the data from
     type: string
--- a/plugins/w3c_logs.yaml
+++ b/plugins/w3c_logs.yaml
@@ -1,4 +1,4 @@
-version: 0.1.0
+version: 0.2.0
 title: W3C
 description: Log Parser for W3C
 parameters:
--- a/plugins/w3c_logs.yaml
+++ b/plugins/w3c_logs.yaml
@@ -153,6 +153,13 @@ template: |
           lazy_quotes: {{ .lazy_quotes }}
           ignore_quotes: {{ .ignore_quotes }}
           delimiter: '{{ .delimiter }}'
+          {{ if .timestamp_layout }}
+          timestamp:
+            parse_from: {{ .timestamp_parse_from }}
+            layout: '{{ .timestamp_layout }}'
+            layout_type: {{ .timestamp_layout_type }}
+            location: {{ .timestamp_location }}
+          {{ end }}
 
           {{ if .header_delimiter }} 
           header_delimiter: '{{ .header_delimiter }}'
@@ -166,17 +173,6 @@ template: |
         {{ if not .header }}
         - type: remove
           field: "attributes.header_fields"
-        {{ end }}
-        {{ if .timestamp_layout }}
-        timestamp:
-          parse_from: {{ .timestamp_parse_from }}
-          layout: {{ .timestamp_layout }}
-          {{ if .timestamp_layout_type }}
-          layout_type: .timestamp_layout_type
-          {{ end }}
-          {{ if .timestamp_location }}
-          location: .timestamp_location
-          {{ end }}
         {{ end }}
           
   service:
--- a/plugins/w3c_logs.yaml
+++ b/plugins/w3c_logs.yaml
@@ -25,7 +25,7 @@ parameters:
     type: string
     default: w3c
   - name: start_at
-    description: At startup, where to start reading logs from the file. Must be set to "beginning" if 'header' is not specified.
+    description: At startup, where to start reading logs from the file. Must be set to "beginning" if 'header' is not specified or if 'delete_after_read' is being used.
     type: string
     supported:
       - beginning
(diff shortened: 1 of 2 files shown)


plugins/w3c_logs.yaml
```diff
@@ -35,6 +35,41 @@ parameters:
...
     default: 512
+  - name: timestamp_layout
+    description: Optional timestamp layout which will parse a timestamp field
+    type: string
+  - name: timestamp_parse_from
+    description: Field to parse the timestamp from, required if timestamp_layout is set
+    type: string
+  - name: timestamp_layout_type
+    description: Optional timestamp layout type for parsing the timestamp, suggested if timestamp_layout is set
+    type: string
```
andykellr, line 46: What are the possible values for this?

Dylan-M: Thanks Andy, I'll get the three values enumerated.

plugins/w3c_logs.yaml
```diff
@@ -74,6 +110,7 @@ template: |
       storage: file_storage
       start_at: '{{ .start_at }}'
       max_concurrent_files: {{ .max_concurrent_files }}
+      delete_after_read: {{ .delete_after_read }}
```
cpheps, line 113: Delete after read only works if `start_at` is at beginning. We should add that to the description and also add a conditional here.

cpheps (changes requested):
Can you please provide a snippet in the PR description of some logs parsed with these new fields in use? We do this on all plugin PRs to ensure they are working as expected.

Also, can you minor bump the version number and run `make create-plugin-docs` to regenerate the plugin documentation.

cpheps (commented):
LGTM can you provide an example from the logging exporter with the new parameters in use?

Dylan-M:
> LGTM can you provide an example from the logging exporter with the new parameters in use?

Not all of them, since some are conflicting. I assume that is okay?

cpheps (approved):

plugins/w3c_logs.yaml
```diff
@@ -35,6 +35,47 @@ parameters:
...
+      - body
+    default: body
+  - name: lazy_quotes
+    description: If true, a quote may appear in an unquoted field and a non-doubled quote may appear in a quoted field. Cannot be true if 'ignore_quotes' is true.
+    type: bool
+    default: false
+  - name: ignore_quotes
+    description: If true, all quotes are ignored, and fields are simply split on the delimiter. Cannot be true if 'lazy_quotes' is true.
+    type: bool
+    default: false
```
BinaryFissionGames, line 74: W3C logs are ALWAYS whitespace separated. The spec for W3C doesn't mention quoting, they are simply "whitespace separated", so we should ALWAYS ignore quotes if they are valid W3C logs. Have we actually run into a situation where we would want to respect quotes for W3C logs in the field?

(if we really need this functionality, I suggest having `ignore_quotes` true by default, so we follow the spec by default)

plugins/w3c_logs.yaml
```diff
@@ -35,6 +35,47 @@ parameters:
...
+      - epoch
+    default: strptime
+  - name: timestamp_location
+    description: The geographic location (timezone) to use when parsing a timestamp that does not include a timezone. The available locations depend on the local IANA Time Zone database.
+    type: string
+    default: UTC
+  - name: parse_from
+    description: Where to parse the data from
+    type: string
+    default: body
```
BinaryFissionGames, line 59: The data will always be on the body; Would a customer ever need to change this?

plugins/w3c_logs.yaml
```diff
@@ -35,6 +35,47 @@ parameters:
     description: Max number of W3C files that will be open during a polling cycle
     type: int
     default: 512
+  - name: timestamp_layout
+    description: Optional timestamp layout which will parse a timestamp field
+    type: string
+  - name: timestamp_parse_from
+    description: Field to parse the timestamp from, required if 'timestamp_layout' is set
+    type: string
```
BinaryFissionGames, line 43: The W3C spec is actually pretty explicit about how the timestamp should be formatted, and what fields it should be on (it's actually split between two fields, `date` and `time`). 

It's defined towards the bottom of this page: https://www.w3.org/TR/WD-logfile.html - Should we just be parsing this as defined by the spec automatically, like we do in IIS? 

I don't think these parsing options will actually work with actual W3C logs, where, again, the date and time are on separate fields that need to be joined to be properly parsed.

plugins/w3c_logs.yaml
```diff
@@ -104,10 +146,21 @@ template: |
       operators:
         - type: filter
           expr: 'body matches "^#"'
-
         - type: csv_parser
           ignore_quotes: true
+          parse_from: '{{ .parse_from }}'
+          parse_to: {{ .parse_to }}
+          lazy_quotes: {{ .lazy_quotes }}
+          ignore_quotes: {{ .ignore_quotes }}
```
BinaryFissionGames, line 154: looks like `ignore_quotes` is defined twice here.

BinaryFissionGames (changes requested):

plugins/w3c_logs.yaml
Dylan-M: While we did see, in a single log file, a non-conforming entry, I've followed your suggestion. I think I did it the right way to make this default and have it still work the other way. Pushing shortly.

plugins/w3c_logs.yaml
Dylan-M: Per your reference to the spec, I've removed the quotes options.

plugins/w3c_logs.yaml
Dylan-M: Per your reference to the spec, I've removed the quotes options.

Dylan-M:
Addressed @BinaryFissionGames' feedback, updated my test log file and results in the description.

BinaryFissionGames (approved):

jsirianni (approved):

PR #980, opened 2023-03-14
feat(opamp): Add environment variable options for TLS

jsirianni:
### Proposed Change

This change adds support for the following OpAMP options using environment variables
- TLS skip verify (skip verifying the server's certificate)
- CA certificate (can be used as an alternative to the underlying operating systems trust store)
- Mutual TLS
  - Private Key
  - Certificate

This solves a container use-case, where the collector's configuration is derived only from environment variables.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/cmd/collector/main.go
+++ b/cmd/collector/main.go
@@ -21,6 +21,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 	_ "time/tzdata"
 
 	"github.com/observiq/observiq-otel-collector/collector"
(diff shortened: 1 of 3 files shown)


cmd/collector/main.go
```diff
@@ -262,3 +276,37 @@ func checkForCollectorRollbackConfig(configPath string) error {
...
+
+	if key := os.Getenv(tlsKeyENV); key != "" {
+		tlsConfig.KeyFile = &key
+		tlsConfigured = true
+	}
+
+	if tlsConfigured {
+		return &tlsConfig, nil
+	}
+	return nil, nil
```
cpheps, line 311: I'm wondering what happens if we give opamp and empty TLSConfig? Would it correctly not do TLS? If so it would be nice to just `return tlsConfig, nil` always.

cpheps: Ah ok I forgot it was being saved back. I can't think of a cleaner way to do this so looks good.

cpheps (approved):

cpheps (approved):

BinaryFissionGames (approved):

PR #994, opened 2023-03-23
fix: Fixed bug with default rules in the mask processor

jmwilliams89:
### Proposed Change
- Changed the behavior of default rules in the mask processor.
  - Default rules are now set at "runtime" rather than "config time"
  - The previous design was flawed, because otel uses mapstructure to unmarshal configs. When mapstructure unmarshals into a map that already has values, it doesn't wipe that map clean with the incoming map. Instead, it will merge the source and destination. In this scenario, only duplicate keys would be overwritten. This means that if a user configured their own rules in the previous design, the processor would load the union of the two rulesets. This should be considered a bug, since defaults should only apply if a user leaves this field blank. This can be accomplished by moving where the default is set and avoiding mapstructure altogether.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/processor/maskprocessor/processor.go
+++ b/processor/maskprocessor/processor.go
@@ -52,7 +52,7 @@ func newProcessor(logger *zap.Logger, cfg *Config) *maskProcessor {
 
 // start is used to start the processor.
 func (p *maskProcessor) start(context.Context, component.Host) error {
-	rules, err := p.cfg.CompileRules()
+	rules, err := p.createRules()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


processor/maskprocessor/processor.go
```diff
@@ -211,3 +211,27 @@ func (p *maskProcessor) createMaskFunc(field string) func(k string, v pcommon.Va
 		return true
 	}
 }
+
+// createRules creates a map of rules for the processor.
+func (p *maskProcessor) createRules() (map[string]*regexp.Regexp, error) {
+	if len(p.cfg.Rules) == 0 {
+		return compileRules(defaultRules)
```
cpheps, line 218: Can we store the default rules as a 'map[string]*regexp`? Using must compile to seed it? They're default so we shouldn't take any chances of them but working.

BinaryFissionGames (approved):

cpheps (approved):

PR #995, opened 2023-03-24
fix: W3C plugin not working when timestamp_parse_from is not configured

cpheps:
### Proposed Change
- Fixed a bug in the w3c plugin where `timestamp_parse_from` is not specified. It's not required  but is included in the plugin template.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/w3c_logs.yaml
+++ b/plugins/w3c_logs.yaml
@@ -137,8 +137,8 @@ template: |
           expr: 'body matches "^#"'
         - type: csv_parser
           ignore_quotes: true
+          parse_to: {{ .parse_to }}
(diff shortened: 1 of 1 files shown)


BinaryFissionGames:
Tested this locally, it looks like by default, the timestamp parsing is not working, because the csv_parser step doesn't respect the `parse_to` option.

Needs: `parse_to: {{ .parse_to }}` on the CSV operator to work properly.

Here's a sample log:
[w3c_format_logs.log](https://github.com/observIQ/observiq-otel-collector/files/11063069/w3c_format_logs.log)

BinaryFissionGames (approved):

PR #1014, opened 2023-04-05
feat(expr): Update with the ability to convert spans and datapoints to a map representation

BinaryFissionGames:
### Proposed Change
* Add ConvertToDatapointResourceGroup and ConvertToSpanResourceGroups functions that convert pdata datapoints and spans into a flattened representation

This change allows us to use these telemetry types in `expr` expressions, which will be necessary to implement a count processor for these two telemetry types.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/expr/span.go
+++ b/expr/span.go
@@ -1,0 +1,68 @@
+package expr
+
+import (
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+const (
+	SpanKindField          = "trace_kind"
+	SpanStatusCodeField    = "trace_status_code"
+	SpanStatusMessageField = "trace_status_message"
+	SpanDurationField      = "span_duration_ms"
+)
+
+var spanKindToString = map[ptrace.SpanKind]string{
+	ptrace.SpanKindUnspecified: "unspecified",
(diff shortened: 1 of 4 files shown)


expr/span.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+// Span specific fields for use in expressions
+const (
+	SpanKindField          = "trace_kind"
+	SpanStatusCodeField    = "trace_status_code"
+	SpanStatusMessageField = "trace_status_message"
+	SpanDurationField      = "span_duration_ms"
```
cpheps, line 26: Do these not cause linter errors? I would expect so since they're exported.

cpheps (approved):

expr/span.go
BinaryFissionGames: One comment on the group is satisfies the linter here. I think the one comment up there is sufficient enough, eventually when the actual processors are implemented, these will be documented there.

Mrod1598 (approved):

PR #1016, opened 2023-04-06
feat: Add datapointcount processor

BinaryFissionGames:
### Proposed Change
* Adds a processor, similar to the logcount processor, that counts the number of datapoints passing through it.

Test config:
```yaml
receivers:
    hostmetrics:
        collection_interval: 30s
        scrapers:
            load:
            filesystem:
            memory:
            network:
    route/example:
processors:
    batch:
    datapointcount:
        route: example
        attributes:
            metric: metric_name
exporters:
    nop:
    logging:
      verbosity: detailed

service:
    telemetry:
      metrics:
        level: none
    pipelines:
        metrics/host:
            receivers: [hostmetrics]
            processors: [datapointcount, batch]
            exporters: [nop]
        metrics/count:
            receivers: [route/example]
            processors: [batch]
            exporters: [logging]
```

After one minute, this will export counts of datapoints for each metric to the logging exporter

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Mrod1598 (approved):

processor/datapointcountprocessor/README.md
```diff
@@ -0,0 +1,87 @@
...
+## How It Works
+1. The user configures the datapoint count processor in their metrics pipeline and a route receiver in their target metrics pipeline.
+2. If any incoming metrics match the `match` expression, they are counted and dimensioned by their `attributes`. Regardless of match, all metrics are sent to the next component in the metrics pipeline.
+3. After each configured interval, the observed metric counts are converted into gauge metrics. These metrics are sent to the configured route receiver.
+
+
+## Configuration
+| Field        | Type     | Default | Description |
+| ---          | ---      | ---     | ---         |
+| match        | string   | `true`  | A boolean [expression](https://github.com/antonmedv/expr/blob/master/docs/Language-Definition.md) used to match which datapoints to count. By default, all datapoints are counted. |
```
cpheps, line 16: Should we link to the actual version of the expression language we're using? Not sure how much it changes.

cpheps (approved):

processor/datapointcountprocessor/README.md
BinaryFissionGames: I do think the "right thing" to do is to hit the actual version, but I'm worried that will end up just being out of sync with the go.mod version very quickly. Maybe we should discuss this with the team and follow up on this in a separate PR, if we decide to link the specific version.

I DO notice we actually reference master basically everywhere, including in BPOP where this is used.

The `expr` language was actually static for a while, until the turn of this year, the maintainer has started adding a couple more features.

PR #1025, opened 2023-04-13
feat(plugin): Pinecone

BominRahmani:
Waiting on bug fix from Pinecone before fully setting up the PR for this. Currently going to stay as Draft.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ opentelemetry-java-contrib-jmx-metrics.jar
 VERSION.txt
 release_deps
 /tmp
+tmp
 
 # OpAmp Files
 collector*.yaml
--- a/plugins/pinecone.yaml
+++ b/plugins/pinecone.yaml
@@ -56,15 +56,15 @@ template: |
         - include: pinecone_request_count_total
           match_type: strict
           action: update
-          new_name: "pinecone.request.count.total"
+          new_name: "pinecone.request.count"
         - include: pinecone_request_error_count_total
           match_type: strict
           action: update
-          new_name: "pinecone.request.error.count.total"
+          new_name: "pinecone.request.error.count"
         - include: pinecone_request_latency_seconds
           match_type: strict
           action: update
-          new_name: "pinecone.request.latency.seconds"
+          new_name: "pinecone.request.latency"
         - include: pinecone_index_fullness
           match_type: strict
           action: update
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -21,6 +21,7 @@ opentelemetry-java-contrib-jmx-metrics.jar
 VERSION.txt
 release_deps
 /tmp
+tmp
```
cpheps, line 24: We already ignore temp can you condense this line and the one above with `**/tmp` I think that should ignore either path.

plugins/pinecone.yaml
```diff
@@ -0,0 +1,103 @@
...
+template: |
+  receivers:
+    prometheus/pinecone:
+      config:
+        scrape_configs:
+          - job_name: 'pinecone'
+            scrape_interval: '{{ .scrape_interval }}'
+            authorization:
+              credentials: '{{ .api_key }}'
+            scheme: https
```
cpheps, line 26: we're forcing TLS here will it always work or is there a locally hosted version of pinecone where they don't require TLS?

plugins/pinecone.yaml
```diff
@@ -0,0 +1,103 @@
+version: 0.0.1
+title: Pinecone Metrics
+description: Metrics receiver for Pinecone
+parameters:
+  - name: environment
+    description: Environment of your project  (example us-east4-gcp)
```
cpheps, line 6: Is this a gcp zone or is it broader than that? Is this easy for users to find?

cpheps (changes requested):
Make sure to run `make create-plugin-docs`

plugins/pinecone.yaml
```diff
@@ -0,0 +1,103 @@
...
+      - context: metric
+        statements:
+        - set(description, "Number of items per pod in the index.") where name == "pinecone.vector.count"
+        - set(unit, "{items}") where name == "pinecone.vector.count"
+
+        - set(description, "Number of data plane calls made by clients.") where name == "pinecone.request.count"
+        - set(unit, "{requests}") where name == "pinecone.request.count"
+
+        - set(description, "Number of data plane calls made by clients that resulted in errors.") where name == "pinecone.request.error.count"
+        - set(unit, "requests") where name == "pinecone.request.error.count"
```
BinaryFissionGames, line 91:
```suggestion
        - set(unit, "{requests}") where name == "pinecone.request.error.count"
```

plugins/pinecone.yaml
```diff
@@ -0,0 +1,103 @@
...
+
+        - set(description, "Distribution of server-side processing latency for pinecone data plane calls.") where name == "pinecone.request.latency"
+        - set(unit, 1) where name == "pinecone.request.latency"
```
BinaryFissionGames, line 94: Latency is in seconds here, it looks like?
```suggestion
        - set(unit, "s") where name == "pinecone.request.latency"
```

BinaryFissionGames (commented):
Could we change the filename from `pinecone.yaml` to `pinecone_metrics.yaml`?

plugins/pinecone.yaml
jsirianni: I could be wrong, but I believe Pinecone is a saas, which would always have TLS.

jsirianni (commented):
Once this is working, I would be curious to see the metric Resource Attributes. We want to be sure they are unique. For example, the user configures two pinecone plugin receivers to collect metrics from two different accounts. Will the resulting metrics be duplicates or will their resource attributes guarantee uniqueness?

cpheps:
Closing this due to staleness we can reopen if we need it.

PR #1031, opened 2023-04-18
feat: Add datadog exporter

jmwilliams89:
### Proposed Change
- Adds the Datadog exporter as an available option
- Updates exporter docs to include Datadog

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cpheps (approved):

go.mod
```diff
@@ -413,6 +439,7 @@ require (
 	github.com/nginxinc/nginx-prometheus-exporter v0.8.1-0.20201110005315-f5a5f8086c19 // indirect
 	github.com/nwaples/rardecode v1.1.0 // indirect
 	github.com/observiq/ctimefmt v1.0.0 // indirect
+	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/datadogexporter v0.75.0
```
BinaryFissionGames, line 442: Can we move this up to the top import block? We should keep the block for just the indirect dependencies.

BinaryFissionGames (changes requested):

go.mod
jmwilliams89: Interesting. That's just where `go get` placed it on its own. I can move it though.

cpheps: Yeah, it's not a big deal but it does make it easier when updating otel when they're all in one spot.

BinaryFissionGames (approved):

PR #1036, opened 2023-04-20
feat(plugin): Add Severity parsing to Kubernetes container logs

jsirianni:
### Proposed Change

The container logs plugin does not contain severity parsing. In an effort to fall in line with [Google's GKE log agent](https://cloud.google.com/blog/products/management-tools/using-logging-your-apps-running-kubernetes-engine), I have added the following:
- parse `attributes.stream` for basic severity
- parse `body.severity` if present, using default mappings
- parse `body.level` if `severity` is not present, using default mappings

This PR handles common severity cases. Support for more advanced options could come in the future, such as custom severity field / custom mapping, etc.

Before:
Screenshot from 2023-04-20 16-07-59

After:
- steam: stdout (info log)
- stream: stderr (error log)
- body.severity (debug log)
steam-info
stream_error
body_debug

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/plugins/container_logs.yaml
+++ b/plugins/container_logs.yaml
@@ -1,4 +1,4 @@
-version: 0.4.1
+version: 0.5.0
 title: Kubernetes Container Logs
 description: Log parser for Kubernetes Container logs. This plugin is meant to be used with the OpenTelemetry Operator for Kubernetes (https://github.com/open-telemetry/opentelemetry-operator).
 parameters:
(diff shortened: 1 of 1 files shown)


plugins/container_logs.yaml
```diff
@@ -138,16 +138,43 @@ template: |
...
+
+          # After parsing the log body as json, parse severity
+          # or level if present, with severity taking precedence.
+          - type: router
+            default: file_name_parser
+            routes:
+              - expr: 'body.severity != nil'
+                output: body_severity_parser
+              - expr: 'body.level != nil'
+                output: body_level_parser
```
cpheps, line 164: Are these mutally exclusive?

cpheps (approved):

plugins/container_logs.yaml
jsirianni: I think so. Im not sure I have observed logs which have both fields. Google looks for `severity`. I added `level` because it is common.

There is probably a lot of room for improvement here, but picking these two fields was low hanging fruit.

BinaryFissionGames (approved):

PR #1040, opened 2023-04-26
fix(plugin): Container logs cri-o regex, allow empty message

jsirianni:
### Proposed Change

Updated the `cri-o` regex parser to handle logs which do not have a message. This is a rare occurrence but I do see it every once and a while.

The regex parser will fail and cause dependent operators to fail (three errors per failure).
```
entry does not contain the combine_field
```
```
remove: field does not exist
```
```
regex pattern does not match
```

The regex has been updated to allow for a missing space after the partial field and missing message field.

See https://regex101.com/r/Mk8fY5/1

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/plugins/container_logs.yaml
+++ b/plugins/container_logs.yaml
@@ -101,7 +101,7 @@ template: |
           # 2022-06-18T16:52:59.639114537Z stdout F {"message":"registered Stackdriver tracing","severity":"info","timestamp":"2022-06-18T16:52:59.639034532Z"}
           - id: containerd_cri_parser
             type: regex_parser
-            regex: '^(?P<time>[^\s]+) (?P<stream>\w+) (?P<partial>\w)?\s+?(?P<log>.*)'
+            regex: '^(?P<time>[^\s]+) (?P<stream>\w+) (?P<partial>\w)? ?(?P<log>.*)'
           - type: recombine
             source_identifier: attributes["log.file.name"]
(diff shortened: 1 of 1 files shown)


dsvanlani (approved):

plugins/container_logs.yaml
```diff
@@ -101,7 +101,7 @@ template: |
           # 2022-06-18T16:52:59.639114537Z stdout F {"message":"registered Stackdriver tracing","severity":"info","timestamp":"2022-06-18T16:52:59.639034532Z"}
           - id: containerd_cri_parser
             type: regex_parser
-            regex: '^(?P<time>[^\s]+) (?P<stream>\w+) (?P<partial>\w)?\s+?(?P<log>.*)'
+            regex: '^(?P<time>[^\s]+) (?P<stream>\w+) (?P<partial>\w)? ?(?P<log>.*)'
```
cpheps, line 104: Instead of a space could we just replace the `+` with a `*` to indicate zero or more? Or are we certain it's exactly one space always?

cpheps (approved):

cpheps (commented):
Can we bump the version of the plugin please?

plugins/container_logs.yaml
jsirianni: It is always one space but your solution seems more appropriate.

PR #1041, opened 2023-04-27
`feat(M365receiver):` 

dpaasman00:
### Proposed Change
otel receiver for Microsoft Office 365.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dehaansa (commented):
Overall good bones to the receiver, but had a few comments and a couple refactor thoughts.

receiver/m365receiver/client.go
```diff
@@ -0,0 +1,174 @@
...
+		var respErr csvError
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return []string{}, err
+		}
+		err = json.Unmarshal(body, &respErr)
+		if err != nil {
+			return []string{}, err
+		}
+		if respErr.ErrorCSV.Code == "InvalidAuthenticationToken" {
```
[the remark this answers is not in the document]
dehaansa: I saw later in the review that the stale token logic is a step above in the scraper code 👍

receiver/m365receiver/client.go
```diff
@@ -0,0 +1,174 @@
...
+
+	defer func() { _ = resp.Body.Close() }()
+	csvReader := csv.NewReader(resp.Body)
+
+	//parse out 2nd line & return csv data
+	_, err = csvReader.Read()
+	if err != nil {
+		return []string{}, err
+	}
+	data, err := csvReader.Read()
```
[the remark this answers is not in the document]
dpaasman00: ya, only need the first line of data which is the 2nd line in the csv

receiver/m365receiver/metadata.yaml
```diff
@@ -0,0 +1,205 @@
...
+    description: The number of active sites across the Sharepoint.
+    unit: "{files}"
+    sum:
+      value_type: int
+      monotonic: false
+      aggregation: cumulative
+    attributes: []
+  m365.sharepoint.pages.viewed.count:
+    enabled: true
+    description: The number of unique pages viewed across all sites.
```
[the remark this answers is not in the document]
dpaasman00: Ya I can go through and revamp the descriptions to make it more clear

receiver/m365receiver/metadata.yaml
```diff
@@ -0,0 +1,205 @@
+type: m365receiver
+
+status:
+  stability: development
+  pipelines: [metrics]
+
+attributes:
+  teamsDevices:
+    name_override: kind
```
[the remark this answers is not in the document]
dpaasman00: sounds good

receiver/m365receiver/README.md
```diff
@@ -0,0 +1 @@
+receiver for microsoft 365
```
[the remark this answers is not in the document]
dpaasman00: Ya definitely

.gitignore
```diff
@@ -26,3 +26,9 @@ release_deps
 collector*.yaml
 manager*.yaml
 logging*.yaml
+
+#m365recv dev files
+m365metricsconfig.yaml
```
dehaansa, line 31: If these just live in your local or tmp directory, there's no need to add them specifically to the gitignore.

dpaasman00: gotcha

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,537 @@
...
+	if l.record.NextStartTime != nil {
+		st = l.record.NextStartTime.Format(layout)
+	} else {
+		st = pcommon.NewTimestampFromTime(time.Now().Add(-l.pollInterval)).AsTime().UTC().Format(layout)
+	}
+
+	now := time.Now().UTC()
+
+	auditWG := &sync.WaitGroup{}
+	auditWG.Add(5)
```
dehaansa, line 178: especially in case of future additions of more audit logs, this should be `len(l.audits)`

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,537 @@
...
+
+	l.mu.Lock()
+	data, err := l.client.GetJSON(ctx, l.root+audit.route, now.Format(layout), st)
+	if err != nil {
+		if err.Error() == "authorization denied" { // troubleshoot stale token
+			l.logger.Debug("possible stale token; attempting to regenerate")
+			err = l.client.GetToken()
+			if err != nil { // something went wrong generating token
+				l.logger.Error("error creating authorization token", zap.Error(err))
+				l.mu.Unlock()
```
dehaansa, line 203: more common pattern is to `defer l.mu.Unlock()` right when you call .Lock(), though in the current form it will hold the lock through the transform & consume code. Tiny refactor to put GetJSON & error handling into a separate function would fix that and be cleaner code.

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,537 @@
...
+			l.logger.Error(audit.route) //debug only
+			l.mu.Unlock()
+			return
+		}
+	}
+	l.mu.Unlock()
+
+	logs := l.transformLogs(pcommon.NewTimestampFromTime(now), audit, data)
+
+	l.consumerMu.Lock()
```
dehaansa, line 223: Should just immediately call defer Unlock()

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,537 @@
...
+func (l *m365LogsReceiver) transformLogs(now pcommon.Timestamp, audit *auditMetaData, data logData) plog.Logs {
+	logs := plog.NewLogs()
+	resourceLogs := logs.ResourceLogs().AppendEmpty()
+	ra := resourceLogs.Resource().Attributes()
+	ra.PutStr("m365.audit", audit.name)
+	ra.PutStr("m365.organization_id", l.cfg.TenantID)
+
+	for i := 0; i < len(data.logs); i++ {
+		log := data.logs[i]
+		logRecord := resourceLogs.ScopeLogs().AppendEmpty().LogRecords().AppendEmpty()
```
dehaansa, line 242: The set of logs should belong to the same ScopeLogsSlice, so resourcelogs.ScopeLogs().AppentEmpty() should be called & captured into a variable outside of the loop.

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,537 @@
...
+
+		// log body
+		logRecord.Body().SetStr("\"C" + data.body[i])
```
dehaansa, line 245: Could use a comment explaining the `\"C`

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,537 @@
...
+
+		// timestamp
+		ts, err := time.Parse(layout, log.CreationTime)
+		if err != nil {
+			l.logger.Warn("unable to interpret when an event was created, expecting a RFC3339 timestamp", zap.String("timestamp", log.CreationTime), zap.String("log", log.ID))
```
dehaansa, line 250: Is there a reason we're not using time.RFC3339 as the layout?

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,537 @@
...
+	}
+
+	if bytes == nil {
+		l.record = &logRecord{}
+		return
+	}
+
+	var record logRecord
+	if err = json.Unmarshal(bytes, &record); err != nil {
+		l.logger.Error("unable to decode stored record for events, continuing without a checkpoint", zap.Error(err))
```
dehaansa, line 295: stored record for logs

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,537 @@
...
+	}
+	if log.SharepointSourceFileName != "" {
+		m.PutStr("sharepoint.source.file.name", log.SharepointSourceFileName)
+	}
+	if log.ExchangeMailboxGUID != "" {
+		m.PutStr("exchange.mailbox.id", log.ExchangeMailboxGUID)
+	}
+	if log.AzureActor != nil {
+		slice := m.PutEmptySlice("azure.actors")
+		aMap := slice.AppendEmpty().SetEmptyMap()
```
dehaansa, line 322: We should be appending the empty map & setting values inside the for loop, this is currently just modifying the same map for every actor.

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,537 @@
...
+		l.record = &logRecord{}
+		return
+	}
+	l.record = &record
+}
+
+// adds any attributes to log that may or may not be present
+func parseOptionalAttributes(m *pcommon.Map, log *jsonLogs) {
+	recordType := matchRecordType(log.RecordType)
+	if log.Workload != "" {
```
dehaansa, line 305: Repeated code setting simple string optional attributes can be refactored to a function, something like

```
func setAttributeIfNotEmpty(m *pcommon.Map, field, value string) {
    if field != "" {
        m.PutStr(field, value)
    }
}
```

receiver/m365receiver/logs.go
dehaansa: Alternatively, just call wg.Add(1) in the loop, for the same reason.

receiver/m365receiver/logs.go
dpaasman00: the timestamps returned by the api don't include the time zone offsets that are specified in time.RFC3339, so the time.Parse call was always returning a non nil error and setting the timestamp to time.Now instead of log.CreationTime, so I made a custom layout to better match CreationTime

receiver/m365receiver/client.go
```diff
@@ -0,0 +1,424 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	// troubleshoot error handling, mainly sub already enabled
+	// no error if sub already enabled, not troubleshooting stale token
+	// only called while code is synchronous right after a GetToken call
+	// if token is stale, regenerating it won't fix anything
+	if resp.StatusCode != 200 {
```
dehaansa, line 359: Could refactor error handling out of the three functions to clean up client a bit.

receiver/m365receiver/client.go
```diff
@@ -0,0 +1,424 @@
...
+			return []string{}, nil
+		}
+		return []string{}, err
+	}
+
+	return data, nil
+}
+
+// function for getting log data
+func (m *m365Client) GetJSON(ctx context.Context, endpoint string, end string, start string) (logData, error) {
```
dehaansa, line 282: Just noticed that only this function accepts context as a parameter, they probably all should & request with context.

receiver/m365receiver/client.go
```diff
@@ -0,0 +1,460 @@
...
+		Content string `json:"contentUri"`
+	}{}
+	if err = json.Unmarshal(body, &lr); err != nil {
+		return "", err
+	}
+	return lr[0].Content, nil
+}
+
+// handles requests for client functions
+func (m *m365Client) makeRequest(ctx context.Context, r ReqType, endpoint string, opts ...Option) (*http.Response, error) {
```
dehaansa, line 419: I think this function should be more generic, and contain less knowledge about the caller(s).

Something like the following, with a few new options for setting auth or providing a body. The body would then be turned into a strings.NewReader outside of the scope of the function where appropriate, and the callers would set the appropriate http method.

```
func (m *m365Client) makeRequest(ctx context.Context, httpMethod, endpoint string,  opts ...Option) (*http.Response, error) {

    req, err := http.NewRequestWithContext(ctx, "POST", endpoint, body)
    if err != nil {
        return nil, err
    }
    for _, o := range opts {
        o(req)
    }
    
    return m.client.Do(req)
}

func WithToken(token string) Option {
	return func(r *http.Request) {
		r.Header.Set("Authorization", "Bearer "+m.token)
	}
}

func WithBody(body io.ReadCloser) Option {
	return func(r *http.Request) {
		r.Body = body
		r.Header.Set("Content-Type", "application/x-www-form-urlencoded")
	}
}
```

receiver/m365receiver/factory.go
```diff
@@ -0,0 +1,97 @@
...
+		createDefaultConfig,
+		receiver.WithMetrics(createMetricsReceiver, metadata.Stability),
+		receiver.WithLogs(createLogsReceiver, metadata.Stability),
+	)
+}
+
+func createDefaultConfig() component.Config {
+	return &Config{
+		ScraperControllerSettings: scraperhelper.ScraperControllerSettings{
+			CollectionInterval: 5 * time.Second,
```
dehaansa, line 47: Make sure this is set to a significantly longer collection interval before merging.

dehaansa: the README should also talk about the frequency of metrics being available (one data point per day) as well.

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,514 @@
...
+	now := time.Now().UTC()
+
+	auditWG := &sync.WaitGroup{}
+	for i := 0; i < len(l.audits); i++ {
+		auditWG.Add(1)
+		go l.poll(ctx, now, st, &l.audits[i], auditWG)
+	}
+	auditWG.Wait()
+
+	l.record.NextStartTime = &now
```
dehaansa, line 184: Since you said the logs are not always immediately available in the API, we should only update NextStartTime to be equal to the latest timestamp of log we did collect, so that we will keep requesting from the last log collected, allowing us to hopefully avoid lost data.

dehaansa: Discussed offline, this shouldn't be necessary based on the behavior of the API.

dehaansa (approved):

receiver/m365receiver/factory.go
```diff
@@ -0,0 +1,97 @@
...
+		createDefaultConfig,
+		receiver.WithMetrics(createMetricsReceiver, metadata.Stability),
+		receiver.WithLogs(createLogsReceiver, metadata.Stability),
+	)
+}
+
+func createDefaultConfig() component.Config {
+	return &Config{
+		ScraperControllerSettings: scraperhelper.ScraperControllerSettings{
+			CollectionInterval: 24 * time.Hour,
```
JonathanWamsley, line 47: The default shows 12h in the scope doc. Is this correct?

receiver/m365receiver/factory.go
```diff
@@ -0,0 +1,97 @@
...
+		},
+		HTTPClientSettings: confighttp.HTTPClientSettings{
+			Timeout: 10 * time.Second,
+		},
+		MetricsBuilderConfig: metadata.DefaultMetricsBuilderConfig(),
+		Logs: &LogsConfig{
+			PollInterval:   5 * time.Minute,
```
JonathanWamsley, line 54: Is this a sensible default for this log receiver?

JonathanWamsley (changes requested):
Looks pretty good so far! 
Can you  prettify the json output though, then I'll start verifying the logs/metrics output. Thanks

receiver/m365receiver/factory.go
dpaasman00: Not too sure, Microsoft doesn't suggest a polling interval. That default was picked pretty arbitrarily. The most Microsoft says on the topic is in the screenshot below. In my limited testing, it seems like it can take up to an hour for logs to be reported, but I'm sure that can change as the load does. As for the default I'm open to a different one if you think it'd be better.
Screenshot 2023-05-30 at 8 07 14 AM

JonathanWamsley: If it takes an hour for logs to reported, then a polling frequency of 5 minutes may be quite excessive. Perhaps a default of 15 minutes would be more appropriate. I can't really find any other information to go off of. What are your thoughts  @dehaansa ?

JonathanWamsley:
@dpaasman00 
For the tests, I usually compare integration tests that use in a docker container to the unit test to see if all the metrics are being received. In this case they are both mocked so it is hard to tell. Can you put the output from your live instance for the metrics/logs?  Thanks

receiver/m365receiver/factory.go
dehaansa: While there may be an initial gap before logs are available, logs should then be somewhat consistently populating and available. 5 minutes seems like a reasonable number to me, to balance responsiveness & API usage.

JonathanWamsley (approved):

receiver/m365receiver/metadata.yaml
```diff
@@ -0,0 +1,205 @@
...
+    attributes: []
+  m365.teams.device_usage.count:
+    enabled: true
+    description: The number of unique users by device/platform that have used Teams in the last 7 days.
+    unit: "{users}"
+    sum:
+      value_type: int
+      monotonic: false
+      aggregation: cumulative
+    attributes: [teamsDevices]
```
BinaryFissionGames, line 97: I think working `users` into this metric name would be helpful to tell what it is at a glance.
Something like `m365.teams.device_usage.users.count` or even just `m365.teams.device_usage.users`

receiver/m365receiver/metadata.yaml
```diff
@@ -0,0 +1,205 @@
...
+  m365.teams.meetings.count:
+    enabled: true
+    description: The number of MS Teams meetings for users in the organization in the last 7 days.
+    unit: "{meetings}"
+    sum:
+      value_type: int
+      monotonic: false
+      aggregation: cumulative
+    attributes: []
+  m365.teams.message.team.count:
```
BinaryFissionGames, line 116: Just to match the `m365.teams.messages` namespace of the metric below:
```suggestion
  m365.teams.messages.team.count:
```

receiver/m365receiver/metadata.yaml
```diff
@@ -0,0 +1,205 @@
...
+  m365.outlook.email_activity.count:
+    enabled: true
+    description: The number of email actions by members over the period of time in the organization Outlook.
+    unit: "{emails}"
+    sum:
+      value_type: int
+      monotonic: false
+      aggregation: cumulative
+    attributes: [outlookActivity]
+  m365.outlook.storage.count:
```
BinaryFissionGames, line 179: Same here, I think used instead of count makes more sense.
```suggestion
  m365.outlook.storage.used:
```

receiver/m365receiver/logs.go
```diff
@@ -0,0 +1,514 @@
...
+	case 5:
+		return "Application"
+	case 6:
+		return "ServicePrincipal"
+	case 7:
+		return "CustomPolicy"
+	case 8:
+		return "SystemPolicy"
+	default:
+		return "impossible"
```
BinaryFissionGames, line 436: For these switch statements, I think "unknown" might be better, they could be hit and show up in logs if something changes on the MS side.

receiver/m365receiver/client.go
```diff
@@ -0,0 +1,441 @@
...
+	// follows redirect link(s) to get json array of logs
+	// parse json array into individual strings, still in json
+	// append to data each body and corresponding json struct
+	var data = []logData{}
+	for _, l := range redirectLinks {
+		body, err := m.followLink(ctx, l)
+		if err != nil {
+			return []logData{}, err
+		}
+		bodies := m.parseBody(body)
```
BinaryFissionGames, line 289: I see what you're doing here with parseBody, but I think there's a better approach here than doing string manipulations like this, by unmarshalling into a slice of json.RawMessage first.

Here's a short example of what I mean:
https://go.dev/play/p/0L01vLDlLTs?v=goprev

BinaryFissionGames (changes requested):
Could you try and rebase onto/ merge the release/v1.17.0 branch? I think the diff is looking a little funky because of it.

Overall looking good, just have a few comments.

receiver/m365receiver/config.go
```diff
@@ -0,0 +1,78 @@
...
+		return fmt.Errorf("client_id is invalid; must be a GUID")
+	}
+
+	if c.ClientSecret == "" {
+		return fmt.Errorf("missing client_secret; required")
+	}
+	re := regexp.MustCompile("^[a-zA-Z0-9-_.~]{1,40}$")
+	if !re.MatchString(c.ClientSecret) {
+		return fmt.Errorf("client_secret is invalid; does not follow correct structure")
+	}
```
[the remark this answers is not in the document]
dpaasman00: I just grabbed those from somewhere in microsoft's doc, they said a client secret would always follow that structure

BinaryFissionGames (approved):
Seems to be working great for me, nice work!

PR #1059, opened 2023-05-11
feat: Add remove empty values processor

BinaryFissionGames:
### Proposed Change
* Adds the remove empty values processor, which removes empty values from resource attributes, attributes, and log record bodies.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


processor/removeemptyvaluesprocessor/factory.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func createTracesProcessor(
+	ctx context.Context,
+	set processor.CreateSettings,
+	cfg component.Config,
+	nextConsumer consumer.Traces,
+) (processor.Traces, error) {
+	oCfg := cfg.(*Config)
+	sp := newEmptyValueProcessor(set.Logger, *oCfg)
```
cpheps, line 66: nit: The other variables for `newEmptyValueProcessor` are `tmp`

processor/removeemptyvaluesprocessor/factory.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+func createTracesProcessor(
+	ctx context.Context,
+	set processor.CreateSettings,
+	cfg component.Config,
+	nextConsumer consumer.Traces,
+) (processor.Traces, error) {
+	oCfg := cfg.(*Config)
```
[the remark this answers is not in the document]
BinaryFissionGames: Good call, I think I copied this piece from the sampling processor, so we might want to update it there.

I do think it's probably safe, but we should definitely protect from panics wherever possible.

processor/removeemptyvaluesprocessor/config.go
```diff
@@ -0,0 +1,75 @@
...
+// MapKey represents a key into a particular map (denoted by field)
+type MapKey struct {
+	field string
+	key   string
+}
+
+// UnmarshalText unmarshals the given []byte into a MapKey.
+// The format of the key is "<field>.<path-to-key>"
+func (m *MapKey) UnmarshalText(text []byte) error {
+	field, key, found := bytes.Cut(text, []byte("."))
```
cpheps, line 39: Can I specify `body`, `resource`, or `attributes` to ignore all of that.  This seems to indicate no but I can also see specifying `body` in the exclude is the same as `enable_log_body = false`.

processor/removeemptyvaluesprocessor/config.go
```diff
@@ -0,0 +1,75 @@
...
+	if !found {
+		return fmt.Errorf("failed to determine field: %s", text)
+	}
+
+	if len(key) == 0 {
+		return fmt.Errorf("key part of (%s) must be non-zero in length", text)
+	}
+
+	for _, validField := range []string{attributesField, resourceField, bodyField} {
```
cpheps, line 48: nit: can we store this slice off somewhere so we don't recreate it on ever Unmarshal?

processor/removeemptyvaluesprocessor/config.go
```diff
@@ -0,0 +1,75 @@
...
+	EnableResourceAttributes bool     `mapstructure:"enable_resource_attributes"`
+	EnableAttributes         bool     `mapstructure:"enable_attributes"`
+	EnableLogBody            bool     `mapstructure:"enable_log_body"`
+	EmptyStringValues        []string `mapstructure:"empty_string_values"`
+	ExcludeKeys              []MapKey `mapstructure:"exclude_keys"`
+}
+
+// Validate validates the processor configuration
+func (cfg Config) Validate() error {
+	return nil
```
cpheps, line 74: Are we effectively shoving the validating into the UnmarshalText function?

processor/removeemptyvaluesprocessor/processor.go
```diff
@@ -0,0 +1,266 @@
...
+	c                    Config
+	excludeResourceKeys  []MapKey
+	excludeAttributeKeys []MapKey
+	excludeBodyKeys      []MapKey
+}
+
+func newEmptyValueProcessor(logger *zap.Logger, cfg Config) *emptyValueProcessor {
+	var excludeResourceKeys []MapKey
+	var excludeAttributeKeys []MapKey
+	var excludeBodyKeys []MapKey
```
cpheps, line 39: Would it be better to store these as a Set basically a `map[string]struct{}` so searching for a match is O(1) rather than O(len(exclude))?

processor/removeemptyvaluesprocessor/processor.go
```diff
@@ -0,0 +1,266 @@
...
+// Any keys that don't have the prefix are removed from the returned list.
+func trimMapKeyPrefix(prefix string, keys []MapKey) []MapKey {
+	outKeys := make([]MapKey, 0, len(keys))
+	for _, mk := range keys {
+		if !strings.HasPrefix(mk.key, prefix+".") {
+			// prefix was not found, so this key does not belong to the submap.
+			continue
+		}
+
+		trimmedKey := mk.key[len(prefix)+1:]
```
cpheps, line 184: nit: Might be cleaner to use `strings.TrimPrefix` here. I know it's the same thing but it's easier to read.

processor/removeemptyvaluesprocessor/config.go
cpheps: This is more just putting the thought out there I realize it doesn't work this way.

processor/removeemptyvaluesprocessor/processor.go
BinaryFissionGames: I avoided TrimPrefix because it's less performant here, but maybe I can rework this block to use it (ideally I'd use CutPrefix here, but it's not available in go1.20, so I basically just inlined the code from it here)

processor/removeemptyvaluesprocessor/config.go
BinaryFissionGames: Yeah, this might be a more elegant way to handle it than these boolean values. I'll rework it to work like that, and remove the booleans.

processor/removeemptyvaluesprocessor/config.go
BinaryFissionGames: Pretty much. MapKey can only be created through unmarshalling, and the unmarshal function validates that the string it unmarshals is in the expected format.

I could probably move the field validation down into this Validate function, if that seems better. I think the way I see it is that if the MapKey is not in the expected format, it's not actually a "MapKey", and shouldn't actually unmarshal to one, but I think either way would be fine.

BinaryFissionGames: Actually, thinking about it, I guess an empty MapKey could be created programatically, I'll move that field validation logic into the Validate funciton.

processor/removeemptyvaluesprocessor/processor.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// cleanMap removes empty values from the map, as defined by the config.
+func cleanMap(m pcommon.Map, c Config, excludeKeys map[string]struct{}) {
+	m.RemoveIf(func(s string, v pcommon.Value) bool {
+		for key := range excludeKeys {
+			if key == s {
+				return false
+			}
+		}
```
cpheps, line 175: Since this is a map now I think you can just do a look up.

```go
if _, ok := excludeKeys[key]; ok {
  return false
}
```

cpheps (approved):

processor/removeemptyvaluesprocessor/processor.go
BinaryFissionGames: good catch, kinda defeats the whole purpose without that, huh?

processor/removeemptyvaluesprocessor/README.md
```diff
@@ -0,0 +1,79 @@
...
+```yaml
+receivers:
+  plugin:
+    path: "./plugins/nginx_logs.yaml"
+
+processors:
+  removeemptyvalues:
+    empty_string_values:
+      # Remove fields with the value of "-"
+      - "-"
```
Mrod1598, line 67: might be good to have an example explicitly showing that you need to specify an empty string here if using this field

Mrod1598 (approved):

PR #1104, opened 2023-05-31
deps: Match bindplane-op tool versions

antonblock:
### Proposed Change

- Update tool versions

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,13 +85,13 @@ build-windows-x86:
 install-tools:
 	go install github.com/client9/misspell/cmd/misspell@v0.3.4
 	go install github.com/google/addlicense@v1.1.0
-	go install github.com/goreleaser/goreleaser@v1.16.1
-	go install github.com/mgechev/revive@v1.2.3
+	go install github.com/goreleaser/goreleaser@v1.18.2
+	go install github.com/mgechev/revive@v1.3.1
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -85,13 +85,13 @@ build-windows-x86:
 install-tools:
 	go install github.com/client9/misspell/cmd/misspell@v0.3.4
 	go install github.com/google/addlicense@v1.1.0
-	go install github.com/goreleaser/goreleaser@v1.16.1
-	go install github.com/mgechev/revive@v1.2.3
+	go install github.com/goreleaser/goreleaser@v1.18.2
+	go install github.com/mgechev/revive@v1.3.1
```
cpheps, line 89: I think the github actions need to be updated to use these versions of the tools.

cpheps (changes requested):

Makefile
antonblock: I don't see revive in the actions

cpheps: Yeah looks like we just use their action. I double checked. Thanks for updating goreleaser.

cpheps (approved):

BinaryFissionGames (approved):

PR #1123, opened 2023-06-19
feat: Add OTTL to extract metric processor

BinaryFissionGames:
### Proposed Change
* Adds OTTL fields to the extract metric processor

The expr fields are again kept to provide backwards compatability. The processor can only be configured in OTTL mode or expr mode, and the fields cannot be mixed.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


processor/metricextractprocessor/config.go
```diff
@@ -50,31 +54,58 @@ const (
...
 type Config struct {
-	Route      string            `mapstructure:"route"`
-	Match      string            `mapstructure:"match"`
-	Extract    string            `mapstructure:"extract"`
-	MetricName string            `mapstructure:"metric_name"`
-	MetricUnit string            `mapstructure:"metric_unit"`
-	MetricType string            `mapstructure:"metric_type"`
-	Attributes map[string]string `mapstructure:"attributes"`
+	Route          string            `mapstructure:"route"`
+	Match          *string           `mapstructure:"match"`
```
Mrod1598, line 70: have we alked about how deprecation looks like in this repo yet? are we just following otel standard of 2 releases i think it is? do we have our deperecation schedule anywhere?

processor/metricextractprocessor/config.go
```diff
@@ -50,31 +54,58 @@ const (
...
+	MetricUnit     string            `mapstructure:"metric_unit"`
+	MetricType     string            `mapstructure:"metric_type"`
+	Attributes     map[string]string `mapstructure:"attributes"`
+	OTTLAttributes map[string]string `mapstructure:"ottl_attributes"`
 }
 
 // Validate validates the config.
 func (c Config) Validate() error {
+	usesExprFields := c.Extract != "" || c.Match != nil || c.Attributes != nil
+	usesOTTLFields := c.OTTLExtract != "" || c.OTTLMatch != nil || c.OTTLAttributes != nil
```
Mrod1598, line 84: nit: you already have IsOTTL below. might be easier to just use that?

Mrod1598 (approved):

processor/metricextractprocessor/config.go
BinaryFissionGames: I think it might be confusing here why isOTTL is relevant. It's also not a direct substitute I don't think? Mostly I just feel it's clearer written this way, though.

PR #1126, opened 2023-06-23
fix: Add timeouts to collector shutdown

BinaryFissionGames:
### Proposed Change
This change adds timeouts to the internal collector shutdown.

This means that the collector will not take large amounts of time to shutdown. This is important during opamp, where long shutdowns can make the collector seem like it's hung up, and also during shutdown on Windows, where the service manager does not kill the service when it refuses to shutdown nicely.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -23,16 +23,20 @@ import (
 	"sync"
 	"time"
 
+	"github.com/observiq/observiq-otel-collector/factories"
 	"go.opentelemetry.io/collector/otelcol"
 	"go.uber.org/zap"
 )
+
+// The timeout for how long to wait on shutting down the collector before bailing out during a Restart.
+var collectorRestartTimeout = 10 * time.Second
 
 // Collector is an interface for running the open telemetry collector.
 //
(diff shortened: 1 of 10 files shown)


collector/collector.go
```diff
@@ -154,14 +181,34 @@ func (c *collector) Stop() {
...
 
 	c.svc.Shutdown()
+
+	shutdownCompleteChan := make(chan struct{})
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			// Cancel the start context if the stop context is cancelled.
+			// Ideally, we'd be able to pass a context into shutdown, but the OTEL collector
```
Mrod1598, line 191: Is that this something we should try to contribute in the future?

Mrod1598 (approved):

collector/collector.go
BinaryFissionGames: Yeah, I'll look into writing an issue. Seems like it could be useful upstream.

PR #1132, opened 2023-06-27
chore: Updated User Agent from GCP Exporters

cpheps:
### Proposed Change
Updated default user agent in googlecloud and gmp exporters to match Stanza's old default

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jmwilliams89 (approved):

exporter/googlemanagedprometheusexporter/factory.go
```diff
@@ -37,10 +37,10 @@ const (
 )
 
 // NewFactory creates a factory for the googlecloud exporter
-func NewFactory() exporter.Factory {
+func NewFactory(collectorVersion string) exporter.Factory {
```
BinaryFissionGames, line 40: Figured I'd mention that this breaks compatability with the opentelemetry builder tool, https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder

BinaryFissionGames (approved):

exporter/googlemanagedprometheusexporter/factory.go
cpheps: Spoke offline and we're ok with this risk as these changes are on wrappers of existing OTel exporters. Users will likely use the OTel version of those exporters rather than ours.

PR #1135, opened 2023-06-28
fix: Revamp M365 receiver

dpaasman00:
### Proposed Change
- Updates README to be explicit about unusual API (and receiver) behavior and to consistently refer to Microsoft 365 rather than Microsoft Office 365.
- Changes scrape interval for metrics to 1 hour
- Set log body to parsed version rather than a string (also updates log test data with this change)

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/m365receiver/README.md
```diff
@@ -1,8 +1,15 @@
...
 and logs via the [Microsoft Management API](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-schema).
 
+## Important Notes
+Due to the nature of the APIs being used to collect data from Microsoft, there are some important notes that should be acknowledged before continuing.
+
+- Metrics collected by this receiver can be 48 hours old. This is because the Reports endpoint of the Microsoft Graph API can take up to 48 hours to be updated with new data. The timestamps for metrics will represent when the data was scraped from the API rather than when the data points occurred (48 hours earlier). Because of this, the metrics receiver has a strict scrape interval of 1 hour that cannot be changed. This is done so that the receiver can scrape data from the API once it becomes available, within a reasonable amount of time. This also means that most data points are duplicates. Please keep this in mind when viewing metric data. For more info please see Microsoft's [documentation](https://learn.microsoft.com/en-us/microsoft-365/admin/activity-reports/activity-reports?view=o365-worldwide#how-to-get-to-the-reports-dashboard).
```
dehaansa, line 9: The scrape interval _can_ be changed from what I see, and we do want it to be configurable in case a user wishes to collect more (or less) frequently. The 1hr is simply what we currently believe to be the best user experience compromise. I think just remove the bit that says that it's strict/cannot be changed, no need to add more context.

receiver/m365receiver/README.md
```diff
@@ -1,8 +1,15 @@
...
+- As part of [configuring Microsoft 365](#configuring-microsoft-365) a client secret will need to be generated in Azure Active Directory. When this is done, a date for when the secret expires will need to be selected. Remember and keep track of this date in some manner. When the client secret expires the receiver will fail and need to be restarted with a newly generated client secret.
```
dehaansa, line 11: > Remember and keep track of this date in some manner. When the client secret expires the receiver will fail and need to be restarted with a newly generated client secret.
I think it's clearer to just say that 'the user is responsible for tracking expiration and acquiring new secrets'

receiver/m365receiver/README.md
```diff
@@ -1,8 +1,15 @@
-# Microsoft Office 365 Receiver
-Receives metrics from [Microsoft Office 365](https://www.microsoft365.com/)
+# Microsoft 365 Receiver
+Receives metrics from [Microsoft 365](https://www.microsoft365.com/)
 via the [Microsoft Graph API](https://learn.microsoft.com/en-us/graph/api/overview?view=graph-rest-1.0&preserve-view=true),
 and logs via the [Microsoft Management API](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-schema).
 
+## Important Notes
+Due to the nature of the APIs being used to collect data from Microsoft, there are some important notes that should be acknowledged before continuing.
```
dehaansa, line 7: from Microsoft 365*

jsirianni (approved):
:+1:

receiver/m365receiver/logs.go
```diff
@@ -246,8 +246,17 @@ func (l *m365LogsReceiver) transformLogs(now pcommon.Timestamp, audit *auditMeta
 	for _, log := range data {
 		logRecord := scopeLogs.LogRecords().AppendEmpty()
 
-		// log body
-		logRecord.Body().SetStr(log.body)
+		// parses body string and sets that as log body, but uses string if parsing fails
+		parsedBody := map[string]interface{}{}
```
BinaryFissionGames, line 250: nit: `any` is an alias for `interface{}`, you'll see `interface{}` in a lot of old code. We should use `any` in new code we write.
```suggestion
		parsedBody := map[string]any{}
```

BinaryFissionGames (approved):

dehaansa (approved):

PR #1113, opened 2023-06-08
fix: Build 32-bit arm images for raspberry pi devices

proffalken:
### Proposed Change

A simple change to the build config to provide Ubuntu 32-bit armv7 docker images.

Fixes #1112 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

PR #1136, opened 2023-06-29
deps(container): Replaced deprecated openjdk container image with eclipse-temurin

cpheps:
### Proposed Change
The [openjdk](https://hub.docker.com/_/openjdk) container image is deprecated. It has been replaced with a suggested alternative [eclipse-temurin](https://hub.docker.com/_/eclipse-temurin).

This should resolve the container issues in #1113 as the new container supports `arm32v7`

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/docker/Dockerfile.ubi8
+++ b/docker/Dockerfile.ubi8
@@ -23,11 +23,10 @@ RUN curl -L \
     "https://github.com/open-telemetry/opentelemetry-java-contrib/releases/download/${JMX_JAR_VERSION}/opentelemetry-jmx-metrics.jar"
 
 
-# OpenJDK stage provides the Java runtime used by JMX receivers.
-# Contrib's integration tests use openjdk 1.8.0
-# https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/jmxreceiver/testdata/Dockerfile.cassandra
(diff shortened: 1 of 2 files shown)


jsirianni (approved):
Tested ubuntu / ubi images for arm6 and amd64 and they all give me the right java build

```bash
otel@6cd7668b2da3:~$ java -version
openjdk version "1.8.0_372"
OpenJDK Runtime Environment (Temurin)(build 1.8.0_372-b07)
OpenJDK 64-Bit Server VM (Temurin)(build 25.372-b07, mixed mode)
```

ARM64 images give the right architecture of java
```bash
root@2bde7ba52ede:/etc/otel# file $(which java)
/opt/java/openjdk/bin/java: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, not stripped
```

Same with amd64
```bash
/opt/java/openjdk/bin/java: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.18, not stripped
```

BinaryFissionGames (approved):

PR #1141, opened 2023-07-05
Upgrade OpAMP library dependency to the latest commit

andykellr:
### Proposed Change
Upgrade opamp-go to v0.9.0. There are no functional changes right now, just protobuf naming changes.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


opamp/observiq/observiq_client.go
```diff
@@ -210,6 +210,13 @@ func (c *Client) Connect(ctx context.Context) error {
 			// SaveRemoteConfigStatusFunc
 		},
 		PackagesStateProvider: c.packagesStateProvider,
+		Capabilities: protobufs.AgentCapabilities_AgentCapabilities_ReportsStatus |
```
cpheps, line 213: I realized we haven't gotten to tests yet but can we make sure we add a check in the `TestNewClient` test that the capabilities match what we expect?

PR #1026, opened 2023-04-13
feat(Apache Druid)

mrsillydog:
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

PR #1155, opened 2023-07-17
feat(Create Apache Druid logs plugin)

mrsillydog:
### Proposed Change
Adding a plugin for logs collection from an Apache Druid instance.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

To get example logs, Druid can be spun up through the docs [here](https://hub.docker.com/r/apache/druid).

This PR is part of https://github.com/observIQ/observiq-otel-collector/pull/1026 - split up into two, a receiver and a plugin, per the recommendation in the original PR.


plugins/apache_druid_logs.yaml
```diff
@@ -0,0 +1,296 @@
...
+    type: bool
+    default: true
+  - name: broker_log_path
+    description: Absolute filepath containing Apache druid broker logs. Required if enable_broker_logs is true
+    type: "string"
+  - name: enable_coordinator_overlord_logs
+    description: Enable to collect Apache Druid coordinator and overlord logs
+    type: bool
+    default: true
+  - name: coordinator_overlord_log_path
```
dehaansa, line 16: Should these all be marked as `required: true` and have a `relevant_if` block for only relevant if their `enabled_` is true? Probably a question for BPOP team for what the standards are.

plugins/apache_druid_logs.yaml
```diff
@@ -0,0 +1,296 @@
...
+  - name: start_at
+    description: At startup, where to start reading logs from the file (`beginning` or `end`)
+    type: string
+    valid_values:
+     - beginning
+     - end
+    default: end
+  # Apache Druid logs do not include a timezone in their timestamps, so we need to set it explicitly.
+  - name: timezone
+    description: Timezone to use when parsing the timestamp
```
dehaansa, line 56: Description should state the condition that the logs timestamp don't contain a timezone.

plugins/apache_druid_logs.yaml
dsvanlani: Yeah thats how we usually do this sort of thing, you'd want

```yaml
relevantIf:
        - name: enable_coordinator_overlord_logs
          operator: equals
          value: "regex"
required: true
```

Mrod1598: yes that's usually how we try to do these, is showing only relevant. Starting to get some standards docs up (destinations and sources incoming eventually): https://github.com/observIQ/engineering-standards/blob/main/bindplane-development/processor-standards.md. but not seeing anything related to relevantifs on there yet. I'll bring it up in our code quality meeting.

docs/plugins/apache_druid_logs.md
```diff
@@ -0,0 +1,45 @@
...
+## Example Config:
+
+Below is an example of a basic config
+
+```yaml
+receivers:
+  plugin:
+    path: ./plugins/apache_druid_logs.yaml
+    parameters:
+      enable_broker_logs: true
```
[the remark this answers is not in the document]
mrsillydog: This file is autogenerated by `make create-plugin-docs`, and I recall that doing anything besides using the autogenerated version breaks CI. Unfortunately, the make command just seems to populate the sample config based on default values, of which there isn't one for the filepaths.

cpheps:
Closing due to staleness this can be reopened in the future.

PR #1165, opened 2023-07-21
fix(plugin): Container logs: Exclude bindplane-agent logs by default

jsirianni:
### Proposed Change

BindPlane deploys the agent with the pod name `bindplane-node-agent-*` and `bindplane-cluster-agent-*`. These pods should be excluded by default, to prevent circular logging.

Using BindPlane, I can see the agent's logs by filtering on `k8s.namespace.name: bindplane-agent`:

Screenshot from 2023-07-21 18-19-43

Screenshot from 2023-07-21 18-25-31

To simulate this change, I had to update the source configuration with the same exclusion:

Screenshot from 2023-07-21 18-31-07

With the filter in place, and the exclusion, the agent is reporting zero logs (as expected).

##### Checklist
- [x] Changes are tested
- [x] CI has passed


cpheps (approved):
Can you bump the plugin version?

BinaryFissionGames (approved):

PR #1171, opened 2023-08-01
deps: Update OTel to v0.82.0

BinaryFissionGames:
### Proposed Change
* Updates opentelemetry to [v0.82.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.82.0)
* Updates JMX jar to [v1.28.0](https://github.com/open-telemetry/opentelemetry-java-contrib/releases/tag/v1.28.0)

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


JAVA_CONTRIB_VERSION
```diff
@@ -1 +1 @@
-v1.15.0
```
Mrod1598, line 1: thats a jump, what's this file for?

Mrod1598 (approved):

JAVA_CONTRIB_VERSION
BinaryFissionGames: It tracks the version of the JMX jar, used by the JMX receiver; We haven't updated in a while, but we really should be keeping up to date.

antonblock (approved):

JAVA_CONTRIB_VERSION
BinaryFissionGames: Also, to clarify, the reason why we moved to update it today is because a vulnerability scan found a few out-of-date dependencies in that jar, and we're looking to rectify that.

PR #1191, opened 2023-08-10
fix: When shutting down on windows, set a max timeout of 20 seconds

BinaryFissionGames:
### Proposed Change
* When shutting down, we cap the time for stopping to 20 seconds
  * This fixes issues where shutting down the service can be unresponsive, and you get an error message in Service Manager
  * This fixes an issue where the process living too long can cause the update process to fail on windows.
* Wait for the windows service to actually be stopped in the update logic
  * This patches a race condition in the update path, where the collector may fail to properly update if the process is not actually stopped by the time it tries to overwrite to collector binary
* Update the updater logging for windows to match Linux, and use ISO8601 timestamps instead of unix timestamps.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/internal/service/service_windows.go
+++ b/internal/service/service_windows.go
@@ -158,7 +158,7 @@ func (sh windowsServiceHandler) shutdown(s chan<- svc.Status) error {
 	var err error
 	select {
 	case <-time.After(windowsServiceShutdownTimeout):
-		err = errors.New("the service failed to shut down in a timely manner")
+		err = fmt.Errorf("the service failed to shut down in a timely manner (timeout: %s)", windowsServiceShutdownTimeout)
 	case stopErr := <-stopErrChan:
 		err = stopErr
 	}
--- a/internal/service/service_windows.go
+++ b/internal/service/service_windows.go
@@ -24,11 +24,16 @@ import (
 	"os/signal"
 	"path/filepath"
 	"syscall"
+	"time"
 
 	"go.uber.org/zap"
 	"golang.org/x/sys/windows"
 	"golang.org/x/sys/windows/svc"
 )
+
+// windowsServiceShutdownTimeout is the amount of time to wait for the underlying service to stop before
+// forcefully stopping the process.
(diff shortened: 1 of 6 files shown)


internal/service/service_windows.go
```diff
@@ -145,7 +150,18 @@ func (sh windowsServiceHandler) shutdown(s chan<- svc.Status) error {
...
-	err := sh.svc.Stop(stopTimeoutCtx)
+	stopErrChan := make(chan error, 1)
+	go func() {
+		stopErrChan <- sh.svc.Stop(stopTimeoutCtx)
+	}()
+
+	var err error
+	select {
+	case <-time.After(windowsServiceShutdownTimeout):
+		err = errors.New("the service failed to shut down in a timely manner")
```
cpheps, line 161: Should we os.Exit(1) here or is this ok to just exit as is? I'm ok with this just wanted to ask.

cpheps (approved):

internal/service/service_windows.go
BinaryFissionGames: What I like about not calling `os.Exit(1)` here is that it properly sets the service state to stopped and exits through the service handler using whatever communication exists there. 

I'm actually not 100% sure on the benefits of that, or if it's necessary, but that's my reasoning on doing it this way.

BinaryFissionGames:
I tested that:
* The original issue exists (host metrics -> GCP with blocking all outbound traffic causes it to take over a minute to shutdown, upgrade failed such that the agent did not start back up)
* When upgrading from this version to a newer version, the issue is fixed
* The service shuts down properly from the service manager when overloaded like this (after 20 seconds, the service shuts down)
* Updater logs are now displaying debug logs (like they do on linux)
* Updater logs have ISO8601 timestamps now

internal/service/service_windows.go
antonblock: Maybe redundant, but I'd be in favor of including how long we waited in this message

antonblock (approved):

PR #1209, opened 2023-08-29
Wait for collector to properly shut down in plugin receiver

BinaryFissionGames:
### Proposed Change
Wait for the collector's Run function to return before considering shutdown complete

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/pluginreceiver/receiver.go
+++ b/receiver/pluginreceiver/receiver.go
@@ -34,6 +34,8 @@ type Receiver struct {
 	logger         *zap.Logger
 	createService  createServiceFunc
 	service        Service
+
+	doneChan chan struct{}
 }
 
 // Start starts the receiver's internal service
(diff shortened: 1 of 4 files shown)


receiver/pluginreceiver/receiver.go
```diff
@@ -34,6 +34,8 @@ type Receiver struct {
 	logger         *zap.Logger
 	createService  createServiceFunc
 	service        Service
+
+	doneChan chan struct{}
```
cpheps, line 38: Does a wait group make more sense here? Maybe it makes it so you don't have to init a new channel in tests?

I think either one works so just wanted to bring it up, not pushing for it.

BinaryFissionGames: Wait group could work too, for sure. I personally prefer the channel here just because it means I don't have to do the whole extra goroutine pattern in shutdown to utilize the context.

jmwilliams89 (approved):

PR #1215, opened 2023-09-05
Pass Version into OpAMP Client

cpheps:
### Proposed Change
Modified to pass the version into the OpAMP client rather than pulling it from the version package. This way we can easily set collector version in Amplify scaling tool.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):

jsirianni (approved):
Looks good. Tested against BindPlane.

PR #1280, opened 2023-10-12
feat: Azure Blob Rehydration Receiver

cpheps:
### Proposed Change
Adds the Azure Blob Rehydration Receiver. The receiver is meant to be used to rehydrate OTLP data that has been exported using the [Azure Blob Exporter](https://github.com/observIQ/bindplane-agent/tree/release/v1.37.0/exporter/azureblobexporter).

This is not a traditional receiver in the sense that it continuously gathers data. It rather rehydrates all blobs within a specified time range. Then rechecks on an interval if any more blobs were added to that time range. It will not produce any more data after the initial rehydration is completed.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


receiver/azureblobrehydrationreceiver/config.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"errors" // timeFormat is the format for the starting and end time
+	"fmt"
+	"time"
+
+	"go.opentelemetry.io/collector/component"
+)
+
+const timeFormat = "2006-01-02T15:04"
```
jsirianni, line 25: Did you mean to have the comment above the const?

```suggestion
	"errors"
	"fmt"
	"time"

	"go.opentelemetry.io/collector/component"
)

// timeFormat is the format for the starting and end time
const timeFormat = "2006-01-02T15:04"
```

jsirianni (commented):
Couple small comments. I still need to build and test.

jsirianni (commented):
Could ending_time be made optional? If unset, it runs forever? I think the hacky workaround is to set before / ending dates to extremely far out values. Is this a bad idea?

Couple timestamp questions
1. Should we document how timestamps are preserved? Reading metrics from months in the past will not be useful on most platforms (I assume)
2. The Azure Blob exporter is not setting log timestamps (or should this be the job of the producing receiver?). This might be outside of this PR's scope but I think we should address it.

Screenshot from 2023-10-12 09-53-35

I am using bindplane custom source with an agent built from this branch.

receiver/azureblobrehydrationreceiver/config.go
cpheps: Yeah not sure what happened there good catch.

cpheps:
> Could ending_time be made optional? If unset, it runs forever? I think the hacky workaround is to set before / ending dates to extremely far out values. Is this a bad idea?

I ran this by product originally and the request was for an exact rehydration period. 

> Should we document how timestamps are preserved? Reading metrics from months in the past will not be useful on most platforms 

I'm not sure I really do think rehydration is most effective for logs or moving between say Azure and S3 storage for any OTLP type. I think this receiver is a very specific use case and we may need to iterate on it if it doesn't quite fill customer needs as is.

> The Azure Blob exporter is not setting log timestamps (or should this be the job of the producing receiver?). This might be outside of this PR's scope but I think we should address it.

I think this is the original receiver. The exporter should just turn the logs into OTLP json and store them.

jsirianni:
> > The Azure Blob exporter is not setting log timestamps (or should this be the job of the producing receiver?). This might be outside of this PR's scope but I think we should address it.
> 
> I think this is the original receiver. The exporter should just turn the logs into OTLP json and store them.

My mistake, the log timestamps are correct:
```
ObservedTimestamp: 2023-10-12 13:41:00.867006207 +0000 UTC\n
Timestamp: 1970-01-01 00:00:00 +0000 UTC
```

jsirianni (approved):

receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -0,0 +1,417 @@
...
+
+// strings that indicate what type of telemetry is in a blob
+const (
+	metricBlobSignifier = "metrics_"
+	logsBlobSignifier   = "logs_"
+	tracesBlobSignifier = "traces_"
+)
+
+// parseBlobPath returns true if the blob is within the existing time range
+func parseBlobPath(prefix *string, blobName string) (blobTime *time.Time, telemetryType component.DataType, err error) {
```
BinaryFissionGames, line 288: Just wondering if it's worth it to do all this state machine stuff over using regex and capture groups? Might be easier to understand.

receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -0,0 +1,417 @@
...
+
+	if err := r.consumer.Consume(r.ctx, blobBuffer); err != nil {
+		return fmt.Errorf("consume: %w", err)
+	}
+	return nil
+}
+
+// gzipDecompress does a gzip decompression on the passed in contents
+func gzipDecompress(contents []byte) ([]byte, error) {
+	gr, err := gzip.NewReader(bytes.NewBuffer(contents))
```
BinaryFissionGames, line 406: We should close this and check the error at the end of the function, I think it's possible close returns a checksum error.

receiver/azureblobrehydrationreceiver/config.go
```diff
@@ -0,0 +1,102 @@
...
+	// EndingTime the UTC timestamp to rehydrate up until.
+	EndingTime string `mapstructure:"ending_time"`
+
+	// DeleteOnRead indicates if a file should be deleted once it has been processed
+	// Default value of false
+	DeleteOnRead bool `mapstructure:"delete_on_read"`
+
+	// PollInterval is the interval at which the Azure API is scanned for blobs.
+	// Default value of 1m
+	PollInterval time.Duration `mapstructure:"poll_interval"`
```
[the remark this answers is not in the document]
cpheps: Yeah it's a good call out. I originally didn't as this isn't really a scraper it's more a one and done. The poll interval is really just so we don't busy spin. I can look at adding it though

receiver/azureblobrehydrationreceiver/receiver.go
cpheps: You are 100% I think I over engineered this problem. Let me look at using regexes.

receiver/azureblobrehydrationreceiver/config.go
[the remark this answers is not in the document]
cpheps: So I just tried this and looks like scrapers are only for metrics? We want scraping behavior but we're dealing with all telemetry types. I factored out the main scraping logic into a function so I could emulate a do while loop. So calling the initial iteration before we enter the loop.

receiver/azureblobrehydrationreceiver/receiver.go
cpheps: Ok so I started going the regex route and I found it was going to be almost exactly as complex code wise. The main thing I saw was parsing the timestamp. The regex gives me an array of capture groups that I have to iterate through and either construct a string like I am now  and pass it to `time.Parse` or do a bunch of string conversion and checking to use `time.Date`. The latter actually seemed to produce more code as I have to error check all the string conversions rather than just checking the error from `time.Parse`. 

I'd be happy to discuss this if I'm missing something but it didn't seem like I was gaining much simplicity when I tried to implement it.

BinaryFissionGames: OK, this is what I came up with:
```
var blobNameRegex = regexp.MustCompile(`^(?:[^/]*/)?year=(\d{4})/month=(\d{2})/day=(\d{2})/hour=(\d{2})/(?:minute=(\d{2})/)?([^/].*)$`)

// parseBlobPath returns true if the blob is within the existing time range
func parseBlobPath(prefix *string, blobName string) (blobTime *time.Time, telemetryType component.DataType, err error) {
	matches := blobNameRegex.FindStringSubmatch(blobName)
	if matches == nil {
		err = errInvalidBlobPath
		return
	}

	year := matches[1]
	month := matches[2]
	day := matches[3]
	hour := matches[4]

	minute := "00"
	if matches[5] != "" {
		minute = matches[5]
	}

	lastPart := matches[6]

	timeString := fmt.Sprintf("%s-%s-%sT%s:%s", year, month, day, hour, minute)

	// Parse the expected format
	parsedTime, timeErr := time.Parse(timeFormat, timeString)
	if timeErr != nil {
		err = fmt.Errorf("parse blob time: %w", timeErr)
		return
	}
	blobTime = &parsedTime

	switch {
	case strings.Contains(lastPart, metricBlobSignifier):
		telemetryType = component.DataTypeMetrics
	case strings.Contains(lastPart, logsBlobSignifier):
		telemetryType = component.DataTypeLogs
	case strings.Contains(lastPart, tracesBlobSignifier):
		telemetryType = component.DataTypeTraces
	}

	return
}
```

This fails one of the tests, where a different error is returned (year=nope would not match the regex since nope is not digits), but otherwise it passes tests.

What I like about the regex solution is that it totally avoids that state-machine loop, which can be kinda hard to follow due to how the control flow jumps around, while this solution reads easily from top to bottom.

cpheps: Solution looks good. Yeah the test, needed to be tweaked to trigger the error it was targeting. I'm pushing up that change.

BinaryFissionGames (approved):

PR #1325, opened 2023-11-02
chore: Removed archived k8s reference

cpheps:
### Proposed Change
Removed reference to archived Kubernetes repo.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (approved):

BinaryFissionGames:
This repo is also referenced in `docs/getting-started.md` and `config/google_cloud_exporter/kubernetes/README.md`.

(Side note, `docs/getting-started.md` seems outdated and like it was written for a different format, we might want to rework that one in the future)

cpheps:
>(Side note, docs/getting-started.md seems outdated and like it was written for a different format, we might want to rework that one in the future)

Good call I'll create a ticket for new sprint to clean this up.

cpheps:
Removed the additional references. I'm unsure if the kubernetes deployment example still makes sense with that missing.

BinaryFissionGames:
> Removed the additional references. I'm unsure if the kubernetes deployment example still makes sense with that missing.

It seems to me like it was just a link to that repo for the example config, basically. So we should probably just remove that whole kubernetes directory.

cpheps:
> > Removed the additional references. I'm unsure if the kubernetes deployment example still makes sense with that missing.
> 
> It seems to me like it was just a link to that repo for the example config, basically. So we should probably just remove that whole kubernetes directory.

@jsirianni Can you give your input here. I'm happy to remove it but I want to double check on the intent of that doc.

BinaryFissionGames (approved):

PR #1330, opened 2023-11-08
feat: Add default HTTP log receiver

dpaasman00:
### Proposed Change
This PR adds a new receiver to the bindplane-agent which serves as a default HTTP log receiver. In actuality this was created for the sole purpose of ingesting logs from the Datadog-Agent. The receiver is made and described as a "default HTTP receiver" but the main focus of it is for this specific use case. As a result there are likely several limitations that prevent it from being a "default HTTP log receiver", but that is okay for the time being. 

The correct and eventual solution for this use case and others like it is creating an HTTP input operator for Stanza and then making a receiver built around that, similar to the TCPLog receiver and Stanza TCP input operator. However that approach was not taken because we needed a solution for the Datadog-Agent use case quickly and the Stanza HTTP operator would take too long to realize. 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/httpreceiver/logs.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	return pLogs
+}
+
+// parsePayload transforms the payload into []map[string]interface structure
+func parsePayload(payload []byte) ([]map[string]interface{}, error) {
+	rawLogsArray := []json.RawMessage{}
+	rawLogObject := json.RawMessage{}
+	if err := json.Unmarshal(payload, &rawLogsArray); err != nil {
```
schmikei, line 222: I'm curious if it would be more efficient to find the first non-whitespace character and check for `[` or `{` to do specify the decoding since unmarshalling is sort of a heavy call. Maybe not necessary for initial PR but an idea.

schmikei (commented):
Looks mostly good I think, I think I still need to test but had one potential optimization that may or may not be worth doing

receiver/httpreceiver/logs.go
dpaasman00: no that makes sense to me, especially if Unmarshal is a heavier call

schmikei: I was more thinking of something like this, 

```
// parsePayload transforms the payload into []map[string]interface structure
func parsePayload(payload []byte) ([]map[string]interface{}, error) {
	firstChar := seekFirstNonWhitespace(string(payload))
	switch firstChar {
	case "{":
		rawLogObject := json.RawMessage{}
		if err := json.Unmarshal(payload, &rawLogObject); err != nil {
			return nil, err
		}
		return parseJSONObject(payload)
	case "[":
		rawLogsArray := []json.RawMessage{}
		if err := json.Unmarshal(payload, &rawLogsArray); err != nil {
			return nil, err
		}
		return parseJSONArray(rawLogsArray)
	default:
		return nil, errors.New("unsupported payload format, expected either a JSON object or array")
	}
}

func seekFirstNonWhitespace(s string) string {
	firstChar := ""
	for _, r := range s {
		if unicode.IsSpace(r) {
			continue
		}
		firstChar = string(r)
		break
	}
	return firstChar
}
```

You're current solution does not account for whitespace in the payload.

i.e. 
			payload:      `[{"status": "info", "timestamp": "8402734234", "message": {"data": "hello world"}, "hostname": "localhost"}]`,

will work but

			payload:      ` [{"status": "info", "timestamp": "8402734234", "message": {"data": "hello world"}, "hostname": "localhost"}]`,

will not

schmikei: I also think the parseJSONObject function can just return the first unmarshal no? unless i'm missing something

dpaasman00: yea the whitespace thing is fair, and the switch statement does look cleaner

dpaasman00: the parseJSONObject func doesn't return the first unmarshal cus the parsePayload func returns an array of the map[string]interface{}, so just creating the array in the parseJSONObject so it fits well with the other functions

schmikei: It seems like we don't even need that func though since we can inline declare a slice:

```go
switch firstChar {
	case "{":
		rawLogObject := map[string]interface{}{}
		if err := json.Unmarshal(payload, &rawLogObject); err != nil {
			return nil, err
		}
		return []map[string]interface{}{rawLogObject}, nil
	case "[":
		rawLogsArray := []json.RawMessage{}
		if err := json.Unmarshal(payload, &rawLogsArray); err != nil {
			return nil, err
		}
		return parseJSONArray(rawLogsArray)
	default:
		return nil, errors.New("unsupported payload format, expected either a JSON object or array")
	}
```

receiver/httpreceiver/config.go
```diff
@@ -0,0 +1,68 @@
...
+	"go.opentelemetry.io/collector/config/configtls"
+	"go.uber.org/multierr"
+)
+
+// Config defines the configuration for an HTTP receiver
+type Config struct {
+	Endpoint string                      `mapstructure:"endpoint"`
+	Path     string                      `mapstructure:"path"`
+	TLS      *configtls.TLSServerSetting `mapstructure:"tls"`
+}
```
BinaryFissionGames, line 32: Any reason not to embed HTTPServerSettings? 
https://github.com/open-telemetry/opentelemetry-collector/blob/0482753f5203952be14d07be0cd4a45a571d47ec/config/confighttp/confighttp.go#L214

It looks like it has some pretty robust handling of stuff like max body size and auth extensions when using the ToServer function:
https://github.com/open-telemetry/opentelemetry-collector/blob/0482753f5203952be14d07be0cd4a45a571d47ec/config/confighttp/confighttp.go#L289

receiver/httpreceiver/logs.go
```diff
@@ -0,0 +1,268 @@
...
+// parsePayload transforms the payload into []map[string]any structure
+func parsePayload(payload []byte) ([]map[string]any, error) {
+	firstChar := seekFirstNonWhitespace(string(payload))
+	switch firstChar {
+	case "{":
+		rawLogObject := map[string]any{}
+		if err := json.Unmarshal(payload, &rawLogObject); err != nil {
+			return nil, err
+		}
+		return []map[string]interface{}{rawLogObject}, nil
```
BinaryFissionGames, line 229: super nit:
```suggestion
		return []map[string]any{rawLogObject}, nil
```

receiver/httpreceiver/config.go
dpaasman00: No, not that i'm aware of. I'll change it up to use that.

receiver/httpreceiver/config.go
```diff
@@ -0,0 +1,68 @@
...
+	if _, _, err := net.SplitHostPort(c.Endpoint); err != nil {
+		errs = multierr.Append(errs, errBadEndpoint)
+	}
+	if c.TLS != nil {
+		if c.TLS.CertFile == "" {
+			errs = multierr.Append(errs, errNoCert)
+		}
+		if c.TLS.KeyFile == "" {
+			errs = multierr.Append(errs, errNoKey)
+		}
```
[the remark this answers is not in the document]
dpaasman00: Yea I'll take a look

receiver/httpreceiver/README.md
```diff
@@ -0,0 +1,60 @@
+# HTTP Receiver
+This receiver is capable of collecting logs for a variety of services, serving as a default HTTP log receiver. Anything that is able to send JSON structured logs to an endpoint using HTTP will be able to utilize this receiver.
+
+## Supported Pipelines
```
cpheps, line 4: Can we add a minimum collector version section? See other components for how this section should look.

receiver/httpreceiver/config.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Validate ensures an HTTP receiver config is correct
+func (c *Config) Validate() error {
+	if c.ServerSettings.Endpoint == "" {
+		return errNoEndpoint
+	}
+
+	var errs error
+	if _, _, err := net.SplitHostPort(c.ServerSettings.Endpoint); err != nil {
+		errs = multierr.Append(errs, errBadEndpoint)
```
cpheps, line 49: So in our version of the collector we try not to use multierr were possible. We usually just exit on an error. If you want to join errors use [errors.Join](https://pkg.go.dev/errors#example-Join)

receiver/httpreceiver/logs.go
```diff
@@ -0,0 +1,232 @@
...
+	"go.uber.org/zap"
+)
+
+type httpLogsReceiver struct {
+	path              string
+	serverSettings    *confighttp.HTTPServerSettings
+	telemetrySettings component.TelemetrySettings
+	server            *http.Server
+	consumer          consumer.Logs
+	wg                *sync.WaitGroup
```
cpheps, line 42: Usually wait groups don't need to be pointers until they are passed into functions. Can we change this to a non-pointer? That also means you don' need to initialize it.

receiver/httpreceiver/logs.go
```diff
@@ -0,0 +1,232 @@
...
+	}, nil
+}
+
+// Start calls startListening
+func (r *httpLogsReceiver) Start(ctx context.Context, host component.Host) error {
+	return r.startListening(ctx, host)
+}
+
+// startListening starts serve on the server using TLS depending on receiver configuration
+func (r *httpLogsReceiver) startListening(_ context.Context, host component.Host) error {
```
cpheps, line 64: If we aren't using the context can we remove it from the function signature?

receiver/httpreceiver/logs.go
```diff
@@ -0,0 +1,232 @@
...
+	r.logger.Debug("starting receiver HTTP server")
+	var err error
+	r.server, err = r.serverSettings.ToServer(host, r.telemetrySettings, r)
+	if err != nil {
+		return err
```
cpheps, line 69: Can we add wrapping to all these errors? So this one would be `fmt.Errorf("to server: %w", err)`

receiver/httpreceiver/logs.go
```diff
@@ -0,0 +1,232 @@
...
+	if err := r.server.Shutdown(ctx); err != nil {
+		return err
+	}
+	r.logger.Debug("waiting for shutdown to complete")
+	r.wg.Wait()
+	return nil
+
+}
+
+// handleRequest is the function the server uses for requests; calls ConsumeLogs
```
cpheps, line 129: This comment should start with `ServeHTTP`. 

Should this method be exported?

receiver/httpreceiver/logs.go
```diff
@@ -0,0 +1,232 @@
...
+		r.logger.Debug("received request to path that does not match the configured path", zap.String("request path", req.URL.Path))
+		return
+	}
+
+	// read in request body
+	r.logger.Debug("reading in request body")
+	payload, err := io.ReadAll(req.Body)
+	if err != nil {
+		rw.WriteHeader(http.StatusUnprocessableEntity)
+		r.logger.Debug("failed to read logs payload", zap.Error(err), zap.String("remote", req.RemoteAddr))
```
cpheps, line 143: This should probably be an Error log so customers can see requests are not being processed.

receiver/httpreceiver/logs.go
dpaasman00: Yea, in order to use `ToServer` and `ToListener` with the `HTTPServerSettings`, `ToServer` gets passed a http handler interface with a function called `ServeHTTP`

dpaasman00: At least that's how I was able to get it to work

receiver/httpreceiver/config.go
```diff
@@ -0,0 +1,65 @@
...
+	"net"
+	"path"
+
+	"go.opentelemetry.io/collector/config/confighttp"
+)
+
+// Config defines the configuration for an HTTP receiver
+type Config struct {
+	Path           string                         `mapstructure:"path"`
+	ServerSettings *confighttp.HTTPServerSettings `mapstructure:"server"`
```
BinaryFissionGames, line 29: Small recommendation here:
```suggestion
	ServerSettings *confighttp.HTTPServerSettings `mapstructure:",squash"`
```

This way we don't need the extra `server` key when specifying endpoint/tls. I think this is the usual pattern with this struct (e.g. https://github.com/open-telemetry/opentelemetry-collector/blob/2a9d870a57038cb228ef20e7620f18a93c92b9d3/receiver/otlpreceiver/config.go#L25)

BinaryFissionGames (commented):
Looking good, just one suggestion on the configuration

receiver/httpreceiver/config.go
dpaasman00: oh yea perfect, didn't realize i could do that

schmikei (approved):
Approving since I initially reviewed, I'll defer to the agent team on if/when to merge

PR #1331, opened 2023-11-09
feat: Add Chronicle exporter

Mrod1598:
### Proposed Change

- Add Chronicle exporter

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


exporter/chronicleexporter/config.go
```diff
@@ -0,0 +1,76 @@
...
+
+package chronicleexporter
+
+import (
+	"errors"
+
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+)
+
+var regions = map[string]string{
```
cpheps, line 23: nit: comment please with link to docs.

factories/exporters.go
```diff
@@ -94,4 +95,5 @@ var defaultExporters = []exporter.Factory{
 	splunkhecexporter.NewFactory(),
 	sumologicexporter.NewFactory(),
 	zipkinexporter.NewFactory(),
+	chronicleexporter.NewFactory(),
```
cpheps, line 98: nit: can you put this in alphabetical order?

exporter/chronicleexporter/exporter.go
```diff
@@ -0,0 +1,129 @@
...
+	return &chronicleExporter{
+		endpoint:   buildEndpoint(cfg),
+		cfg:        cfg,
+		logger:     params.Logger,
+		httpClient: httpClient,
+		marshaler:  newMarshaler(*cfg),
+	}, nil
+}
+
+func buildEndpoint(cfg *Config) string {
```
cpheps, line 82: Can you add some comments on this to indicate how this works or point to docs? It sounds like it wasn't easy to track down this answer so we should document what we know.

exporter/chronicleexporter/config.go
```diff
@@ -0,0 +1,78 @@
...
+	CredsFilePath string `mapstructure:"creds_file_path"`
+
+	// Creds are the Google credentials JSON file.
+	Creds string `mapstructure:"creds"`
+
+	// LogType is the type of log that will be sent to Chronicle.
+	LogType string `mapstructure:"log_type"`
+
+	// RawLogField is the field name that will be used to send raw logs to Chronicle.
+	RawLogField string `mapstructure:"raw_log_field"`
```
BinaryFissionGames, line 55: Thoughts on making this an OTTL expression (https://github.com/observIQ/bindplane-agent/blob/f484d280bd5e105de4bde4dbc7e6a284c3b5de32/expr/ottl_expression.go#L71)? Then parsing and evaluating the expression would be easy.
We should also validate this field in Validation.

exporter/chronicleexporter/config.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// CustomerID is the customer ID that will be used to send logs to Chronicle.
+	CustomerID string `mapstructure:"customer_id"`
+}
+
+// Validate checks if the configuration is valid.
+func (cfg *Config) Validate() error {
+	if cfg.CredsFilePath == "" && cfg.Creds == "" {
+		return errors.New("either creds_file_path or creds is required")
+	}
```
BinaryFissionGames, line 65: No support for deafult credentials? I think we should support it, because even if it's more difficult for customers to configure, it's the "recommended" approach by Google and generally more secure.

https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentials

exporter/chronicleexporter/exporter.go
```diff
@@ -0,0 +1,131 @@
...
+
+	resp, err := ce.httpClient.Do(request)
+	if err != nil {
+		return fmt.Errorf("send request to Chronicle: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		respBody := []byte{}
+		_, err := resp.Body.Read(respBody)
```
BinaryFissionGames, line 120: Read() may not read the whole response, you'll need to read until EOF is encountered
```suggestion
		respBody, err := io.ReadAll(resp.Body)
```

exporter/chronicleexporter/factory.go
```diff
@@ -0,0 +1,83 @@
...
+// createLogsExporter creates a new log exporter based on this config.
+func createLogsExporter(
+	ctx context.Context,
+	params exporter.CreateSettings,
+	cfg component.Config,
+) (exporter.Logs, error) {
+	chronicleCfg, ok := cfg.(*Config)
+	if !ok {
+		return nil, consumererror.NewPermanent(errors.New("invalid config type"))
+	}
```
BinaryFissionGames, line 61: nit: No need to wrap this in consumererror, as this is not a "ConsumeXXX" function.

exporter/chronicleexporter/factory.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// Validate the config
+	if err := chronicleCfg.Validate(); err != nil {
+		return nil, err
+	}
```
BinaryFissionGames, line 66: nit: The collector calls validate already on every component in the config, so you could remove this extra call to validate.

exporter/chronicleexporter/marshal.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+}
+
+func (ce *marshaler) MarshalRawLogs(ld plog.Logs) ([]byte, error) {
+	rawLogs, err := ce.extractRawLogs(ld)
+	if err != nil {
+		return nil, fmt.Errorf("extract raw logs: %w", err)
+	}
+
+	rawLogData := map[string]interface{}{
```
BinaryFissionGames, line 52: nit: 
```suggestion
	rawLogData := map[string]any{
```

exporter/chronicleexporter/marshal.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func parseLogField(field string) (string, []string, error) {
+	parts := strings.SplitN(field, `["`, 2)
+
+	if len(parts) == 1 {
+		return parts[0], nil, nil
+	}
+
+	re := regexp.MustCompile(`\["(.*?)"\]`)
```
BinaryFissionGames, line 173: This should be hoisted out of this function so it is compiled only once.

exporter/chronicleexporter/factory.go
```diff
@@ -0,0 +1,83 @@
...
+	"go.opentelemetry.io/collector/exporter"
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+)
+
+// TypeStr for the exporter type.
+const TypeStr = "chronicle"
+
+// Factory is the factory for the Chronicle exporter.
+type Factory struct {
+}
```
BinaryFissionGames, line 33: Are these used? I think `typeStr` comes from the metadata package now, and a factory struct should be unnecessary.

exporter/chronicleexporter/doc.go
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:generate mdatagen metadata.yaml
+
+// Package chronicleexporter stores OpenTelemetry data as a Chronicle exporter.
```
BinaryFissionGames, line 17:
```suggestion
// Package chronicleexporter exports OpenTelemetry data to Chronicle.
```

Mrod1598: Updated.

exporter/chronicleexporter/factory.go
Mrod1598: Updated.

exporter/chronicleexporter/marshal.go
Mrod1598: Updated.

exporter/chronicleexporter/factory.go
Mrod1598: Updated.

exporter/chronicleexporter/exporter.go
Mrod1598: Updated.

exporter/chronicleexporter/config.go
Mrod1598: Updated.

exporter/chronicleexporter/config.go
Mrod1598: Updated for most of it but is there anything I'm missing in the switch case for the types?

exporter/chronicleexporter/marshal.go
```diff
@@ -0,0 +1,173 @@
...
+				return "", fmt.Errorf("marshal body map: %w", err)
+			}
+			return string(bytes), nil
+		default:
+			return "", errors.New("unsupported body type")
+		}
+	default:
+		return "", fmt.Errorf("unsupported top level field: %s", field)
+	}
+}
```
BinaryFissionGames, line 173: Is this still in use?

exporter/chronicleexporter/marshal.go
```diff
@@ -0,0 +1,173 @@
...
+					if err != nil {
+						return nil, fmt.Errorf("marshal log record: %w", err)
+					}
+
+					rawLog = string(bytesLogRecord)
+				} else {
+					rawLog, err = m.getRawField(ctx, logRecord, scopeLog.Scope(), resourceLog.Resource())
+					if err != nil {
+						return nil, fmt.Errorf("get raw field: %w", err)
+					}
```
BinaryFissionGames, line 103: Already discussed this, but just leaving it as a note; We should just log and continue here if the log record's field can't be retrieved for some reason.

exporter/chronicleexporter/config.go
BinaryFissionGames: Discussed this offline, the current cases for the switch seems fine, we can add more in later if needed.

cpheps (approved):

BinaryFissionGames (changes requested):

exporter/chronicleexporter/marshal.go
Mrod1598: nope

BinaryFissionGames (approved):

PR #1346, opened 2023-11-15
fix: Put Chronicle into its own module

Mrod1598:
### Proposed Change

- Put Chronicle into its own module
- Also Remove Retry and Timeout Settings from doc since we don't show it in the rest of the exporters 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


go.mod
```diff
@@ -6,9 +6,9 @@ require (
 	github.com/google/uuid v1.4.0
 	github.com/mholt/archiver/v3 v3.5.1
 	github.com/observiq/bindplane-agent/exporter/azureblobexporter v1.40.0
+	github.com/observiq/bindplane-agent/exporter/chronicleexporter v0.0.0-00010101000000-000000000000
```
cpheps, line 9: set this to v1.40.0 instead of the all zero value.

cpheps (changes requested):

cpheps (approved):

BinaryFissionGames (approved):

PR #1347, opened 2023-11-16
feat(service): SysV init.d style startup script

Dylan-M:
### Proposed Change
This startup script will be needed to support any Linux distros that aren't using systemd. This is mostly older distros, but some modern tiny distros still use init.d scripts as well.

Known issues/concerns to be addressed in a future PR:
- error.log doesn't get rotated (still researching best way to do this)
- Has only been tested on EL6. Needs to be tested on SLES11, and a modern tiny distro as well.

- [X] Changes are tested
- [ ] CI has passed

### DIFF
--- a/service/observiq-otel-collector
+++ b/service/observiq-otel-collector
@@ -36,9 +36,9 @@ start() {
   else
     echo " * Starting $PROGRAM"
     if [ $DEBUG ]; then
-    COMMAND=/opt/observiq-otel-collector/observiq-otel-collector --config config.yaml" 1> /dev/null 2>> $OIQ_OTEL_COLLECTOR_HOME/log/error.log
+      COMMAND='"/opt/observiq-otel-collector/observiq-otel-collector --config config.yaml" 1> /dev/null 2>> $OIQ_OTEL_COLLECTOR_HOME/log/error.log'
     else
-    COMMAND=/opt/observiq-otel-collector/observiq-otel-collector --config config.yaml" 1>> $OIQ_OTEL_COLLECTOR_HOME/log/error.log 2>&1
+      COMMAND='"/opt/observiq-otel-collector/observiq-otel-collector --config config.yaml" 1>> $OIQ_OTEL_COLLECTOR_HOME/log/error.log 2>&1'
     fi
     export OIQ_OTEL_COLLECTOR_HOME
     export OIQ_OTEL_COLLECTOR_STORAGE
--- a/service/observiq-otel-collector
+++ b/service/observiq-otel-collector
@@ -27,7 +27,7 @@ HTTP_PROXY=
 HTTPS_PROXY=
 # set this to a user such as observiq-otel-collector, to run as that user
 # observiq-otel-collector user is created during installation
-# leave this variable empty to run as root
+# leave this variable empty to run as root (Default behavior)
 RUNUSER=
 PIDFILE=/var/run/observiq-otel-collector.pid
 DEBUG=false
--- a/service/observiq-otel-collector
+++ b/service/observiq-otel-collector
@@ -25,7 +25,9 @@ OIQ_OTEL_COLLECTOR_HOME=/opt/observiq-otel-collector
 OIQ_OTEL_COLLECTOR_STORAGE=/opt/observiq-otel-collector/storage
 HTTP_PROXY=
 HTTPS_PROXY=
-USER=root
+# set this to a user such as observiq-otel-collector, to run as that user
+# observiq-otel-collector user is created during installation
+USER=
 PIDFILE=/var/run/observiq-otel-collector.pid
 DEBUG=false
 RETVAL=0
--- a/service/observiq-otel-collector
+++ b/service/observiq-otel-collector
@@ -48,7 +48,7 @@ start() {
     export HTTP_PROXY
     export HTTPS_PROXY
     if [ -n "$RUNUSER" ]; then
-      su -p $RUNUSER -c "$COMMAND" &
+      su -p "$RUNUSER" -c "$COMMAND" &
     else
       nohup "$COMMAND" &
     fi
--- a/service/observiq-otel-collector
+++ b/service/observiq-otel-collector
@@ -25,6 +25,7 @@ OIQ_OTEL_COLLECTOR_HOME=/opt/observiq-otel-collector
 OIQ_OTEL_COLLECTOR_STORAGE=/opt/observiq-otel-collector/storage
 HTTP_PROXY=
 HTTPS_PROXY=
+USER=root
 PIDFILE=/var/run/observiq-otel-collector.pid
 DEBUG=false
 RETVAL=0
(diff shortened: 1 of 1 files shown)


service/observiq-otel-collector
```diff
@@ -0,0 +1,114 @@
...
+  . /etc/init.d/functions
+fi
+
+if [ -e /lib/lsb/init-functions ]; then
+  PROC=true
+  . /lib/lsb/init-functions
+fi
+
+# Pull in sysconfig settings
+[ -f /etc/sysconfig/observiq-otel-collector ] && . /etc/sysconfig/observiq-otel-collector
```
jsirianni, line 20: What would this file be used for if it existed, environment variables to configure the collector?

service/observiq-otel-collector
```diff
@@ -0,0 +1,114 @@
...
+    if [ $DEBUG ]; then
+      COMMAND='"/opt/observiq-otel-collector/observiq-otel-collector --config config.yaml" 1> /dev/null 2>> $OIQ_OTEL_COLLECTOR_HOME/log/error.log'
+    else
+      COMMAND='"/opt/observiq-otel-collector/observiq-otel-collector --config config.yaml" 1>> $OIQ_OTEL_COLLECTOR_HOME/log/error.log 2>&1'
+    fi
+    export OIQ_OTEL_COLLECTOR_HOME
+    export OIQ_OTEL_COLLECTOR_STORAGE
+    export HTTP_PROXY
+    export HTTPS_PROXY
+    if [ -n "root" ]; then
```
jsirianni, line 47: Do we need this condition? Is there a scenario where the service would not be executed as the root user?

Also, how do we support running the collector as the `observiq-otel-collector` user (which is created and managed by the linux packages)? This does not need to be solved right away, just curious as we do have a solution for systemd.

```
    if [ -n "root" ]; then
             ^--^ SC2157: Argument to -n is always true due to literal strings.
```

jsirianni (changes requested):
Can you run [shellcheck](https://github.com/koalaman/shellcheck) and resolve any findings?
```
shellcheck --exclude=SC1091 service/observiq-otel-collector
```

For false positives, add a comment above the offender like this:
```bash
# shellcheck disable=SC2039
```

service/observiq-otel-collector
Dylan-M: For more information, this doc is good. Currently, we probably wouldn't be using this. However, in theory all the environment variables could be moved to this file and used that way.

https://docs.oracle.com/en/operating-systems/oracle-linux/6/admin/ol_about_kernparm.html

service/observiq-otel-collector
Dylan-M: I've addressed all shellcheck feedback (interesting some of them, since I copied this from an existing RHEL created script).

Additionally added the support to specify a user in place of the check you asked about, which was a placeholder I had forgotten.

service/observiq-otel-collector
```diff
@@ -0,0 +1,121 @@
...
+    if [ $DEBUG ]; then
+      COMMAND="cd /opt/observiq-otel-collector/ && nohup /opt/observiq-otel-collector/observiq-otel-collector --config config.yaml 1> /dev/null 2>> $OIQ_OTEL_COLLECTOR_HOME/log/error.log"
+    else
+      COMMAND="cd /opt/observiq-otel-collector/ && nohup /opt/observiq-otel-collector/observiq-otel-collector --config config.yaml 1>> $OIQ_OTEL_COLLECTOR_HOME/log/error.log 2>&1"
+    fi
+    export OIQ_OTEL_COLLECTOR_HOME
+    export OIQ_OTEL_COLLECTOR_STORAGE
+    export HTTP_PROXY
+    export HTTPS_PROXY
+    if [ -n "$USER" ]; then
```
jsirianni, line 48: I think this just checks if $USER is set, but what we really want is to check if USER is not "root". If the user configures the service to run as `observiq-otel-collector`, this condition will run the collector as the the non root user.

```suggestion
    if [ "$USER" != "root" ]; then
```

Dylan-M: Actually, my mistake is setting it at all. Leaving it empty runs as root. I'll be submitting another commit shortly that corrects that, and adds a comment about it.

jsirianni: Can we use a different variable name? In my env, $USER is set. It may not be the case here, but maybe to be safe. Something like `PROCESS_USER`?

service/observiq-otel-collector
```diff
@@ -0,0 +1,123 @@
...
+
+PROGRAM=observiq-otel-collector
+LOCKFILE=/var/lock/$PROGRAM
+OIQ_OTEL_COLLECTOR_HOME=/opt/observiq-otel-collector
+OIQ_OTEL_COLLECTOR_STORAGE=/opt/observiq-otel-collector/storage
+HTTP_PROXY=
+HTTPS_PROXY=
+# set this to a user such as observiq-otel-collector, to run as that user
+# observiq-otel-collector user is created during installation
+RUNUSER=
```
cpheps, line 30: Should we set this to `observiq-otel-collector` instead of leaving it blank?

service/observiq-otel-collector
```diff
@@ -0,0 +1,123 @@
...
+RETVAL=0
+start() {
+  if [ -f $PIDFILE ]; then
+    PID=$(cat $PIDFILE)
+    echo " * $PROGRAM already running: $PID"
+    RETVAL=2
+  else
+    echo " * Starting $PROGRAM"
+    if [ $DEBUG ]; then
+      COMMAND="cd /opt/observiq-otel-collector/ && nohup /opt/observiq-otel-collector/observiq-otel-collector --config config.yaml 1> /dev/null 2>> $OIQ_OTEL_COLLECTOR_HOME/log/error.log"
```
cpheps, line 42: I'm not as familiar with init.d do we need `nohup` here if we're using the `observiq-otel-collector` user?

service/observiq-otel-collector
```diff
@@ -0,0 +1,123 @@
...
+      COMMAND="cd /opt/observiq-otel-collector/ && nohup /opt/observiq-otel-collector/observiq-otel-collector --config config.yaml 1>> $OIQ_OTEL_COLLECTOR_HOME/log/error.log 2>&1"
+    fi
+    export OIQ_OTEL_COLLECTOR_HOME
+    export OIQ_OTEL_COLLECTOR_STORAGE
+    export HTTP_PROXY
+    export HTTPS_PROXY
+    if [ -n "$RUNUSER" ]; then
+      su -p "$RUNUSER" -c "$COMMAND" &
+    else
+      nohup "$COMMAND" &
```
cpheps, line 53: The `nohup` here means we're running a nested `nohup` right? Are there any side-effects to that?

service/observiq-otel-collector
```diff
@@ -0,0 +1,123 @@
...
+    RETVAL=$?
+    [ "$RETVAL" -eq 0 ] && touch $LOCKFILE
+  fi
+}
+
+stop() {
+  if [ -f $PIDFILE ]; then
+    PID=$(cat $PIDFILE);
+    printf " * Stopping %s... " "$PROGRAM"
+    kill "$PID" > /dev/null 2>&1
```
cpheps, line 67: In the systemd file we give the service 20 seconds to shut down to avoid it forever hanging. Is there any concern here of no having the collector shutdown within 20 seconds?

service/observiq-otel-collector
```diff
@@ -0,0 +1,123 @@
...
+    echo "stopped"
+    rm $PIDFILE && rm -f $LOCKFILE
+    RETVAL=0
+  else
+    echo " * $PROGRAM is not running"
```
cpheps, line 72: where do echos go from init.d files?

cpheps (changes requested):
I'll leave the expertise for this up to @jsirianni but I do have some questions on how things work verse systemd just for my own understanding.

service/observiq-otel-collector
jsirianni: It will default to `root` if `RUNUSER` is unset, which is what we want. It feels clunky, but if you look at the `if` condition, it will use `su` to run the process as `$RUNUSER`, if set.

Dylan-M: Depends. Do we want to run as that user by default, or do we want to run as root by default?

service/observiq-otel-collector
Dylan-M: Still need nohup. All nohup does is keep the process running after the temporary shell created during script execution exits.

service/observiq-otel-collector
Dylan-M: Negative. The $COMMAND variable contains a chain of commands executed in turn. The way this will expand is:
`nohup cd /opt/observiq-otel-collector/ && nohup /opt/observiq-otel-collector/observiq-otel-collector --config config.yaml`

The double ampersand meaning, if command on left succeeds, run command on the right.

service/observiq-otel-collector
Dylan-M: init.d startup scripts don't have a concept of timeouts. Generally on init.d systems, if the service fails to stop it is up to the user to manually clean it up. One of the biggest advantages of systemd.

service/observiq-otel-collector
Dylan-M: This is the output to the user who ran the service command.

service/observiq-otel-collector
jsirianni: Root by default would match systemd behavior

cpheps (approved):

PR #1359, opened 2023-11-22
feat: Override chronicle log type

Mrod1598:
### Proposed Change

- Rework exporter to send multiple payloads organized by log_type.
     - had to do this  because you can only specify one log_type per payload
- Check for log type on attributes. if it exists and maps to a known chronicle log_type, use that value instead. 

## Testing
The easiest way I've been able to test this is to send two types of logs to chronicle. win event logs and another one.
and set the default log_type to any type other than win event and the exporter will overwrite it for the windows logs.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


exporter/chronicleexporter/marshal.go
```diff
@@ -28,83 +28,71 @@ import (
...
-						continue
-					}
+				if rawLog == "" {
+					continue
 				}
 
-				entries = append(entries, entry{
+				entries[logType] = append(entries[logType], entry{
 					LogText:   rawLog,
 					Timestamp: logRecord.Timestamp().AsTime().Format(time.RFC3339Nano),
```
BinaryFissionGames, line 97: This is tangential to this PR, but we should fall back to ObservedTimestamp if the timestamp is unset (zero value)

BinaryFissionGames (changes requested):

exporter/chronicleexporter/config.go
```diff
@@ -55,6 +55,9 @@ type Config struct {
 	// LogType is the type of log that will be sent to Chronicle.
 	LogType string `mapstructure:"log_type"`
 
+	// OverrideLogType is the field name that will be used to override the log type.
+	OverrideLogType bool `mapstructure:"override_log_type"`
```
BinaryFissionGames, line 59: Comment indicates this is a field name, but this is a bool

BinaryFissionGames: Could we also add this field to the README's table?

exporter/chronicleexporter/marshal.go
```diff
@@ -28,85 +28,80 @@ import (
 	"go.uber.org/zap"
 )
 
-// marshaler is an interface for marshalling logs.
-//
-//go:generate mockery --name logMarshaler --output ./internal/mocks --with-expecter --filename mock_marshaler.go --structname MockMarshaler
-type logMarshaler interface {
-	MarshalRawLogs(ctx context.Context, ld plog.Logs) ([]byte, error)
+const logTypeField = `attributes["log_type"]`
```
BinaryFissionGames, line 31: We should document that this field is read to get the log type too.

BinaryFissionGames (changes requested):

exporter/chronicleexporter/config.go
Mrod1598: updated comment and added to readme, was originally a field but we decided to just have a set val

BinaryFissionGames (approved):

cpheps (approved):

PR #1369, opened 2023-12-01
fix(m365 receiver): Add nil check to shutdown

dpaasman00:
### Proposed Change
The collector panicked with a nil pointer dereference if it was unable to create an authorization token when first started. This occurred because the attempt to create the token happens inside `Start()` before a cancel func is assigned to the recv. The collector would error out of `Start()` and then call `Shutdown()` and encounter the nil pointer dereference when trying to call `cancel()`. This PR adds a nil check on the cancel funcs before they are called, allowing the collector to fail correctly and not panic with a nil ptr dereference. 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/m365receiver/logs.go
```diff
@@ -136,7 +136,9 @@ func (l *m365LogsReceiver) Start(ctx context.Context, host component.Host) error
 
 func (l *m365LogsReceiver) Shutdown(ctx context.Context) error {
 	l.logger.Debug("shutting down logs receiver")
-	l.cancel()
+	if l.cancel != nil {
+		l.cancel()
+	}
 	l.wg.Wait()
 	return l.checkpoint(ctx)
```
BinaryFissionGames, line 143: In `checkpoint`, could we add a check for if `l.record` is nil? It wasn't obvious to me how we avoid clobbering the old checkpoint with the nil record in this scenario.

BinaryFissionGames (approved):

Mrod1598 (approved):

PR #1380, opened 2023-12-14
fix: Replace region with endpoint

Mrod1598:
### Proposed Change
- Remove regional endpoints and replace them with just an endpoint field. 
- link out to the endpoints in the creds section.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/chronicleexporter/README.md
+++ b/exporter/chronicleexporter/README.md
@@ -14,26 +14,22 @@ This exporter facilitates the sending of logs to Chronicle, which is a security 
 
 1. The exporter uses the configured credentials to authenticate with the Google Cloud services.
 2. It marshals logs into the format expected by Chronicle.
-3. It sends the logs to the appropriate regional Chronicle endpoint.
+3. It sends the logs to the appropriate Chronicle endpoint.
 
 ## Configuration
 
 The exporter can be configured using the following fields:
 
-| Field               | Type   | Default | Required | Description                                                                                                                                                           |
-| ------------------- | ------ | ------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `region`            | string |         | `false`  | The region where the data will be sent, it must be one of the predefined regions. if no region is specfied defaults to `https://malachiteingestion-pa.googleapis.com` |
-| `creds_file_path`   | string |         | `true`   | The file path to the Google credentials JSON file.                                                                                                                    |
-| `creds`             | string |         | `true`   | The Google credentials JSON.                                                                                                                                          |
-| `log_type`          | string |         | `true`   | The type of log that will be sent.                                                                                                                                    |
-| `raw_log_field`     | string |         | `false`  | The field name for raw logs.                                                                                                                                          |
-| `customer_id`       | string |         | `false`  | The customer ID used for sending logs.                                                                                                                                |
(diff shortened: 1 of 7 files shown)


exporter/chronicleexporter/README.md
```diff
@@ -14,57 +14,39 @@ This exporter facilitates the sending of logs to Chronicle, which is a security
...
+Besides, the default endpoint, there are also regional endpoints that can be used [here](https://cloud.google.com/chronicle/docs/reference/ingestion-api#regional_endpoints)
+.
 
 For additional information on accessing Chronicle, see the [Chronicle documentation](https://cloud.google.com/chronicle/docs/reference/ingestion-api#getting_api_authentication_credentials).
 
 ## Example Configuration
 
 ```yaml
 chronicle:
-  region: "Europe Multi-Region"
```
cpheps, line 67: Do we need an endpoint field in this example?

exporter/chronicleexporter/config.go
```diff
@@ -78,12 +65,6 @@ func (cfg *Config) Validate() error {
 		return errors.New("log_type is required")
 	}
 
-	if cfg.Region != "" {
```
cpheps, line 81: Do we need to check for empty endpoint here.

cpheps (changes requested):

exporter/chronicleexporter/config.go
Mrod1598: I figured no because we have a default endpoint that'd get populated

exporter/chronicleexporter/README.md
Mrod1598: I figured no because we have a default endpoint that'd get populated

exporter/chronicleexporter/config.go
cpheps: I don't know if a user puts a blank string in if that overrides it. If it doesn't then this is fine.

exporter/chronicleexporter/README.md
cpheps: Can we should an example with an endpoint?

cpheps (approved):

exporter/chronicleexporter/config.go
Mrod1598: I'll test

exporter/chronicleexporter/README.md
```diff
@@ -14,58 +14,52 @@ This exporter facilitates the sending of logs to Chronicle, which is a security
...
 ### Log Type
 
 If the `attributes["log_type"]` field is present in the log, and maps to a known Chronicle `log_type` the exporter will use the value of that field as the log type. If the `attributes["log_type"]` field is not present, the exporter will use the value of the `log_type` configuration field as the log type.
 
 ## Credentials
 
-This exporter requires a Google Cloud service account with access to the Chronicle API. The service account must have access to the following endpoint(s):
+This exporter requires a Google Cloud service account with access to the Chronicle API. The service account must have access to the endpoint specfied in the config.
+Besides, the default endpoint, there are also regional endpoints that can be used [here](https://cloud.google.com/chronicle/docs/reference/ingestion-api#regional_endpoints)
+.
```
BinaryFissionGames, line 42:
```suggestion
Besides the default endpoint, there are also regional endpoints that can be used [here](https://cloud.google.com/chronicle/docs/reference/ingestion-api#regional_endpoints).
```

BinaryFissionGames (changes requested):

BinaryFissionGames (approved):

PR #1382, opened 2023-12-14
feat: Add Chronicle Forwarder

Mrod1598:
### Proposed Change

- The chronicle forwarder requires the logs to be in an unparsed format to be able to ingest and parse the logs correctly, this is essentially a way to export a specific field out of the OTLP format to make it easier for Chronicle to ingest it properly.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


exporter/chronicleforwarderexporter/README.md
```diff
@@ -0,0 +1,55 @@
...
+| syslog.tls.key_file  | string       |               | `false`  | Configure the receiver to use TLS.                |
+| syslog.tls.cert_file | string       |               | `false`  | Configure the receiver to use TLS.                |
+| file.path            | string       |               | `false`  | The path to the file for storing logs.            |
+
+## Example Configurations
+
+### Syslog Configuration Example
+
+```yaml
+chronicleforwarderexporter:
```
BinaryFissionGames, line 38: This should just be `chronicleforwarder` in both examples, I think

exporter/chronicleforwarderexporter/config.go
```diff
@@ -0,0 +1,101 @@
...
+type SyslogConfig struct {
+	// Host is the Chronicle forwarder endpoint to send logs to.
+	Host string `mapstructure:"host"`
+
+	// port is the port to send logs to.
+	Port int `mapstructure:"port"`
+
+	// Network is the network protocol to use.
+	Network string `mapstructure:"network"`
+
```
BinaryFissionGames, line 64: I think we should use confignet.NetAddr for these fields
https://pkg.go.dev/go.opentelemetry.io/collector/config/confignet#NetAddr

exporter/chronicleforwarderexporter/config.go
```diff
@@ -0,0 +1,101 @@
...
+	confighttp.HTTPServerSettings `mapstructure:",squash"`
```
BinaryFissionGames, line 65: I think we're only using this for TLS, this isn't an HTTP server.

We should use 
https://pkg.go.dev/go.opentelemetry.io/collector/config/configtls#TLSClientSetting 
here if I understand correctly.

exporter/chronicleforwarderexporter/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+		msg = fmt.Sprintf("%s%s", msg, "\n")
+	}
+	_, err := fmt.Fprint(ce.writer, msg)
+	return err
+}
+
+// Shutdown stops the exporter and is invoked during shutdown.
+func (ce *chronicleForwarderExporter) Shutdown(_ context.Context) error {
+	if ce.writer != nil {
+		if err := ce.writer.(io.Closer).Close(); err != nil {
```
BinaryFissionGames, line 113: Any reason not to make `ce.writer` an [io.WriteCloser](https://pkg.go.dev/io#WriteCloser)?

exporter/chronicleforwarderexporter/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return nil
+}
+
+func (ce *chronicleForwarderExporter) send(msg string) error {
+	if !strings.HasSuffix(msg, "\n") {
+		msg = fmt.Sprintf("%s%s", msg, "\n")
+	}
+	_, err := fmt.Fprint(ce.writer, msg)
```
BinaryFissionGames, line 106: Could we just use the `ce.writer.Write()` here instead?

exporter/chronicleforwarderexporter/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+		var err error
+		if cfg.Syslog.TLSSetting != nil {
+			tlsConfig, err := cfg.Syslog.TLSSetting.LoadTLSConfig()
+			if err != nil {
+				return nil, fmt.Errorf("load TLS config: %w", err)
+			}
+			conn, err = tls.Dial(cfg.Syslog.Network, endpoint, tlsConfig)
+		} else {
+			conn, err = net.Dial(cfg.Syslog.Network, endpoint)
+		}
```
BinaryFissionGames, line 61: This seems a little brittle to me, what happens if the connection is broken for a second, but could be re-established later? I think because we only dial once, any network failure would end up being permanent until the collector restarts.

exporter/chronicleforwarderexporter/exporter_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+// MockWriter is a mock implementation of io.Writer.
+type MockWriter struct {
+	mock.Mock
+}
+
+func (m *MockWriter) Write(p []byte) (n int, err error) {
+	args := m.Called(p)
+	return args.Int(0), args.Error(1)
+}
```
BinaryFissionGames, line 40: nit: You could probably get away with using bytes.Buffer here instead of a mock.

BinaryFissionGames (changes requested):

exporter/chronicleforwarderexporter/exporter.go
Mrod1598: would it be better to dial on every log push?

BinaryFissionGames: Yeah, it looks like that's what the upstream syslog exporter does
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c49dbc1e3f7189abb22e64fc26b72b0c896d1385/exporter/syslogexporter/exporter.go#L98

exporter/chronicleforwarderexporter/exporter.go
BinaryFissionGames: Probably actually makes more sense to use [io.WriteString](https://pkg.go.dev/io#WriteString)

Mrod1598: updated.

exporter/chronicleforwarderexporter/exporter.go
Mrod1598: reworked it to dial during logs push

exporter/chronicleforwarderexporter/exporter_test.go
Mrod1598: removed.

exporter/chronicleforwarderexporter/exporter.go
Mrod1598: updated. didn't even notice that was a thing

exporter/chronicleforwarderexporter/config.go
Mrod1598: updated.

exporter/chronicleforwarderexporter/config.go
Mrod1598: updated

exporter/chronicleforwarderexporter/README.md
Mrod1598: whoops updated.

exporter/chronicleforwarderexporter/README.md
```diff
@@ -0,0 +1,55 @@
+# Chronicle Forwarder Exporter
+
+The Chronicle Forwarder Exporter is designed for forwarding logs to a Chronicle endpoint using either Syslog or File-based methods. This exporter supports customization of data export types and various configuration options to tailor the connection and data handling to specific needs.
```
cpheps, line 3: Does it forward logs to a chronicle endpoint or an existing chronicle forwarder? Maybe this is right the wording just does not match my current understanding.

exporter/chronicleforwarderexporter/README.md
```diff
@@ -0,0 +1,55 @@
...
+1. For Syslog, it establishes a network connection to the specified Chronicle forwarder endpoint.
+2. For File, it writes logs to a specified file path.
+
+## Configuration
+
+| Field                | Type         | Default Value | Required | Description                                       |
+| -------------------- | ------------ | ------------- | -------- | ------------------------------------------------- |
+| export_type          | string       | `syslog`      | `true`   | Type of export, either `syslog` or `file`.        |
+| syslog               | SyslogConfig |               | `true`   | Configuration for Syslog connection.              |
+| file                 | File         |               | `true`   | Configuration for File connection.                |
```
cpheps, line 24: I think it might make sense to break `SyslogConfig` and `File` out into their own sub tables.

Also, should `File` be `FileConfig` to match the wording of `SyslogConfig`?

exporter/chronicleforwarderexporter/config.go
```diff
@@ -0,0 +1,96 @@
...
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+	"go.uber.org/zap"
+)
+
+const (
+	// ExportTypeSyslog is the syslog export type.
+	ExportTypeSyslog = "syslog"
+
+	// ExportTypeFile is the file export type.
+	ExportTypeFile = "file"
```
cpheps, line 33: nit: These don't need to be exported.

exporter/chronicleforwarderexporter/config.go
```diff
@@ -0,0 +1,96 @@
...
+		if cfg.Syslog.NetAddr.Endpoint == "" {
+			return fmt.Errorf("incomplete syslog configuration: endpoint is required")
+		}
+	}
+
+	if cfg.ExportType == ExportTypeFile {
+		if cfg.File.Path == "" {
+			return fmt.Errorf("file path is required for file export type")
+		}
+	}
```
cpheps, line 85: Thoughts on each config type have a `.Validate()` function?

This if statement could also be changed to a switch something like this:

```go
switch cfg.ExportType {
    case ExportTypeSyslog:
        return cfg.Syslog.Validate()
    case ExportTypeFile:
        return cfg.File.Validate()
    default:
        return errors.New("export_type must be either 'syslog' or 'file'")
}
```

exporter/chronicleforwarderexporter/config.go
```diff
@@ -0,0 +1,96 @@
...
+// File defines configuration for sending to.
+type File struct {
+	// Path is the path to the file to send to Chronicle.
+	Path string `mapstructure:"path"`
+}
+
+// Validate validates the Chronicle exporter configuration.
+func (cfg *Config) Validate() error {
+	if cfg.ExportType != ExportTypeSyslog && cfg.ExportType != ExportTypeFile {
+		return fmt.Errorf("export_type must be either 'syslog' or 'file'")
```
cpheps, line 72: There are a few of these but `fmt.Errorf` should only be used if there is string interpolation. If it's just a raw string use `errors.New`

exporter/chronicleforwarderexporter/exporter.go
```diff
@@ -0,0 +1,121 @@
...
+		cfg:       cfg,
+		logger:    params.Logger,
+		marshaler: newMarshaler(*cfg, params.TelemetrySettings),
+	}, nil
+}
+
+func (ce *chronicleForwarderExporter) Capabilities() consumer.Capabilities {
+	return consumer.Capabilities{MutatesData: false}
+}
+func (ce *chronicleForwarderExporter) logsDataPusher(ctx context.Context, ld plog.Logs) error {
```
cpheps, line 51: nit: newline between these functions.

exporter/chronicleforwarderexporter/exporter.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+			conn, err = tls.Dial(ce.cfg.Syslog.NetAddr.Transport, ce.cfg.Syslog.NetAddr.Endpoint, tlsConfig)
+		} else {
+			conn, err = net.Dial(ce.cfg.Syslog.NetAddr.Transport, ce.cfg.Syslog.NetAddr.Endpoint)
+		}
+
+		if err != nil {
+			return fmt.Errorf("dial: %w", err)
+		}
+		ce.writer = conn
```
cpheps, line 90: Thoughts on breaking these out into two functions? `openSyslogWriter` and `openFileWriter`

exporter/chronicleforwarderexporter/exporter_test.go
```diff
@@ -0,0 +1,130 @@
...
+	go func() {
+		for {
+			conn, err := ln.Accept()
+			if err != nil {
+				log.Println("Error accepting connection:", err)
+				return
+			}
+			go handleSyslogConnection(t, conn, logReceived)
+		}
+	}()
```
cpheps, line 85: Does something make sure these goroutines are done running when this test ends?

exporter/chronicleforwarderexporter/go.mod
```diff
@@ -0,0 +1,65 @@
+module github.com/observiq/bindplane-agent/exporter/chronicleforwarderexporter
+
+go 1.20
+
+require (
+	github.com/observiq/bindplane-agent/expr v1.41.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl v0.91.0
+	github.com/stretchr/testify v1.8.4
+	go.opentelemetry.io/collector/component v0.91.0
+	go.opentelemetry.io/collector/config/configtls v0.90.0
```
cpheps, line 10: Should this be v0.91.0

exporter/chronicleforwarderexporter/README.md
```diff
@@ -0,0 +1,55 @@
+# Chronicle Forwarder Exporter
```
cpheps, line 1: Make sure this is linked to in `docs/exporters.md`

cpheps (changes requested):

exporter/chronicleforwarderexporter/README.md
Mrod1598: linked

exporter/chronicleforwarderexporter/exporter_test.go
Mrod1598: Added a check for a canceled context.

exporter/chronicleforwarderexporter/go.mod
Mrod1598: Updated

exporter/chronicleforwarderexporter/exporter.go
Mrod1598: Sure, broke it up

exporter/chronicleforwarderexporter/config.go
Mrod1598: yeah, bad habit, updated.

exporter/chronicleforwarderexporter/config.go
Mrod1598: Updated

exporter/chronicleforwarderexporter/config.go
Mrod1598: updated

exporter/chronicleforwarderexporter/README.md
Mrod1598: yeah i put `chronicle forwarder endpoint` below this, updated this to match that

exporter/chronicleforwarderexporter/README.md
Mrod1598: That actually should just be removed. below that I break down both of those in the same style as like the collector-contrib docs: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/filelogreceiver/README.md .
does that work for you?

exporter/chronicleforwarderexporter/exporter.go
Mrod1598: Updated

exporter/chronicleforwarderexporter/README.md
cpheps: Looks good.

cpheps (approved):

exporter/chronicleforwarderexporter/exporter.go
```diff
@@ -0,0 +1,128 @@
...
+func (ce *chronicleForwarderExporter) openWriter() error {
+	switch ce.cfg.ExportType {
+	case exportTypeSyslog:
+		return ce.openSyslogWriter()
+	case exportTypeFile:
+		return ce.openFileWriter()
+	default:
+		return errors.New("unsupported export type")
+	}
+}
```
BinaryFissionGames, line 83: I think this method should return `(io.WriteCloser, error)`, and I think the writer shouldn't be a part of the struct, since it only lives for the duration of the `logsDataPusher` (it is closed and unusable after that method)

BinaryFissionGames (changes requested):

exporter/chronicleforwarderexporter/exporter.go
```diff
@@ -0,0 +1,121 @@
...
+	return err
+}
+
+// Shutdown stops the exporter and is invoked during shutdown.
+func (ce *chronicleForwarderExporter) Shutdown(_ context.Context) error {
+	if ce.writer != nil {
+		if err := ce.writer.Close(); err != nil {
+			return fmt.Errorf("close writer: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
Mrod1598: Updated

exporter/chronicleforwarderexporter/exporter.go
Mrod1598: Updated

BinaryFissionGames (approved):

PR #1383, opened 2023-12-14
feat(ppc-linux): Add PowerPC architectures for Linux OS Releases

Dylan-M:
### Proposed Change
This PR will introduce PowerPC (ppc64 and ppc64le) Linux build and release artifacts. These have been tested on an Enterprise Linux ppc64le system.

Access for testing can be provided. The test system is running on SiteOx, and we rented it only for a month.

##### Checklist
- [X] Changes are tested
- [x] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -21,15 +21,25 @@ builds:
       - arm64
       - arm
       - "386"
+      - ppc64
+      - ppc64le
     ignore:
       - goos: windows
         goarch: arm
       - goos: windows
         goarch: arm64
+      - goos: windows
+        goarch: ppc64
+      - goos: windows
+        goarch: ppc64le
       - goos: linux
         goarch: "386"
       - goos: darwin
         goarch: "386"
+      - goos: darwin
+        goarch: ppc64
+      - goos: darwin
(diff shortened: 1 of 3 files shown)


cpheps (commented):
This does build. Are we expecting PowerPC builds to work with BPOP and does upgrading them work? I see we're creating a PowerPC updater too.

Dylan-M:
> This does build. Are we expecting PowerPC builds to work with BPOP and does upgrading them work? I see we're creating a PowerPC updater too.

That is correct. The updater should work, since it is all just using the existing Linux code and services.

Dylan-M:
> Looks good. We'll need to do some stuff on the BPOP side to support upgrades, but I think this should otherwise just work for the initial install.

Cool. Is there a specific branch I should be targeting? This branch being in merge freeze.

BinaryFissionGames:
> Cool. Is there a specific branch I should be targeting? This branch being in merge freeze.

You can point this to the `release/v1.44.0` branch, I've just created it. That branch will be rebased onto main after we release, so if you run into trouble rebasing after release of v1.43.0, let me know.

cpheps (approved):

PR #1384, opened 2023-12-14
feat(install_scripts): Check BindPlane URL before installing

Dylan-M:
### Proposed Change
Add functionality to the install scripts to have it check the provided endpoint for connectivity prior to installation, and offer the user the ability to opt out.

This is feature gated by a command line flag (-c|--check-bp-url) to preserve backwards compatibility.

After this gets released, the BindPlane "Install Agent" should have the `-c` flag added

##### Checklist
- [X] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/install/install_macos.sh
+++ b/scripts/install/install_macos.sh
@@ -286,7 +286,7 @@ connection_check()
       HTTP_ENDPOINT="$(echo "${opamp_endpoint}" | sed 's#^ws#http#' | sed 's#/v1/opamp$##')"
       info "Testing connection to BindPlane: $fg_magenta$HTTP_ENDPOINT$reset..."
 
-      if curl -s "${HTTP_ENDPOINT}" > /dev/null; then
+      if curl --max-time 20 -s "${HTTP_ENDPOINT}" > /dev/null; then
         succeeded
       else
         failed
--- a/scripts/install/install_macos.sh
+++ b/scripts/install/install_macos.sh
@@ -73,7 +73,7 @@ printf() {
 }
 
 increase_indent() { indent="$INDENT_WIDTH$indent" ; }
-decrease_indent() { indent="${indent#*$INDENT_WIDTH}" ; }
+decrease_indent() { indent="${indent#*"$INDENT_WIDTH"}" ; }
 
 # Color functions reset only when given an argument
 bold() { command printf "$bold$*$(if [ -n "$1" ]; then command printf "$reset"; fi)" ; }
@@ -281,19 +281,18 @@ check_prereqs()
 # Test connection to BindPlane if it was specified
 connection_check()
 {
-  info "Help Me Obi-Wan Kenobi!"
-  if [ $check_bp_url == "true" ] ; then
+  if [ "$check_bp_url" = "true" ] ; then
     if [ -n "$opamp_endpoint" ]; then
-      HTTP_ENDPOINT="$(printf "${opamp_endpoint}" | sed 's#ws#http#' | sed 's#/v1/opamp$##')"
+      HTTP_ENDPOINT="$(echo "${opamp_endpoint}" | sed 's#ws#http#' | sed 's#/v1/opamp$##')"
       info "Testing connection to BindPlane: $fg_magenta$HTTP_ENDPOINT$reset..."
 
-      if curl -s ${HTTP_ENDPOINT} > /dev/null; then
+      if curl -s "${HTTP_ENDPOINT}" > /dev/null; then
         succeeded
       else
         failed
-        warn "Connection to BindPlane has failed. Do you wish to continue installation?"
+        warn "Connection to BindPlane has failed."
         increase_indent
-        printf "${fg_yellow}Do you wish to continue installation?${reset}  "
+        printf "%sDo you wish to continue installation?%s  " "$fg_yellow" "$reset"
         prompt "n"
         decrease_indent
         read -n 1 -p "" input
@@ -512,8 +511,8 @@ install_package()
   # Move files to install dir
   for f in $FILES
   do
-    rm -rf "$INSTALL_DIR/$f"
-    cp "$TMP_DIR/artifacts/$f" "$INSTALL_DIR/$f" || error_exit "$LINENO" "Failed to copy artifact $f to install dir"
+    rm -rf "$INSTALL_DIR/${f:?}"
+    cp "$TMP_DIR/artifacts/${f:?}" "$INSTALL_DIR/${f:?}" || error_exit "$LINENO" "Failed to copy artifact $f to install dir"
   done
   decrease_indent
   succeeded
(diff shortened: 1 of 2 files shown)


scripts/install/install_macos.sh
```diff
@@ -272,6 +278,36 @@ check_prereqs()
...
+      HTTP_ENDPOINT="$(printf "${opamp_endpoint}" | sed 's#ws#http#' | sed 's#/v1/opamp$##')"
+      info "Testing connection to BindPlane: $fg_magenta$HTTP_ENDPOINT$reset..."
+
+      if curl -s ${HTTP_ENDPOINT} > /dev/null; then
+        succeeded
+      else
+        failed
+        warn "Connection to BindPlane has failed. Do you wish to continue installation?"
+        increase_indent
+        printf "${fg_yellow}Do you wish to continue installation?${reset}  "
```
BinaryFissionGames, line 296: shellcheck is hitting me with this for this line:
https://www.shellcheck.net/wiki/SC2059

Maybe just needs an ignore comment like elsewhere in the script

scripts/install/install_macos.sh
```diff
@@ -272,6 +278,36 @@ check_prereqs()
...
+        prompt "n"
+        decrease_indent
+        read -n 1 -p "" input
```
BinaryFissionGames, line 299: The `-n` and `-p` flags aren't posix compliant, do we need these flags?

scripts/install/install_macos.sh
```diff
@@ -272,6 +278,36 @@ check_prereqs()
   decrease_indent
 }
 
+# Test connection to BindPlane if it was specified
+connection_check()
+{
+  info "Help Me Obi-Wan Kenobi!"
```
BinaryFissionGames, line 284: Leftover debug statement?

BinaryFissionGames (changes requested):
I only marked the macos script, but most of these apply to the linux script as well.

scripts/install/install_macos.sh
```diff
@@ -272,6 +278,36 @@ check_prereqs()
...
+  if [ $check_bp_url == "true" ] ; then
+    if [ -n "$opamp_endpoint" ]; then
+      HTTP_ENDPOINT="$(printf "${opamp_endpoint}" | sed 's#ws#http#' | sed 's#/v1/opamp$##')"
```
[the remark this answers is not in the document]
BinaryFissionGames: Shellcheck also hits this line with https://www.shellcheck.net/wiki/SC2059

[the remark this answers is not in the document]
Dylan-M: Good catch, and you're right it would nuke any of them. I'll fix that in my next revision.

scripts/install/install_macos.sh
Dylan-M: I forgot to run shellcheck after making some "changes" before commiting. Thanks for the reminder.

scripts/install/install_macos.sh
Dylan-M: They are not. This was an opinionated take, and makes it so you can just hit y or n without pressing enter. If we remove these flags, you'll need to hit enter after y or n.

scripts/install/install_macos.sh
Dylan-M: Yessir, good catch. Thanks.

Dylan-M: On this, I've noticed an inconsistency between OSX and unix scripts, where we run some checks before parsing arguments on OSX, but not in the unix script. Should that be fixed?

scripts/install/install_macos.sh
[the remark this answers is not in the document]
Dylan-M: Well, I thought it would nuke them, but...
```
echo "wss://some-wss-url" | sed 's#ws#http#'                   
https://some-wss-url
```

Tested on both OSX and Linux.

That said, still putting in explicit protection.

scripts/install/install_macos.sh
BinaryFissionGames: If you still want that old behavior, I found this on stack overflow:
```sh
# Adapted from https://unix.stackexchange.com/questions/464930/can-i-read-a-single-character-from-stdin-in-posix-shell
read_byte() 
{
  if [ -t 0 ]; then
    # disable icanon mode if this is a terminal
    saved_tty_settings=$(stty -g)
    stty -icanon min 1 time 0
  fi
 
  by=$(dd bs=1 count=1 2> /dev/null)
  
  printf "%s" "$by"
 
  if [ -t 0 ]; then
    # revert tty settings to their original values
    stty "$saved_tty_settings"
  fi
}
```

Works like:
```sh
input=$(read_byte)
```

scripts/install/install_macos.sh
```diff
@@ -272,6 +278,35 @@ check_prereqs()
...
 
+# Test connection to BindPlane if it was specified
+connection_check()
+{
+  if [ "$check_bp_url" = "true" ] ; then
+    if [ -n "$opamp_endpoint" ]; then
+      HTTP_ENDPOINT="$(echo "${opamp_endpoint}" | sed 's#^ws#http#' | sed 's#/v1/opamp$##')"
+      info "Testing connection to BindPlane: $fg_magenta$HTTP_ENDPOINT$reset..."
+
+      if curl -s "${HTTP_ENDPOINT}" > /dev/null; then
```
BinaryFissionGames, line 289: I tested this and noticed curl took a long time to timeout when I entered a bad hostname.
I think we should set the timeout, maybe 10s is too short but something shorter like this.
```suggestion
      if curl --max-time 10 -s "${HTTP_ENDPOINT}" > /dev/null; then
```

BinaryFissionGames (changes requested):
Seems to be working pretty well, just have one note on the timeout of the curl call.

scripts/install/install_macos.sh
BinaryFissionGames: We should make them consistent, but I'm not overly concerned about it, certainly wouldn't be against a follow-up PR fixing that 😄

scripts/install/install_macos.sh
Dylan-M: Good idea. I didn't even consider this, because both my Mac and my Linux test system time out very quickly.

I think 20 seconds is a reasonable amount of time to wait.

scripts/install/install_macos.sh
Dylan-M: Thanks, but I'll just do without it. I don't like when scripts modify tty settings. If you ctrl-c out of them at just the wrong time, your tty settings are borked. ;)

BinaryFissionGames (approved):

cpheps (approved):

cpheps:
Rebased this, should be good to merge once CI/CD passes.

PR #1434, opened 2024-01-30
fix(container): Java contrib JAR version

jsirianni:
### Proposed Change

Instead of downloading the contrib jar using a multi-stage build, we should use the jar that was downloaded by the build script. This will prevent the container image jar version from becoming out of sync with JAVA_CONTRIB_VERSION (file in this repo).

After building with `make release-test`, Our security tool no longer identifies JAVA vulnerabilities. When I exec into the container and get the jar's checksum, it matches the checksum of the jar found in `release_deps/`.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -256,6 +256,7 @@ dockers:
       - config/example.yaml
       - config/logging.stdout.yaml
       - LICENSE
+      - release_deps/opentelemetry-java-contrib-jmx-metrics.jar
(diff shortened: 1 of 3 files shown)


cpheps (approved):

BinaryFissionGames (approved):
Awesome catch, must have been leftover from before we used goreleaser for building the docker images!

PR #1435, opened 2024-01-30
feat(snapshotsearch): Filter logs

BinaryFissionGames:
### Proposed Change
* Add logic for filtering the logs payload before sending the report to bindplane.
* Add error check when prepping request payload

Tested (mostly) with unit tests. I plan on testing more extensively the e2e once we have all the pieces in place.

I have tested against current bindplane to ensure this change doesn't do anything unexpected with recent telemetry, everything on that front is working.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/internal/report/snapshot/filter.go
+++ b/internal/report/snapshot/filter.go
@@ -1,3 +1,17 @@
+// Copyright observIQ, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package snapshot
 
 import (
--- a/internal/report/snapshot/filter.go
+++ b/internal/report/snapshot/filter.go
@@ -20,14 +20,14 @@ import (
 	"go.opentelemetry.io/collector/pdata/pcommon"
 )
 
-func queryMatchesValue(v pcommon.Value, query string) bool {
+func queryMatchesValue(v pcommon.Value, searchQuery string) bool {
 	switch v.Type() {
 	case pcommon.ValueTypeMap:
 		// Recursively query for a match in the map (depth first search)
-		return queryMatchesMap(v.Map(), query)
+		return queryMatchesMap(v.Map(), searchQuery)
 	case pcommon.ValueTypeSlice:
 		// Iterate each element of the slice
-		return queryMatchesSlice(v.Slice(), query)
+		return queryMatchesSlice(v.Slice(), searchQuery)
 	case pcommon.ValueTypeEmpty:
 		// Cannot match empty value
 		return false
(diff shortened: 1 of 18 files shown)


internal/report/snapshot/filter.go
```diff
@@ -0,0 +1,78 @@
...
+		return false
+	default:
+		// We might be able to actually get away with just doing this for slices/maps, but could lead to
+		// weird edgecases since those slices/maps would be json-ified
+		// Note: Bytes will be base64 encoded and searched that way.
+		return strings.Contains(v.AsString(), query)
+	}
+}
+
+// func matchesAttributes(l plog.LogRecord, search string) {}
```
cpheps, line 42: Should this be removed?

internal/report/snapshot/filter_logs.go
```diff
@@ -0,0 +1,122 @@
...
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/plog"
+)
+
+// filterLogs filters the logs by the given query and timestamp.
+// The returned payload cannot be assumed to be a copy, so it should not be modified.
+func filterLogs(logs plog.Logs, query *string, minTimestamp *time.Time) plog.Logs {
+	// No filters specified, filtered logs are trivially the same as input logs
+	if query == nil && minTimestamp == nil {
+		return logs
```
cpheps, line 29: The base case does not return a copy is that a concern?

internal/report/snapshot/snapshot_buffers.go
```diff
@@ -86,7 +87,7 @@ func (l *LogBuffer) Add(ld plog.Logs) {
 }
 
 // ConstructPayload condenses the buffer and serializes to protobuf
-func (l *LogBuffer) ConstructPayload() ([]byte, error) {
+func (l *LogBuffer) ConstructPayload(searchQuery *string, minimumTimestamp *time.Time) ([]byte, error) {
```
cpheps, line 90: nit: Just for consistency we called the `minimumTimestamp` variable `minTimstamp` in the log filter code.

internal/report/snapshot_reporter.go
```diff
@@ -198,7 +201,7 @@ func (s *SnapshotReporter) SaveMetrics(componentID string, md pmetric.Metrics) {
 }
 
 // prepRequestPayload based on the pipelineType will return a marshaled proto of the OTLP data types for the componentID
-func (s *SnapshotReporter) prepRequestPayload(componentID, pipelineType string) (payload []byte, err error) {
+func (s *SnapshotReporter) prepRequestPayload(componentID, pipelineType string, searchQuery *string, minimumTimestamp *time.Time) (payload []byte, err error) {
```
cpheps, line 204: nit: Same here the `searchQuery` and `minimumTimestamp` variable names are not consistent.

cpheps (changes requested):

internal/report/snapshot/filter_logs.go
cpheps: I think not after looking at the rest of the PR as a copy is passed in by I'll keep this question just to make sure.

BinaryFissionGames: Yeah, the code ends up basically being equivalent if search and mintimestamp are both unspecified.
This function does have a comment above it explaining that a copy cannot be assumed, just in case.

internal/report/snapshot/filter.go
BinaryFissionGames: removed

internal/report/snapshot/snapshot_buffers.go
BinaryFissionGames: good call, made these consistent

cpheps (approved):

shazlehu (approved):
Worked correctly in all my testing. I gave the code a close read reusing it for trace search

PR #1438, opened 2024-02-01
feat(snapshotsearch): Filter traces

shazlehu:
### Proposed Change

Adds searching snapshot traces. To test, use bindplane branch in this PR: https://github.com/observIQ/bindplane-op-enterprise/pull/2016 and follow the testing instructions there.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


internal/report/snapshot/snapshot_buffers.go
```diff
@@ -258,7 +258,7 @@ func (l *TraceBuffer) Add(td ptrace.Traces) {
 }
 
 // ConstructPayload condenses the buffer and serializes to protobuf
-func (l *TraceBuffer) ConstructPayload() ([]byte, error) {
+func (l *TraceBuffer) ConstructPayload(searchQuery *string, minimumTimestamp *time.Time) ([]byte, error) {
```
cpheps, line 261: nit: We changed this in the logging PR but the parameters `searchQuery` and `minimumTimestamp` differ from `query` and `minTimestamp` used in the `filter_traces.go` file. I forget what Brandon standardized, but we should make them all the same.

internal/report/snapshot_reporter.go
```diff
@@ -229,7 +229,7 @@ func (s *SnapshotReporter) prepRequestPayload(componentID, pipelineType string,
 			return []byte{}, nil
 		}
 
-		payload, err = buffer.ConstructPayload()
+		payload, err = buffer.ConstructPayload(searchQuery, minimumTimestamp)
```
cpheps, line 232: Same here naming of these.

cpheps (changes requested):

cpheps (approved):

internal/report/snapshot/filter_traces.go
```diff
@@ -0,0 +1,121 @@
...
+
+// spanMatchesQuery determines if the given span matches the given query string
+func spanMatchesQuery(span ptrace.Span, query string) bool {
+	return queryMatchesMap(span.Attributes(), query) ||
+		strings.Contains(span.Name(), query) ||
+		strings.Contains(span.TraceID().String(), query) ||
+		strings.Contains(span.SpanID().String(), query) ||
+		strings.Contains(span.ParentSpanID().String(), query) ||
+		strings.Contains(span.StartTimestamp().String(), query) ||
+		strings.Contains(span.EndTimestamp().String(), query)
```
BinaryFissionGames, line 120: The ticket lists:
```
all attributes, resource attributes, name, event, kinds
```
I think we're missing kinds. I'm not sure if we should actually match events, I don't think we show them in the UI IIRC.

Also, do we want matching on the timestamp? If we want to we should go back in a follow-up PR and add it for logs.

BinaryFissionGames (changes requested):

internal/report/snapshot/filter_traces.go
cpheps: Yeah, use your best judgment on this. I'm not as familiar with the trace structure so I wasn't sure what should be checked.

shazlehu: Timestamp matching requires parsing the search as a timestamp to be reliable, so we should probably leave it out for now. I also don't know how useful span ID/parent ID will be as these are pretty ephemeral. It seemed like low-hanging fruit, and its absence would frustrate a user who had a good use case (particularly for parent ID in a long-running span)

- [ ]

BinaryFissionGames: Yeah, I think the span IDs are fine to keep, and unlikely to conflict with other things since they are unique IDs.

BinaryFissionGames (approved):

PR #1466, opened 2024-02-14
feat: Add Snowflake Exporter

dpaasman00:
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


exporter/snowflakeexporter/go.mod
```diff
@@ -0,0 +1,97 @@
...
+require (
+	github.com/DATA-DOG/go-sqlmock v1.5.2
+	github.com/jmoiron/sqlx v1.3.5
+	github.com/lib/pq v1.10.9
+	github.com/snowflakedb/gosnowflake v1.7.2
+	github.com/stretchr/testify v1.8.4
+	go.opentelemetry.io/collector/component v0.92.0
+	go.opentelemetry.io/collector/config/configretry v0.92.0
+	go.opentelemetry.io/collector/consumer v0.92.0
+	go.opentelemetry.io/collector/exporter v0.92.0
```
cpheps, line 14: The release branch is on otel 0.94.0 now

exporter/snowflakeexporter/internal/database/database.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// BatchInsert creates a new transaction using the given DB to insert the given data
+func (s *Snowflake) BatchInsert(ctx context.Context, data []map[string]any, sql string) error {
+	// create TX
+	tx, err := s.db.BeginTxx(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("failed to create transaction: %w", err)
+	}
+	defer tx.Rollback()
```
cpheps, line 107: Is this a noop if the commit is successful?

exporter/snowflakeexporter/internal/utility/utility.go
```diff
@@ -0,0 +1,87 @@
...
+// SpanIDToHexOrEmptyString returns a string representation of the SpanID in hex
+func SpanIDToHexOrEmptyString(id pcommon.SpanID) string {
+	if id.IsEmpty() {
+		return ""
+	}
+	return hex.EncodeToString(id[:])
+}
+
+// FlattenExemplars will flatten the given exemplars into slices of the individual fields
+func FlattenExemplars(exemplars pmetric.ExemplarSlice, l *zap.Logger) (pq.StringArray, pq.StringArray, pq.StringArray, pq.StringArray, pq.Float64Array) {
```
cpheps, line 64: These functions don't have tests are we able to test them?

cpheps (changes requested):

exporter/snowflakeexporter/internal/database/database.go
dpaasman00: Yep!

exporter/snowflakeexporter/exporter_traces.go
```diff
@@ -0,0 +1,274 @@
...
+	err := te.db.BatchInsert(ctx, traceMaps, te.insertSQL)
+	if err != nil {
+		return fmt.Errorf("failed to insert trace data: %w", err)
+	}
+	te.logger.Debug("end tracesDataPusher")
+	return nil
+}
+
+func flattenEvents(e ptrace.SpanEventSlice, l *zap.Logger) (pq.StringArray, pq.StringArray, pq.StringArray, pq.Int32Array) {
+	times := pq.StringArray{}
```
BinaryFissionGames, line 243: Is PQ appropriate to use here? It's a postgres database driver, so it seems iffy to me to use it for snowflake.

exporter/snowflakeexporter/internal/utility/utility.go
dpaasman00: Yea I definitely can. I didn't originally because they're from an [internal package](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/coreinternal/traceutil/traceutil.go) in collector-contrib. Because they are in an internal pkg I couldn't import it, so I just copied them here since they're pretty simple

cpheps: Gotcha, lets copy the tests then and also just put a comment on here pointing out where you got them from.

dpaasman00: Gotcha, it doesn't look like they test them so I'll come up with a couple quick ones and add the comments

exporter/snowflakeexporter/exporter_traces.go
BinaryFissionGames: Snowflake seems to have a way to bind array values? Hard for me to tell if this works here.
https://pkg.go.dev/github.com/snowflakedb/gosnowflake#Array

BinaryFissionGames (changes requested):

exporter/snowflakeexporter/README.md
```diff
@@ -0,0 +1,189 @@
...
+| Field | Type | Default | Required | Description |
+| ----- | ---- | ------- | -------- | ----------- |
+| account_identifier | string | | `true` | The account identifier number of the Snowflake account telemetry data will be sent to. |
+| username | string | | `true` | The username for the account the exporter will use to authenticate with Snowflake. |
+| password | string | | `true` | The password for the account the exporter will use to authenticate with Snowflake. |
+| warehouse | string | | `true` | The Snowflake data warehouse that should be used for storing data. |
+| role | string | | `false` | The Snowflake role that the exporter should use to have the correct permissions. Only necessary if the default role of the given user is not the necessary role. |
+| database | string | `bpop` | `false` | The database in Snowflake that the exporter will store telemetry data in. Will create it if it doesn't exist. |
+| parameters | map | | `false` | A map of optional connection parameters that may be used for connecting with Snowflake. |
+| logs.enabled | bool | | `true` | Tells the exporter whether or not logs will be sent to Snowflake. |
```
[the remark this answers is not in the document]
dpaasman00: Yea they can be removed. I originally added them to solve a different problem but when you put it like that I see I don't really need them to solve that problem

exporter/snowflakeexporter/exporter_traces.go
dpaasman00: Yea I don't that feel great about it either. I'll check out that Array function and see how it works

exporter/snowflakeexporter/config.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	if c.Traces.Schema == "" {
+		c.Traces.Schema = defaultTracesSchema
+	}
+	if c.Traces.Table == "" {
+		c.Traces.Table = defaultTable
+	}
+
+	return nil
+}
```
BinaryFissionGames, line 124: For these defaults, I think we should just set them in `factory.go` in createDefaultConfig (unless there's some issue I'm missing with that)

exporter/snowflakeexporter/config.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	if c.Password == "" {
+		return errors.New("password is required")
+	}
+	if c.Warehouse == "" {
+		return errors.New("warehouse is required")
+	}
+	if c.Database == "" {
+		c.Database = defaultDatabase
+	}
```
BinaryFissionGames, line 77: Same here, seems like this could just be specified in createDefaultConfig

exporter/snowflakeexporter/internal/utility/utility.go
```diff
@@ -0,0 +1,102 @@
...
+
+// ConvertAttributesToString converts the pcommon.Map into a JSON string representation
+// this is due to a bug/lacking feature with the snowflake driver that prevents maps from being inserted into VARIANT & OBJECT columns
+// github issue: https://github.com/snowflakedb/gosnowflake/issues/217
+func ConvertAttributesToString(attributes pcommon.Map, logger *zap.Logger) string {
+	m := make(map[string]string, attributes.Len())
+	attributes.Range(func(k string, v pcommon.Value) bool {
+		m[k] = v.AsString()
+		return true
+	})
```
BinaryFissionGames, line 51: Any reason we can't just marshal attributes.AsRaw()?
Otherwise this might look weird with nested attribute maps.

exporter/snowflakeexporter/internal/utility/utility_test.go
```diff
@@ -0,0 +1,121 @@
...
+		expected   string
+	}{
+		{
+			desc: "simple",
+			attributes: map[string]any{
+				"k1": "v1",
+				"k2": "v2",
+			},
+			expected: "{\"k1\":\"v1\",\"k2\":\"v2\"}",
+		},
```
BinaryFissionGames, line 38: I think we should add a test for a nested map, like
```go
map[string]any{
  "k1": map[string]any{
    "k2": "val"
   },
},
```

BinaryFissionGames (changes requested):

exporter/snowflakeexporter/config.go
dpaasman00: I don't think there's an issue with that... the only thing I can think of is if the user intentionally sets one of those values as an empty string the collector will fail when interacting with snowflake. But we probably want to error here in that case rather than just overwriting with the default and not telling the user the config is bad

exporter/snowflakeexporter/internal/utility/utility.go
dpaasman00: Yea I don't see why not, I should've noticed that while adding that new Value func

exporter/snowflakeexporter/internal/utility/utility.go
```diff
@@ -0,0 +1,96 @@
...
+
+// ConvertAttributesToString converts the pcommon.Map into a JSON string representation
+// this is due to a bug/lacking feature with the snowflake driver that prevents maps from being inserted into VARIANT & OBJECT columns
+// github issue: https://github.com/snowflakedb/gosnowflake/issues/217
+func ConvertAttributesToString(m pcommon.Map, logger *zap.Logger) string {
+	bytes, err := json.Marshal(m.AsRaw())
+	if err != nil {
+		logger.Warn("failed to marshal attribute map", zap.Error(err), zap.Any("map", m))
+		return ""
+	}
```
dpaasman00, line 51: Do we know if it's even possible for json.Marshal to fail here? Like is it possible for a pcommon.Map to include a value that would cause it to fail?

cpheps: Likely it won't fail but it's always good to check the error just in case. If it does happen we'll want to know rather than encountering a panic.

cpheps (approved):

exporter/snowflakeexporter/internal/utility/utility.go
dpaasman00: Sounds good!

PR #1468, opened 2024-02-15
feat: Add Lookup Processor

jmwilliams89:
### Proposed Change
- Added a Lookup Processor
  - This processor loads a CSV into memory and will match incoming telemetry to records in this file. If a match is found, all other columns of that record will be added to the telemetry. 
  - At the moment, this processor takes a simplified approach when checking and adding telemetry. Telemetry is only checked and modified at the defined `context` of the processor. We can expand upon this basic approach in the future if other use cases arise.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


processor/lookupprocessor/README.md
```diff
@@ -0,0 +1,47 @@
...
+1. This processor will load and periodically refresh a CSV file in memory.
+2. When telemetry is received, the processor checks if the configured `field` exists in the configured `context`.
+3. If the field exists and the CSV contains a matching value, all other values in that record are added to the `context` of the telemetry. The header of each value is used when adding these fields.
+
+## Configuration
+| Field        | Type     | Default | Description |
+| ---          | ---      | ---     | ---         |
+| csv          | string   | ` `     | The location of the CSV file used for lookups. The processor will periodically reload this in memory every minute. |
+| context      | string   | ` `     | The context of the telemetry to check and use when performing lookups. Supported values are `attributes`, `body`, `resource`. |
+| field        | string   | ` `     | The field to match when performing a lookup. For a lookup to succeed, the field name and value must be the same in the CSV file and telemetry. |
```
BinaryFissionGames, line 19: Is there any way we can use OTTL here for context instead? e.g. context is a OTTL path to a pdata.Map instead, so it can support nested attributes and the like.

BinaryFissionGames (changes requested):
Looks pretty good code-wise, question on desired configuration.

processor/lookupprocessor/README.md
BinaryFissionGames: Asking because we've been wanting to use OTTL as much as possible instead of defining our own systems for specifying paths in telemetery, for consistency.

jmwilliams89: I considered OTTL, but wanted to avoid making this processor too complicated until we've proved out the customer use case first. I tried to architect the config in such a manner that we could adopt OTTL later if it's necessary. At the moment, the simplicity of this design was intentional to avoid getting bogged down in edge cases and get this in a customer's hands first.

processor/lookupprocessor/processor.go
```diff
@@ -0,0 +1,301 @@
...
+		context: cfg.Context,
+		field:   cfg.Field,
+	}
+}
+
+// start starts the processor
+func (p *lookupProcessor) start(_ context.Context, _ component.Host) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	p.cancel = cancel
+	go p.loadCSV(ctx)
```
cpheps, line 52: We should put wait groups around this. Since the collector is an in memory process if this goroutine takes a long time to shut down or doesn't at all it can leak.

processor/lookupprocessor/README.md
jmwilliams89: However, if you think it's extremely easy to add, I would not turn down a code snippet/commit showing how to do it. I just wanted to avoid getting too wrapped up in less likely use cases, until we've gotten feedback.

cpheps (changes requested):

cpheps (approved):
Approving but may review again if we change things to OTTL.

processor/lookupprocessor/README.md
BinaryFissionGames: Eh, I think it would take a bit of refactor, might end up simplifying the code a little but maybe it's not worth the effort right now.

Suggestion instead then; Can we change `resource` to `resource.attributes` so that if we DO want to make the context field OTTL in the future, it's not a breaking change?

processor/lookupprocessor/README.md
```diff
@@ -0,0 +1,47 @@
...
+- Logs
+- Metrics
+- Traces
+
+## How It Works
+1. This processor will load and periodically refresh a CSV file in memory.
+2. When telemetry is received, the processor checks if the configured `field` exists in the configured `context`.
+3. If the field exists and the CSV contains a matching value, all other values in that record are added to the `context` of the telemetry. The header of each value is used when adding these fields.
+
+## Configuration
```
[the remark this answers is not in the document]
BinaryFissionGames: This can lead to non-obvious behavior (e.g. you can imagine 2 logs sharing 1 resource, if they have different IP values they'd both end up with the attributes for the second log, so the first log would have the incorrect attributes on its resource)

processor/lookupprocessor/README.md
jmwilliams89: Yeah, I think that's a reasonable change to make. Since we'll be wrapping this with a BindPlane processor, we can just show it as `resource` there.

processor/lookupprocessor/README.md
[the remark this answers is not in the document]
jmwilliams89: There's also some complexity when it comes to the attributes -> body use case. If the body isn't already a map value, we have to decide whether or not to completely erase the string value or skip this operation altogether.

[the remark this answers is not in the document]
jsirianni: I guess in my head I would like to use a body field for the lookup, and place the resulting values into attributes. In this example, I am checking for `body.ip` and I would like to have `service` and `region` go to attributes. Right now they go to body because my lookup field was under body.

Screenshot from 2024-02-15 14-50-45

I cannot think of a reason someone would want to lookup an attribute and place the values into body, but the opposite would be a fairly common use-case.

Going to resource.attributes is complex like Brandon mentioned, so I suppose that case can be ignored for now (users could use the group by processor to "promote" attributes).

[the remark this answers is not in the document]
jmwilliams89: In this example, would you eventually also move `ip` to attributes, or would you plan on leaving it in body?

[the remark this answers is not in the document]
jsirianni: IP is in the log body so I would probably just leave it where it originated. Its just a thought, I am not a stickler on it and I think it is something that could be added later without breaking the current functionality.

jsirianni (approved):

BinaryFissionGames (approved):

PR #1502, opened 2024-02-27
chore: Merge v1.45.0 to main in prep for release

BinaryFissionGames:
### Proposed Change
Merge v1.45.0 to main in prep for release

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cpheps (approved):

go.mod
```diff
@@ -1,262 +1,267 @@
 module github.com/observiq/bindplane-agent
 
-go 1.20
+go 1.21
```
dpaasman00, line 3: Are we updating the go version on purpose here? totally fine if so, just noticed normally we keep it at 1.20

BinaryFissionGames: Yes. Upstream updated go, so we don't have a choice 😅

dpaasman00: Fair enough

PR #1506, opened 2024-02-27
feat: Synthetic Telemetry Generator Receiver

shazlehu:
### Proposed Change

Adds a receiver that produces synthetic telemetry based on the config. Currently only supports logs.

### Testing

Configure the agent with the following custom logs source:
```yaml
telemetrygeneratorreceiver:
    payloads_per_second: 1
    generators:
        - type: logs
          resource_attributes:
              res_key1: res_value1
              res_key2: res_value2            
          attributes:
              attr_key1: attr_value1
              attr_key2: attr_value2            
          additional_config:
              body: this is the body   
              severity: 4
        - type: logs
          resource_attributes:
              res_key3: res_value3
              res_key4: res_value4            
          attributes:
              attr_key3: attr_value3
              attr_key4: attr_value4            
          additional_config:
              body: this is another body   
              severity: 10         
```

Send it to any destination (custom `nop:` is fine), snapshot console should show logs flowing.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/telemetrygeneratorreceiver/config.go
```diff
@@ -0,0 +1,94 @@
...
+// GeneratorConfig is the configuration for a single generator
+type GeneratorConfig struct {
+	// Type of generator to use, either "logs", "host_metrics", or "windows_events"
+	Type GeneratorType `mapstructure:"type"`
+
+	// ResourceAttributes are additional key-value pairs to add to the resource attributes of telemetry.
+	ResourceAttributes map[string]string `mapstructure:"resource_attributes"`
+
+	// Attributes are Additional key-value pairs to add to the telemetry attributes
+	Attributes map[string]string `mapstructure:"attributes"`
```
cpheps, line 38: These should be `map[string]any` as that's what pdata supports for attributes.

receiver/telemetrygeneratorreceiver/factory.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// createDefaultConfig creates a default configuration
+func createDefaultConfig() component.Config {
+	return &Config{
+		PayloadsPerSecond: 1,
+		Generators: []*GeneratorConfig{
+			{
+				Type: GeneratorTypeLogs,
+			},
```
cpheps, line 48: There should be no generator for the default. A single generator is required on configuration so we don't want to supply one.

receiver/telemetrygeneratorreceiver/generators.go
```diff
@@ -0,0 +1,69 @@
...
+	// GeneratorTypeWindowsEvents is the generator type for windows events
+	GeneratorTypeWindowsEvents GeneratorType = "windows_events"
+)
+
+type generator interface {
+	// SupportsType returns true if the generator supports the given component type, either metrics, logs, or traces.
+	SupportsType(component.Type) bool
+	GenerateMetrics() pmetric.Metrics
+	GenerateLogs() plog.Logs
+	GenerateTraces() ptrace.Traces
```
cpheps, line 44: Can we add comments on these?

receiver/telemetrygeneratorreceiver/go.mod
```diff
@@ -0,0 +1,42 @@
+module github.com/observiq/bindplane-agent/receiver/telemetrygeneratorreceiver
+
+go 1.21.6
```
cpheps, line 3: Can we add this go mod to the dependabot file. Example for MS365 receiver.

https://github.com/observIQ/bindplane-agent/blob/release/v1.46.0/.github/dependabot.yml#L155-L165

receiver/telemetrygeneratorreceiver/host_metrics_generator.go
```diff
@@ -0,0 +1,55 @@
+// Copyright observIQ, Inc.
```
cpheps, line 1: This doesn't have a test file is that coming later?

receiver/telemetrygeneratorreceiver/logs_generator.go
```diff
@@ -0,0 +1,101 @@
+// Copyright observIQ, Inc.
```
cpheps, line 1: No test file, will one be added in a follow up?

receiver/telemetrygeneratorreceiver/logs_receiver.go
```diff
@@ -0,0 +1,60 @@
...
+	lr.generators = newGenerators(cfg, logger, r.supportedTelemetry)
+
+	return lr, nil
+}
+
+// generateTelemetry
+func (r *logsGeneratorReceiver) produce() error {
+	logs := plog.NewLogs()
+	for _, g := range r.generators {
+		l := g.GenerateLogs()
```
cpheps, line 53: Should do a check here with `g.SupportsType(component.Logs)`, and skip if it's false.

receiver/telemetrygeneratorreceiver/receiver.go
```diff
@@ -0,0 +1,97 @@
...
+		defer ticker.Stop()
+
+		err := r.producer.produce()
+		if err != nil {
+			r.logger.Error("Error generating telemetry", zap.Error(err))
+		}
+		for {
+			select {
+			case <-ctx.Done():
+				return
```
cpheps, line 85: Don't listen for done on this context. The context passed into start isn't meant to be used as a signal for when to stop concurrent actions.

cpheps (changes requested):

receiver/telemetrygeneratorreceiver/receiver.go
BinaryFissionGames, line 85: This is a short-lived context, long running operations (ones that outlive Start) should not use this context.

receiver/telemetrygeneratorreceiver/config.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// AdditionalConfig are any additional config that a generator might need.
+	AdditionalConfig map[string]any `mapstructure:"additional_config"`
```
BinaryFissionGames, line 41: This is more a design question, do we want this to be nested, or do we want this to be on the same level as the other options? If we want, we could allow these values to just be place on the same level as `attributes` and `resource_attributes`

https://pkg.go.dev/github.com/mitchellh/mapstructure#hdr-Remainder_Values

receiver/telemetrygeneratorreceiver/generators.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// initialize must called when the generator is created. It is used to prevent the need to
+	// recreate the telemetry every generation cycle
+	initialize()
```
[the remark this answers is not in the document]
shazlehu: The design Corbin and I discussed allows for generators that emit more than one telemetry type. Pushing initialization off until after the constructor avoids building the telemetry for a generator that doesn't produce the type the receiver is using.

receiver/telemetrygeneratorreceiver/config.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"errors"
+	"fmt"
+)
+
+// Config is the configuration for the telemetry generator receiver
+type Config struct {
+	PayloadsPerSecond int                `mapstructure:"payloads_per_second"`
+	Generators        []*GeneratorConfig `mapstructure:"generators"`
```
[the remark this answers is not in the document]
shazlehu: My thinking was that these configs could get pretty hefty, so trying to avoid copying

receiver/telemetrygeneratorreceiver/generators.go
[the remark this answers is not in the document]
cpheps: @BinaryFissionGames Curious if you have a proposed design. I went with this because of the types returned. I tried to think of a generic Generate function.

Are you purposing something like this but one for each type?

```go
type metricsGenerator interface {
    Generate() pmetric.Metrics
}
```

go.mod
```diff
@@ -1,37 +1,36 @@
...
+	github.com/observiq/bindplane-agent/processor/metricstatsprocessor v1.45.0
+	github.com/observiq/bindplane-agent/processor/removeemptyvaluesprocessor v1.45.0
+	github.com/observiq/bindplane-agent/processor/resourceattributetransposerprocessor v1.45.0
+	github.com/observiq/bindplane-agent/processor/samplingprocessor v1.45.0
+	github.com/observiq/bindplane-agent/processor/spancountprocessor v1.45.0
+	github.com/observiq/bindplane-agent/processor/throughputmeasurementprocessor v1.45.0
+	github.com/observiq/bindplane-agent/receiver/azureblobrehydrationreceiver v1.45.0
+	github.com/observiq/bindplane-agent/receiver/httpreceiver v1.45.0
+	github.com/observiq/bindplane-agent/receiver/m365receiver v1.45.0
+	github.com/observiq/bindplane-agent/receiver/pluginreceiver v1.45.0
```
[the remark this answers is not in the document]
cpheps: Resolving as Brandon has duplicate comment.

receiver/telemetrygeneratorreceiver/generators.go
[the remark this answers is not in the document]
cpheps: Resolving this as Brandon had a similar comment.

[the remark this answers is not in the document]
cpheps: I actually would prefer the initialize be part of the newX so that we don't need two methods to do the additional setup. I'll resolve my comment for this.

receiver/telemetrygeneratorreceiver/receiver.go
cpheps: Resolving as Brandon had a similar comment.

receiver/telemetrygeneratorreceiver/config.go
cpheps: I originally designed it to be an additional field. I didn't know mapstructure has the `remain`. I don't have a strong opinion either way.

receiver/telemetrygeneratorreceiver/generators.go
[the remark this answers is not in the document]
BinaryFissionGames: Yeah, I thought about this a little more and I was thinking it would make more sense to have 3 interfaces, one for each type

```go
type metricsGenerator interface {
    GenerateMetrics() pmetric.Metrics
}
```
Then one for logs and one for traces.

Then you can just have a 
`map[string]metricsGeneratorCreateFunc` 
(`type metricsGeneratorCreateFunc func() metricsGenerator`), which would make it easy to:
1. check if a certain generator exists for the given telemetry type
2. create the generator from the type in a typesafe manner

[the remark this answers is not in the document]
BinaryFissionGames: This is how I'd see it being setup upstream, it's similar to how something like consumers are set up in OTel, like you have ConsumeMetrics, ConsumeLogs, etc.

receiver/telemetrygeneratorreceiver/config.go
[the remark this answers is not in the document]
BinaryFissionGames: The GeneratorConfig itself is pretty light. Maps and strings are both fat pointers, I'd prefer using values here just to get the nil safety.

receiver/telemetrygeneratorreceiver/generators.go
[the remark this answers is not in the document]
cpheps: How does upstream do it? Do they use reflection to see if an object implements an interface? I'm thinking if I have say two interfaces and configured all three pipeline types do I have to check at runtime to see if a struct implements an interface before trying to pass it in as such?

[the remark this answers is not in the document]
BinaryFissionGames: So, I think I missed that we were trying to re-use the type for each pipeline when I suggested this.

I think they upstream DOES avoid reflection, but I can't find where they reuse already-created components, if they do (the pipeline creation code is kinda complicated, so I easily could have just missed it).

Personally, I'm in favor of the type assertions here, I feel like in principle it's not really different from the SupportsType function, and if we're gonna be creating a lot of generators it'd be nice to not have to create stub methods for telemetry types that are not used (e.g. I create a logs-only generator and now have to stub out traces and metrics).

That being said, I get that their are downside to that approach, so if we feel like the cons of the type assertion is bad and we feel like we need to use the same struct across traces, metrics, and logs, then I'm OK with the current approach.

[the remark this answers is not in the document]
cpheps: Yeah, I agree with the fact of having extra stub methods. It makes the interface feel too robust. @shazlehu Can we switch to the design Brandon has suggested?

@BinaryFissionGames Also, I wonder if they just do a type check. Can you cast to an interface an interface as a way of seeing if you implement it?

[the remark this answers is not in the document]
BinaryFissionGames: You're talking about a type assertion like `mGen, ok := generator.(metricsGenerator)`, right? That's what I mean to suggest. It's how stuff like e.g. error unwrapping is handled in the go stdlib 

https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/errors/wrap.go;l=17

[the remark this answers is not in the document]
BinaryFissionGames: I'd actually like if we verified somehow that when this is included in multiple pipeline types, that it doesn't create separate receivers. I think it actually does, and doing this type assertion thing is maybe actually pointless.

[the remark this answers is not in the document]
cpheps: >I'd actually like if we verified somehow that when this is included in multiple pipeline types, that it doesn't create separate receivers.

Yeah, I forgot about this. It does create a new receiver per pipeline. Generators don't know what pipeline their in so we need some sort of type assertion or check to see if a generator is compatible with the pipeline it's configured for.

receiver/telemetrygeneratorreceiver/logs_receiver.go
shazlehu: newGenerators only returns generators that support the telemetry type

receiver/telemetrygeneratorreceiver/generators.go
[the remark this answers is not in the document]
cpheps: Ah I missed the map creation part of [this comment](https://github.com/observIQ/bindplane-agent/pull/1506#discussion_r1507851079). Makes sense to me.

receiver/telemetrygeneratorreceiver/config.go
```diff
@@ -0,0 +1,117 @@
...
+	Generators        []GeneratorConfig `mapstructure:"generators"`
+}
+
+// GeneratorConfig is the configuration for a single generator
+type GeneratorConfig struct {
+	// Type of generator to use, either "logs", "host_metrics", or "windows_events"
+	Type generatorType `mapstructure:"type"`
+
+	// ResourceAttributes are additional key-value pairs to add to the resource attributes of telemetry.
+	ResourceAttributes map[string]string `mapstructure:"resource_attributes"`
```
cpheps, line 38: Resource attributes should also be a `map[string]any` they are the same type I believe in pdata.

receiver/telemetrygeneratorreceiver/config.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	return nil
+}
+
+// Validate validates the generator config
+func (g *GeneratorConfig) Validate() error {
+	if g.Type == "" {
+		return errors.New("type must be set")
+	}
```
cpheps, line 67: This check is redundant as it's covered by the default case.

receiver/telemetrygeneratorreceiver/generators.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	return generators
+}
+
+// newMetricsGenerators creates a slice of metricGenerator based on the provided configuration and logger.
+func newMetricsGenerators(cfg *Config, logger *zap.Logger) []metricGenerator {
+	var generators []metricGenerator
+	for _, gen := range cfg.Generators {
+		switch gen.Type {
+		case generatorTypeLogs:
```
cpheps, line 72: copy paste error here. This should be the metrics type.

receiver/telemetrygeneratorreceiver/windows_events_generator.go
```diff
@@ -0,0 +1,39 @@
...
+package telemetrygeneratorreceiver //import "github.com/observiq/bindplane-agent/receiver/telemetrygeneratorreceiver"
+
+import (
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.uber.org/zap"
+)
+
+// windowsEventsMetricsGenerator is a generator for Windows Event Log metrics. It generates a sampling of Windows Event Log metrics
+// emulating the Windows Event Log receiver: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/windowseventlogreceiver
+
```
cpheps, line 24: nit remove space

cpheps (changes requested):
Looks pretty good few smaller things

cpheps (approved):

go.mod
```diff
@@ -32,6 +30,7 @@ require (
 	github.com/observiq/bindplane-agent/receiver/pluginreceiver v1.46.0
 	github.com/observiq/bindplane-agent/receiver/routereceiver v1.46.0
 	github.com/observiq/bindplane-agent/receiver/sapnetweaverreceiver v1.46.0
+	github.com/observiq/bindplane-agent/receiver/telemetrygeneratorreceiver v0.0.0-00010101000000-000000000000
```
BinaryFissionGames, line 33:
```suggestion
	github.com/observiq/bindplane-agent/receiver/telemetrygeneratorreceiver v1.46.0
```

receiver/telemetrygeneratorreceiver/receiver.go
```diff
@@ -0,0 +1,95 @@
...
+
+// telemetryProducer is an interface for producing telemetry used by the specific receivers
+type telemetryProducer interface {
+	// produce should generate telemetry from each generator, update the timestamps to now, and send it to the next consumer
+	produce() error
+}
+type telemetryGeneratorReceiver struct {
+	logger             *zap.Logger
+	cfg                *Config
+	supportedTelemetry component.DataType
```
BinaryFissionGames, line 35: This field seems unused, I think we can remove it.

BinaryFissionGames (changes requested):

BinaryFissionGames (approved):

PR #1507, opened 2024-03-04
feat(sysv init): Add support for EL6 and SLES11

Dylan-M:
### Proposed Change
Have a fully working init.d service file for both EL6 and SLES11SP4.

Changes have been fully dev tested on EL6. However, we do not (and can not) have a SLES11SP4 system at this time. Reaching out to customer to request access to one of theirs.

If anyone from engineering needs/wants to test this, just ask Dylan for an EL6 system.

Additional work needed:
- [x] Install script needs to be updated once build process is finished to properly lay down the correct service file based on Linux OS. I already have the edited service file almost done. I'll attach to this once complete.
- [x] Add support for SysV init script laydown to RPM via the build process

##### Checklist
- [x] Changes are tested (with exceptions noted above)
- [x] CI has passed


Dylan-M:
I've tested this using the install script and RPM locally, and it correctly laid down the service file, started the service, and my agent appeared in my BindPlane instance.

.goreleaser.yml
```diff
@@ -196,6 +202,20 @@ nfpms:
...
           group: root
+      - src: service/observiq-otel-collector
+        dst: /etc/init.d/observiq-otel-collector
+        type: config|noreplace
+        file_info:
+          mode: 0755
+          owner: root
+          group: root
+      - src: service/sysconfig/observiq-otel-collector
+        dst: /etc/sysconfig/observiq-otel-collector
```
jsirianni, line 213: Can you review this page and decide if noreplace is correct here? https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html.

Trying to determine if we expect a user to make changes to this file. Maybe we do?

service/sysconfig/observiq-otel-collector
```diff
@@ -0,0 +1,18 @@
+# Configuration file for the observiq-otel-collector service.
+
+# The server keys are automatically generated if they ommited
+# to change the automatic creation uncomment the approprite 
+# line.
```
jsirianni, line 5: Is this a hold over from something else?

scripts/package/postinstall.sh
```diff
@@ -50,11 +50,20 @@ If you have any other questions please contact us at support@observiq.com
...
+  chmod 755 /etc/init.d/observiq-otel-collector
+  chmod 644 /etc/sysconfig/observiq-otel-collector
+  echo "configured sysv service"
+}
+
 init_type() {
-  systemd_test="$(systemctl 2>/dev/null || : 2>&1)"
-  if command printf "$systemd_test" | grep -q '\-.mount'; then
+  # Determine if we need service or systemctl for prereqs
+  if command -v systemctl > /dev/null 2>&1; then
```
jsirianni, line 61: This seems safe but I want to be sure. I cannot recall why we would parse the output of systemctl. What do you think?

.goreleaser.yml
```diff
@@ -108,6 +108,12 @@ archives:
       - src: release_deps/observiq-otel-collector.service
         dst: "install"
         strip_parent: true
+      - src: release_deps/observiq-otel-collector
+        dst: "install"
+        strip_parent: true
+      - src: release_deps/sysconfig/observiq-otel-collector
+        dst: "install/sysconfig"
+        strip_parent: true
```
jsirianni, line 116: I believe these can be sourced from `service/observiq-otel-collector` and `service/sysconfig/observiq-otel-collector`, and then you can revert the addition to `release-prep`. IIRC, `release-prep` is for generated configs.

jsirianni (changes requested):
This looks great, I still need to run through some extensive testing. Originally I thought we would have a new package, this seems better but will require more testing to ensure our systemd stuff does not contain regressions.

service/sysconfig/observiq-otel-collector
Dylan-M: Yes, the file I copied it from. Fixing now.

scripts/package/postinstall.sh
Dylan-M: We used this method elsewhere. If it is safe elsewhere, I would think it is safe here as well.

.goreleaser.yml
Dylan-M: Okay, I was working from the systemd file as an example, and it is handled the same way. Is it being handled incorrectly?

jsirianni, line 116: This is what it looks like when extracted. Kind of confusing. (two `observiq-otel-collector`)

```
install
├── com.observiq.collector.plist
├── observiq-otel-collector
├── observiq-otel-collector.service
├── sysconfig
│   └── observiq-otel-collector
└── windows_service.json
```

I would be in favor of removing service files from archives, but thats Brandon / Corbin's call.

jsirianni: Its not wrong just more steps, we can keep doing it this way.

Dylan-M: I feel they need to be in the archive for pure manual installs. I know these are rare, but they do happen fro time to time.

Dylan-M: Sounds good. I would like to try and maintain consistency with the systemd file wherever possible.

BinaryFissionGames:
Just an overall problem I'm seeing here is that upgrading an agent with BindPlane on one of these systems will fail, since the updater runs systemctl commands e.g.:
https://github.com/observIQ/bindplane-agent/blob/61b4557bad2d427b7a1addd7b38d4052ce3392c4/updater/internal/service/service_linux.go#L76

Is the goal here to fully support upgrading though BindPlane as well?

Dylan-M:
@BinaryFissionGames I've pushed commits for the updater. I'm still writing new tests for it, but I wouldn't mind feedback if you have any.

updater/internal/service/service_linux.go
```diff
@@ -72,20 +72,38 @@ type linuxService struct {
 
 // Start the service
 func (l linuxService) Start() error {
-	//#nosec G204 -- serviceName is not determined by user input
-	cmd := exec.Command("systemctl", "start", l.serviceName)
-	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("running systemctl failed: %w", err)
+	var cmd *exec.Cmd
+	if l.serviceCmdName == "service" {
```
BinaryFissionGames, line 76: I think I'd rather we take an approach here of having to implementations of the Service interface (one for systemd, one for sysv) instead of having these if statements; I think it would be more straight-forward and reduce the number of redundant checks here.

BinaryFissionGames (changes requested):

updater/internal/service/service_linux.go
Dylan-M: This is now done, thank you for the advice.

jsirianni (changes requested):
I installed the package and started it. When I run the status command, I see
```
Status of /etc/init.d/observiq-otel-collector ()observiq-otel-collector dead but pid file exists
```

`ps` does show the process running
```
root      1645  0.4  2.5 1452392 98220 pts/0   Sl   11:35   0:00 /opt/observiq-otel-collector/observiq-otel-collector
```

1. install package
2. start service
3. reboot
4. start service
5. check status

jsirianni (changes requested):
Having issues stopping the service

```
[root@kb-el6-js ~]# service observiq-otel-collector start
[root@kb-el6-js ~]# service observiq-otel-collector status [  OK  ]
Status of /etc/init.d/observiq-otel-collector (2746)observiq-otel-collector (pid  2746) is running...
[root@kb-el6-js ~]# service observiq-otel-collector stop
Shutting down /etc/init.d/observiq-otel-collector: pidof: invalid options on command line!

pidof: invalid options on command line!

                                                           [FAILED]
[root@kb-el6-js ~]# 
```

updater/internal/service/service_linux.go
```diff
@@ -59,39 +78,45 @@ func NewService(logger *zap.Logger, installDir string, opts ...Option) Service {
...
+	setNewSvcFile(string)
+}
+
+type linuxSystemdService struct {
 	// newServiceFilePath is the file path to the new unit file
 	newServiceFilePath string
 	// serviceName is the name of the service
-	serviceName string
+	serviceName    string
+	serviceCmdName string
```
BinaryFissionGames, line 91: I think we can remove this variable now that we have 2 separate structs (service command name should always be "systemctl" for systemd, "service" for sysv)

updater/internal/path/path_linux.go
```diff
@@ -23,3 +28,36 @@ const LinuxInstallDir = "/opt/observiq-otel-collector"
...
+	}
+	_, filename := filepath.Split(path)
+	return filename
+}
+
+// LinuxServiceFilePath returns the full path to the service file
+func LinuxServiceFilePath() string {
+	var path string
+	var err error
+	path, err = exec.LookPath("/usr/lib/systemd/system/observiq-otel-collector.service")
```
BinaryFissionGames, line 54: LookPath looks specifically for executable files, so I think this will never return a path for the systemd service file.

I think instead we should just hard-code the path based on the detected init system, which this function can just take in as a parameter.

BinaryFissionGames (changes requested):

updater/internal/path/path_linux.go
Dylan-M: The shell scripts are executable, and this should work from my understanding. However, if you still think it should be changed that should be pretty quickly handled.

BinaryFissionGames: I'm talking specifically about the systemd service file here.
```sh
$ sudo ls -la /usr/lib/systemd/system/observiq-otel-collector.service
-rw-r--r-- 1 root root 688 Mar 12 12:41 /usr/lib/systemd/system/observiq-otel-collector.service
```

We should either:
1. Hardcode the path based on init system OR
2. Use os.Stat to check for file existence, it's more suitable for this check than exec.LookPath

jsirianni (approved):

updater/internal/path/path_linux.go
```diff
@@ -14,12 +14,53 @@
...
+// LinuxServiceFilePath returns the full path to the service file
+func LinuxServiceFilePath() string {
+	if LinuxServiceCmdName() == "systemctl" {
+		return SystemdFilePath
+	} else if LinuxServiceCmdName() == "systemctl" {
+		return SysVFilePath
+	}
+
+	return SystemdFilePath
+}
```
BinaryFissionGames, line 66: As this is written, looks like it will always redurn systemdFilePath.
```suggestion
func LinuxServiceFilePath() string {
        serviceCmd := LinuxServiceCmdName()
	
	if serviceCmd == "service" {
		return SysVFilePath
	}

	return SystemdFilePath
}
```

BinaryFissionGames (changes requested):

scripts/install/install_unix.sh
```diff
@@ -640,17 +647,32 @@ install_package()
...
+      # We'll want to restart.
+      info "Restarting service..."
+      service observiq-otel-collector restart
+      succeeded
+    else
+      info "Enabling and starting service..."
+      chkconfig observiq-otel-collector on
+      service observiq-otel-collector start
+      succeeded
+    fi
```
BinaryFissionGames, line 675: Output of the install script looks like this:
```
  Enabling and starting service...
    Succeeded!init.d/observiq-otel-collector:              [  OK  ]
```

I think we want to swallow the output of whatever puts `init.d/observiq-otel-collector:              [  OK  ]` to stdout

scripts/install/install_unix.sh
```diff
@@ -640,17 +647,33 @@ install_package()
...
+      info "Enabling service..."
+      systemctl enable --now observiq-otel-collector > /dev/null 2>&1 || error_exit "$LINENO" "Failed to enable service"
+      succeeded
+    fi
   else
-    info "Enabling service..."
-    systemctl enable --now observiq-otel-collector > /dev/null 2>&1 || error_exit "$LINENO" "Failed to enable service"
-    succeeded
+    # shellcheck disable=SC3010
+    if [[ $(service observiq-otel-collector status) = *running* ]]; then
```
BinaryFissionGames, line 665: Can we swap this out to use `case` syntax instead for POSIX shells?

e.g. 
```sh
case "$(service observiq-otel-collector status)" in
  *running*)
    # restart service
    ;;
  *)
    # enable and start service
    ;;
esac
```

BinaryFissionGames (approved):
Working great. I've tested:
1. installing on debian
2. installing on centos6
3. uninstalling on debian
4. uninstalling on centos6
5. upgrading through bindplane for debian
6. upgrading through bindplane for centos6
7. upgrading by re-running the install script for debian
8. upgrading by re-running the install script for centos6

PR #1553, opened 2024-03-13
fix: Additional Polling of Azure Blob Rehydration Receiver

cpheps:
### Proposed Change
After rehydration was completed the Azure Blob Rehydration receiver would continue to poll the Azure Blob API to ask for all blobs in order to see if they should be rehydrated. This would incur Blob egress charges for API calls. As this receiver is given a time window in the past it makes sense that rehydration is not a long running operation.

To stop extra calls if the receiver detects three consecutive polling cycles of no blobs rehydrated then stop attempting to rehydrate.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/azureblobrehydrationreceiver/receiver.go
+++ b/receiver/azureblobrehydrationreceiver/receiver.go
@@ -174,7 +174,7 @@ func (r *rehydrationReceiver) Shutdown(ctx context.Context) error {
 	return err
 }
 
-// emptyPollLimit is the number of sequential empty polling cycles that can
+// emptyPollLimit is the number of consecutive empty polling cycles that can
 // occur before we stop polling.
 const emptyPollLimit = 3
 
@@ -433,7 +433,7 @@ func getStorageClient(ctx context.Context, host component.Host, storageID *compo
 
 // checkBlobCount checks the number of blobs rehydrated and the current state of the
 // empty counter. If zero blobs were rehydrated increment the counter.
-// If there were blobs rehydrated reset the counter as we want to count number of zeros in a row.
+// If there were blobs rehydrated reset the counter as we want to track consecutive zero sized polls.
 func checkBlobCount(numBlobsRehydrated, emptyBlobsCounter int) int {
 	switch {
 	case emptyBlobsCounter == emptyPollLimit: // If we are at the limit return the limit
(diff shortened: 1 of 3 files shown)


receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -415,3 +430,17 @@ func getStorageClient(ctx context.Context, host component.Host, storageID *compo
...
 }
+
+// checkBlobCount checks the number of blobs rehydrated and the current state of the
+// empty counter. If zero blobs were rehydrated increment the counter.
+// If there were blobs rehydrated reset the counter as we want to track consecutive zero sized polls.
+func checkBlobCount(numBlobsRehydrated, emptyBlobsCounter int) int {
+	switch {
+	case emptyBlobsCounter == emptyPollLimit: // If we are at the limit return the limit
+		return emptyPollLimit
+	case numBlobsRehydrated == 0: // If we no blobs rehydrated increment by one
```
Mrod1598, line 441: nit: this comment is confusing

receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -186,22 +190,31 @@ func (r *rehydrationReceiver) scrape() {
...
 		select {
 		case <-r.ctx.Done():
 			return
 		case <-ticker.C:
-			r.rehydrateBlobs(checkpoint, marker)
+			// Polling for blobs has egress charges so we want to stop polling
+			// after we stop finding blobs.
+			if emptyBlobCounter == emptyPollLimit {
+				continue
+			}
```
BinaryFissionGames, line 205: Does it make sense in this case to just return and break out of this loop and kill this goroutine?

Mrod1598 (approved):

receiver/azureblobrehydrationreceiver/receiver.go
Mrod1598: how does the collector handle shutting down individual components? that seems like it might cause problems.

cpheps: I thought about this and was unsure. I think it's probably ok just to break out. The scrape function is private anyways.

BinaryFissionGames: The collector calls the Shutdown function to shut down the component. It shouldn't cause any problems to exit this goroutine, it'll just free up the resources used here (goroutine and ticker).

BinaryFissionGames (approved):

PR #1598, opened 2024-03-28
fix(pluginreceiver): Add buffer to error channel in case shutdown causes error

BinaryFissionGames:
### Proposed Change
* Fix potential lockup on shutdown of the plugin receiver, where it may attempt to emit an error on an unbuffered channel with no listener.
* Make a NewReceiver method that sets internal fields to the correct values (simplifies setup in tests a bit, especially with this channel that must be set up the same everywhere)

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/pluginreceiver/receiver.go
+++ b/receiver/pluginreceiver/receiver.go
@@ -41,6 +41,7 @@ type Receiver struct {
 	serviceErrChan chan error
 }
 
+// NewReceiver creates a new plugin receiver
 func NewReceiver(
 	plugin *Plugin,
 	renderedConfig *RenderedConfig,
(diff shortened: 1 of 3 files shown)


receiver/pluginreceiver/receiver.go
```diff
@@ -83,12 +105,9 @@ func (r *Receiver) Shutdown(ctx context.Context) error {
...
 func (r *Receiver) startService(ctx context.Context, svc Service) error {
-	errChan := make(chan error)
 	go func() {
-		if err := svc.Run(ctx); err != nil {
-			errChan <- err
-		}
-		close(r.doneChan)
+		r.serviceErrChan <- svc.Run(ctx)
+		close(r.serviceErrChan)
 	}()
```
BinaryFissionGames, line 111: this is where the problem was; previously, errChan was scoped to this method, and once this method exited nothing would be able to listen to that channel, which could cause this goroutine to block indefinitely.

cpheps (approved):

PR #1608, opened 2024-04-04
chore: Sign MSI and Artifact Bundle

cpheps:
### Proposed Change
Added signing to the MSI and artifact bundle during the release process

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -526,7 +526,7 @@ checksum:
   name_template: "{{ .ProjectName }}-v{{ .Version }}-SHA256SUMS"
   algorithm: sha256
   extra_files:
-    - glob: "./observiq-otel-collector.msi*"
+    - glob: "./observiq-otel-collector.msi"
 
 # https://goreleaser.com/customization/sign/
 signs:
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -525,6 +525,8 @@ docker_manifests:
 checksum:
   name_template: "{{ .ProjectName }}-v{{ .Version }}-SHA256SUMS"
   algorithm: sha256
+  extra_files:
+    - glob: "./observiq-otel-collector.msi*"
```
jsirianni, line 529: I expected this to fail locally, but it did not. The file just omitted the msi because I do not have it.

I think we can keep the signature file out of the checksum file. The file does not contain other sig files.
```suggestion
  extra_files:
    - glob: "./observiq-otel-collector.msi"
```

jsirianni, line 529: I think we should probably checksum the artifacts archive

```suggestion
  extra_files:
    - glob: "./observiq-otel-collector.msi*"
    - glob: "observiq-otel-collector-${{ github.ref_name }}-artifacts.tar.gz"
```

jsirianni (changes requested):

.goreleaser.yml
cpheps: Oh yeah I don't know why I added it good call.

cpheps: We can't because the archive contains the checksum. That's why we build it after the fact. It's used for offline agent updates.

jsirianni: Of course!

jsirianni (approved):

PR #1612, opened 2024-04-08
[receiver/azureblobrehydrationreceiver] Add poll_timeout option

jsirianni:
### Proposed Change

The call to the Azure Client's `ListBlobs` will hang for what seems like forever if we do not provide a timeout. It is unclear to me why my agent is experiencing this, but we should set a sane timeout.

### Context

This was obvious to me when I added some logging to that counts the number of blobs returned by ListBlobs. I never encountered my log, despite CPU usage being high. When I stopped the service, the agent could log context.Done errors related to the call to ListBlobs. Once I added the timeout, the call started canceling exactly on the timeout interval.

This situation is odd because I am using a known working configuration. It is possible there is an issue on the Azure side, causing the agent to behave in an unexpected way.

### Other Notes

There is no call to `(Config).Validate` at runtime, configuration errors will not be caught outside of tests. I am not sure where it would be appropriate to check the config.

Also, it seems like the receiver (without storage for check pointing) will eventually calm down and stop looking for blobs. I think it should retry until a given range is successful. 

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


receiver/azureblobrehydrationreceiver/config.go
```diff
@@ -51,6 +51,9 @@ type Config struct {
 	// Default value of 1m
 	PollInterval time.Duration `mapstructure:"poll_interval"`
 
+	// PollTimeout is the timeout for the Azure API to scan for blobs.
+	PollTimeout time.Duration `mapstructure:"poll_timeout"`
```
cpheps, line 55: Can we add this parameter to the readme?

cpheps (changes requested):

BinaryFissionGames:
I'll let Corbin re-review, but just wanted to comment on:
> There is no call to (Config).Validate at runtime, configuration errors will not be caught outside of tests. I am not sure where it would be appropriate to check the config.

The validation is done by the pipeline initialization code, so it doesn't need to be called inside the component itself.

When I try to run with a negative timeout with this pr, for instance:
```json
{"level":"fatal","ts":"2024-04-09T09:08:48.613-0400","caller":"collector/main.go:121","msg":"RunService returned error","error":"failed to start service: error during OpAmp connection: collector failed to start: invalid configuration: receivers::azureblobrehydration: poll_timeout must be at least one second","stacktrace":"main.main\n\t/Users/brandonjohnson/git_repos/observiq-otel-collector/cmd/collector/main.go:121\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:267"}
```

cpheps (approved):

PR #1621, opened 2024-04-15
feat: Add nop exporter

kuiperda:
### Proposed Change
https://app.vivifyscrum.com/boards/114813/sprint-backlog/383427/BPOP-3107

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


factories/exporters.go
```diff
@@ -86,6 +87,7 @@ var defaultExporters = []exporter.Factory{
 	loggingexporter.NewFactory(),
 	logzioexporter.NewFactory(),
 	lokiexporter.NewFactory(),
+	nopexporter.NewFactory(),
```
BinaryFissionGames, line 90: I'm guessing this likely conflicts with the nop exporter we have in this list, I think you'll have to remove that one:
`exportertest.NewNopFactory()`

kuiperda: Yeah that's what I'm finding too.

BinaryFissionGames (approved):

PR #1622, opened 2024-04-16
feat(chronicle): Add support for setting log_type through an attribute

Mrod1598:
### Proposed Change

- Add support for setting a log_type through an attribute rather than the config or automatic detection.
- Add additional validation to the config now that we're using the grpc protos.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/chronicleexporter/README.md
+++ b/exporter/chronicleexporter/README.md
@@ -48,7 +48,7 @@ currently supported log types are:
 - sql_server
 
 
-If the `attributes["chronicle_log_type"]` field is present in the log, we will use it's value in the payload instead of the automatic detection or the `log_type` in the config.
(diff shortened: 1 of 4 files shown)


exporter/chronicleexporter/README.md
```diff
@@ -47,6 +47,9 @@ currently supported log types are:
 - windows_event.system
 - sql_server
 
+
+If the `attributes["chronicle_log_type"]` field is present in the log, we will use it's value in the payload instead of the automatic detection or the `log_type` in the config.
```
BinaryFissionGames, line 51:
```suggestion
If the `attributes["chronicle_log_type"]` field is present in the log, we will use its value in the payload instead of the automatic detection or the `log_type` in the config.
```

BinaryFissionGames (changes requested):
Looks good, minor nit on the README

cpheps (approved):

BinaryFissionGames (approved):

PR #1623, opened 2024-04-17
feat(chronicle): Add host metrics support

Mrod1598:
### Proposed Change
Add support for collecting host metrics and sending them to Chronicle. 
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


exporter/chronicleexporter/hostmetrics.go
```diff
@@ -0,0 +1,143 @@
...
+	request := &api.BatchCreateEventsRequest{
+		Batch: &api.EventBatch{
+			Id:        batchID[:],
+			Source:    source,
+			Type:      api.EventBatch_AGENT_STATS,
+			StartTime: cm.startTime,
+			Events: []*api.Event{
+				{
+					Timestamp:      timestamppb.Now(),
+					CollectionTime: timestamppb.Now(),
```
BinaryFissionGames, line 63: Seems like these should be the same timestamp (one call to timestamppb.Now)

exporter/chronicleexporter/README.md
```diff
@@ -22,18 +22,19 @@ This exporter facilitates the sending of logs to Chronicle, which is a security
...
+| `customer_id`          | string            |                                        | `false`  | The customer ID used for sending logs.                                                      |
+| `override_log_type`    | bool              | `true`                                 | `false`  | Whether or not to override the `log_type` in the config with `attributes["log_type"]`       |
+| `namespace`            | string            |                                        | `false`  | User-configured environment namespace to identify the data domain the logs originated from. |
+| `compression`          | string            | `none`                                 | `false`  | The compression type to use when sending logs. valid values are `none` and `gzip`           |
+| `ingestion_labels`     | map[string]string |                                        | `false`  | Key-value pairs of labels to be applied to the logs when sent to chronicle.                 |
+| `collect_host_metrics` | bool              | `true`                                 | `false`  | Enables collecting metrics about the collector's process and log ingestion metrics          |
```
BinaryFissionGames, line 37: About the naming, should we have this be something like `collect_agent_metrics` instead of `collect_host_metrics`? It seems like the scope is just metrics about the agent specifically, and not so much about the host. (e.g. how much cpu time the agent is taking, # of logs the agent has sent, etc)

exporter/chronicleexporter/hostmetrics.go
Mrod1598: updated

exporter/chronicleexporter/config.go
```diff
@@ -67,6 +67,8 @@ type Config struct {
 
 	// IngestionLabels are the labels that will be attached to logs when sent to Chronicle.
 	IngestionLabels map[string]string `mapstructure:"ingestion_labels"`
+
+	CollectAgentMetrics bool `mapstructure:"collect_agent_metrics"`
```
cpheps, line 71: nit: can we add a comment on this?

exporter/chronicleexporter/exporter.go
```diff
@@ -61,17 +67,38 @@ func newExporter(cfg *Config, params exporter.CreateSettings) (*chronicleExporte
...
+		client:      api.NewIngestionServiceV2Client(conn),
+		marshaler:   marshaller,
+		collectorID: collectorID,
+		exporterID:  exporterID,
+		cancel:      cancel,
+	}
+
+	if cfg.CollectAgentMetrics {
+		go exp.startHostMetricsCollection(ctx)
+	}
```
cpheps, line 99: We aren't tracking this goroutine at all. So when `Shutdown` is called there is no guarantee this will be finished by the time it exits. We should use a channel or wait group to signal this has exited.

cpheps (changes requested):

exporter/chronicleexporter/exporter.go
Mrod1598: Added.

exporter/chronicleexporter/config.go
Mrod1598: added.

exporter/chronicleexporter/exporter.go
```diff
@@ -142,3 +175,34 @@ func (ce *chronicleExporter) buildOptions() []grpc.CallOption {
 
 	return opts
 }
+
+func (ce *chronicleExporter) startHostMetricsCollection(ctx context.Context) {
+	ticker := time.NewTicker(5 * time.Minute)
+	defer ticker.Stop()
+
+	ce.wg.Add(1)
```
cpheps, line 183: We should increment the wait group outside of the goroutine. There's a race condition where it may not get incremented before shutdown if caused.

cpheps (changes requested):

exporter/chronicleexporter/exporter.go
Mrod1598: moved

cpheps (approved):

PR #1663, opened 2024-05-21
feat: New snapshot processor based on custom messages

BinaryFissionGames:
### Proposed Change
* Adds a new snapshot processor. This processor is not used by bindplane-agent, but instead is intended to be included in custom distributions built by OCB.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


.github/workflows/checks.yml
```diff
@@ -11,8 +11,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.21"
-          check-latest: true
+          go-version-file: go.mod
```
BinaryFissionGames, line 14: Latest (v1.21.10) fails testing on macOS (macOS's new linker doesn't work, this is only a problem with the `-race` flag, which requires CGO)

internal/report/go.mod
BinaryFissionGames, line 1: Made `internal/report` package to be its own module so that it's importable by the new snapshot processor

internal/report/snapshot/snapshot_buffers.go
```diff
@@ -86,11 +86,15 @@ func (l *LogBuffer) Add(ld plog.Logs) {
...
 }
 
+// LogsMarshaller marshals logs to a byte representation
+type LogsMarshaller interface {
+	MarshalLogs(ld plog.Logs) ([]byte, error)
+}
+
 // ConstructPayload condenses the buffer and serializes to protobuf
-func (l *LogBuffer) ConstructPayload(searchQuery *string, minimumTimestamp *time.Time) ([]byte, error) {
+func (l *LogBuffer) ConstructPayload(logsMarshaler LogsMarshaller, searchQuery *string, minimumTimestamp *time.Time) ([]byte, error) {
```
BinaryFissionGames, line 95: This change allows the caller to decide how the buffer is marshalled (in this case, it's so the original method for snapshots can use protobuf marshalling, but the new stuff can use json marshalling)

processor/snapshotprocessor/factory.go
```diff
@@ -0,0 +1,127 @@
...
+		processors[set.ID] = sp
+	}
+
+	return sp
+}
+
+// processors is a map of component.ID to an instance of snapshot processor.
+// It is used so that only one instance of a particular snapshot processor exists, even if it's included
+// across multiple pipelines/signal types.
+var processors = map[component.ID]*snapshotProcessor{}
```
djaglowski, line 127: It's possible that components would be instantiated concurrently so I would recommend making this safe for concurrent access.

internal/report/snapshot/snapshot_buffers.go
```diff
@@ -86,11 +86,15 @@ func (l *LogBuffer) Add(ld plog.Logs) {
 	}
 }
 
+// LogsMarshaller marshals logs to a byte representation
+type LogsMarshaller interface {
+	MarshalLogs(ld plog.Logs) ([]byte, error)
+}
```
djaglowski, line 92: Any reason not to just make this a func?
```suggestion
type MarshalLogs func(plog.Logs) ([]byte, error)
```

processor/snapshotprocessor/README.md
```diff
@@ -0,0 +1,57 @@
...
+The snapshot processor is used in custom distributions of the collector to provide snapshot functionality in BindPlane. It is not currently included in the official `bindplane-agent`.
+## Supported pipelines
+
+- Logs
+- Metrics
+- Traces
+
+## How it works
+
+1. The user configured the processor in the pipeline.
```
djaglowski, line 12:
```suggestion
1. The user configures the processor in one or more pipelines.
```

processor/snapshotprocessor/README.md
```diff
@@ -0,0 +1,57 @@
...
+2. Whenever telemetry passes through the pipeline, it is copied and stored temporarily in an internal buffer.
```
djaglowski, line 13:
```suggestion
2. Whenever telemetry passes through the processor, it is copied and stored temporarily in an internal buffer.
```

processor/snapshotprocessor/README.md
```diff
@@ -0,0 +1,57 @@
...
+3. An OpAMP server is able to use a custom message to request the contents of the internal buffer, in order to view a snapshot of the telemetry flowing through the collector.
+
+## Configuration
+
+| Field   | Type   | Default | Required | Description                                                            |
+|---------|--------|---------|----------|------------------------------------------------------------------------|
+| enabled | bool   | `true`  | `false`  | Whether the snapshot processor is enabled or not.                      |
```
djaglowski, line 20: What is the purpose of it being disabled instead of just omitted from the config?

processor/snapshotprocessor/README.md
```diff
@@ -0,0 +1,57 @@
...
+| opamp   | string | `opamp` | `false`  | Specifies the name of the opamp extension for sending custom messages. |
```
djaglowski, line 21: What is the purpose if there is no opamp extension?

processor/snapshotprocessor/factory.go
```diff
@@ -0,0 +1,127 @@
...
+		sp = newSnapshotProcessor(set.Logger, cfg, set.ID)
+		processors[set.ID] = sp
+	}
+
+	return sp
+}
+
+// processors is a map of component.ID to an instance of snapshot processor.
+// It is used so that only one instance of a particular snapshot processor exists, even if it's included
+// across multiple pipelines/signal types.
```
djaglowski, line 126: What is the value in mixing data from multiple pipelines. I see the snapshot request is asking for a specific data type, so does this really only serve to combine data from multiple pipelines of the same type? If so, what's the value in that?

processor/snapshotprocessor/README.md
BinaryFissionGames: There really isn't a purpose, I've kept it this way to be 100% compatible with the config of the previous processor.

internal/report/snapshot/snapshot_buffers.go
BinaryFissionGames: We can, it's basically equivalent, but it makes it just slightly more awkward to use:

Current:
```go
payload, err := logBuffer.ConstructPayload(&plog.ProtoMarshaler{}, nil, nil)
```

As function:
```go
payload, err := logBuffer.ConstructPayload((&plog.ProtoMarshaler{}).MarshalLogs, nil, nil)
```

I don't really mind either way, so if you think there's a benefit to the function approach, I'm happy to change it.

djaglowski: Ah, I see the motivation then.

processor/snapshotprocessor/README.md
BinaryFissionGames: There is none. I think I was treating the required field as "you need to specify this field" and not "you can't leave this field explicitly empty". I think it makes more sense to mark as required.

processor/snapshotprocessor/factory.go
BinaryFissionGames: Yeah, good catch. We technically could use different structs for traces/logs/metrics, as long as you always get the same processor for the unique combination of ID + telemetry type. Both the current way (one struct for all) and splitting it will work.

Do you think there's value in splitting it up into separate structs, one for each telemetry type? Just wondering if there's any benefit to splitting it by telemetry type instead of keeping it as the one struct.

BinaryFissionGames: Sorry, I think I misinterpreted the question.

Yes, we want and need that behavior, based on how BindPlane currently works.

Just as an example, imagine you have a setup with two pipelines like this (this is what BindPlane basically does when you have 1 source, 2 destinations):
```mermaid
block-beta
columns 3
  block:pipeline1:3
    columns 3
    source1["Source 1"]
    processor1<["Snapshot"]>(right)
    dest1["Destination 1"]
  end
  block:pipeline2:3
    columns 3
    source1Again["Source 1"]
    processor2<["Snapshot"]>(right)
    dest2["Destination 2"]
  end
```

In this case, the snapshot processor is shared between the two pipelines. BindPlane will send a custom message to get the contents of the buffer. Currently, BindPlane expects a single message to be sent back with the combined buffer (e.g. the buffer is assumed shared between the pipelines).

If they aren't the same, BindPlane will get multiple messages and the first one will "win", leading to inconsistent snapshots.

It's not impossible to envision BindPlane working differently, I just think it makes it more complicated, because BindPlane would need to now have specialized logic to wait and recombine the buffers after the fact (BindPlane always wants the telemetry from every pipeline the processor is in), when I think it's much simpler if the buffers are just combined to begin wit. This is also in-line with the existing behavior. 

In addition, there doesn't seem to be an advantage to having them apart. BindPlane would not be able to single out any particular instance, meaning a request for `snapshotprocessor/1` would always trigger a response from every `snapshotprocessor/1` either way, so why not combine them into one message by having them have a single combined buffer?

BinaryFissionGames: For reference, this is what the service section for config generated by BindPlane looks like:
```
service:
    extensions:
        - bindplane
        - health_check
        - opamp
    pipelines:
        logs/source0_01HY1DFHNQZXTB7GMCMSF7ZR4A__nop-0:
            exporters:
                - nop/nop
            processors:
                - snapshotprocessor/_s0_source0_01HY1DFHNQZXTB7GMCMSF7ZR4A
                - snapshotprocessor/_d0_nop-0
                - snapshotprocessor
            receivers:
                - otlp/source0_01HY1DFHNQZXTB7GMCMSF7ZR4A
        metrics/source0_01HY1DFHNQZXTB7GMCMSF7ZR4A__nop-0:
            exporters:
                - nop/nop
            processors:
                - snapshotprocessor/_s0_source0_01HY1DFHNQZXTB7GMCMSF7ZR4A
                - snapshotprocessor/_d0_nop-0
                - snapshotprocessor
            receivers:
                - otlp/source0_01HY1DFHNQZXTB7GMCMSF7ZR4A
        metrics/source1_01HY1FDMKP8ACCKE9VXRDPGR9P__nop-0:
            exporters:
                - nop/nop
            processors:
                - resourcedetection/source1_01HY1FDMKP8ACCKE9VXRDPGR9P
                - snapshotprocessor/_s0_source1_01HY1FDMKP8ACCKE9VXRDPGR9P
                - snapshotprocessor/_d0_nop-0
                - snapshotprocessor
            receivers:
                - hostmetrics/source1_01HY1FDMKP8ACCKE9VXRDPGR9P
        metrics/source2_01HY1HEABD06HX7Y5H8RDS23HR__nop-0:
            exporters:
                - nop/nop
            processors:
                - resourcedetection/source2_01HY1HEABD06HX7Y5H8RDS23HR
                - snapshotprocessor/_s0_source2_01HY1HEABD06HX7Y5H8RDS23HR
                - snapshotprocessor/_d0_nop-0
                - snapshotprocessor
            receivers:
                - hostmetrics/source2_01HY1HEABD06HX7Y5H8RDS23HR
        traces/source0_01HY1DFHNQZXTB7GMCMSF7ZR4A__nop-0:
            exporters:
                - nop/nop
            processors:
                - snapshotprocessor/_s0_source0_01HY1DFHNQZXTB7GMCMSF7ZR4A
                - snapshotprocessor/_d0_nop-0
                - snapshotprocessor
            receivers:
                - otlp/source0_01HY1DFHNQZXTB7GMCMSF7ZR4A
```

You can see that `snapshotprocessor` is in every pipeline, and BindPlane expects one response that returns a shared buffer between all these pipelines.

djaglowski (commented):
lgtm

dpaasman00 (approved):
looks good and was working well when testing with the bindplane-op pr

andykellr (approved):

PR #1691, opened 2024-06-19
feat: custom message based throughput measurements

BinaryFissionGames:
### Proposed Change
* Adds custom message based throughput measurements.

These measurements are emitted over an OpAMP connection.

In bindplane-agent, throughput measurements are associated with a static registry, which holds all the throughput measurements. The measurements in this repository is reported to BindPlane over custom messages if the BindPlane server supports the capability.

In an OCB build of the collector, an optional bindplane extension may be specified to use as a registry for throughput metrics. In this case, the bindplane extension will use an opamp extension to report throughput metrics to the BindPlane server.

Regardless of either of these, the throughputmeasurements processor still registers metrics with the collector's self-metrics, so the metrics remain available over prometheus, which makes this change backwards compatible.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dpaasman00 (approved):
Tested against BindPlane with our agent and OCB collector and it was all working well.

PR #1692, opened 2024-06-20
chore: release test ease of use

BinaryFissionGames:
### Proposed Change
* Use `VERSION` make file var for GORELEASER_CURRENT_TAG in release-test
  * This lets you set the version of the build by doing `VERSION=v1.xx.x make release-test`
* Create a dummy windows MSI if one isn't present
  * This is required for checksums, but usually it's not worth going through the effort of actually getting one of these built. So a dummy one will be created if one doesn't exist

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -219,7 +219,8 @@ release-prep:
 # Build and sign, skip release and ignore dirty git tree
 .PHONY: release-test
 release-test:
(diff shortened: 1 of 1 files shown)


jsirianni (approved):
Works great

PR #1706, opened 2024-06-26
deps: windows event log updates

dpaasman00:
### Proposed Change

Adds changes to the Windows Event Log receiver and Stanza that aren't yet upstream that are made [here]( https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33601)

Uses commit hash `ed29be57a76fb4938324b184f50b5ecb16e76628`

Tested by building for windows and moving file to a windows vm. There confirmed the agent was able to start with a config that uses newly added windows config options indicating it was using the changes present in the above PR

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


antonblock (approved):
Other stanza operators working without any change

PR #1756, opened 2024-07-18
feat: Updated MacOS install script

dpaasman00:
### Proposed Change

Updates MacOS install script and service file to run with opamp supervisor

To test, run `make release-test` and then run `/scripts/install/install_macos.sh -f /path/to/darwin/tarball ...` with the rest of the arguments like endpoint and secret key (i just copied these from the cmd in the BP UI). 

You should see the agent is connected.

A couple notes:

- the agent log has a lot of errors about not being able to connect to the opamp server, but rolling out a config to the agent works fine and fixes it... not sure why this happens yet

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/install/install_macos.sh
+++ b/scripts/install/install_macos.sh
@@ -24,6 +24,7 @@ PREREQS="printf sed uname tr find grep"
 TMP_DIR="${TMPDIR:-"/tmp/"}observiq-otel-collector" # Allow this to be overriden by cannonical TMPDIR env var
 INSTALL_DIR="/opt/observiq-otel-collector"
 MANAGEMENT_YML_PATH="$INSTALL_DIR/manager.yaml"
+SUPERVISOR_YML_PATH="$INSTALL_DIR/supervisor-config.yaml"
 SCRIPT_NAME="$0"
 INDENT_WIDTH='  '
 indent=""
@@ -534,6 +535,7 @@ install_package()
   # If an endpoint was specified, we need to write the manager.yaml
   if [ -n "$OPAMP_ENDPOINT" ]; then
     create_manager_yml "$MANAGEMENT_YML_PATH"
+    create_supervisor_config "$SUPERVISOR_YML_PATH"
   fi
 
   # Install jmx jar
@@ -582,6 +584,38 @@ create_manager_yml()
     command printf 'endpoint: "%s"\n' "$OPAMP_ENDPOINT" > "$manager_yml_path"
     [ -n "$OPAMP_LABELS" ] && command printf 'labels: "%s"\n' "$OPAMP_LABELS" >> "$manager_yml_path"
     [ -n "$OPAMP_SECRET_KEY" ] && command printf 'secret_key: "%s"\n' "$OPAMP_SECRET_KEY" >> "$manager_yml_path"
+    succeeded
+  fi
+}
+
+
+create_supervisor_config()
+{
+  supervisor_yml_path="$1"
+  if [ ! -f "$supervisor_yml_path" ]; then
+    info "Creating supervisor config..."
+
+    # Note here: We create the file and change permissions of the file here BEFORE writing info to it
+    # We do this because the file may contain a secret key, so we want 0 window when the
+    # file is readable by anyone other than root
+    command printf '' >> "$supervisor_yml_path"
+    chmod 0600 "$supervisor_yml_path"
+
+    command printf 'server:\n' > "$supervisor_yml_path"
+    command printf '  endpoint: "%s"\n' "$OPAMP_ENDPOINT" >> "$supervisor_yml_path"
+    command printf '  headers:\n' >> "$supervisor_yml_path"
+    [ -n "$OPAMP_SECRET_KEY" ] && command printf '    Authorization: "Secret-Key %s"\n' "$OPAMP_SECRET_KEY" >> "$supervisor_yml_path"
+    [ -n "$OPAMP_LABELS" ] && command printf '    X-Bindplane-Attribute: "%s"\n' "$OPAMP_LABELS" >> "$supervisor_yml_path"
+    command printf '  tls:\n' >> "$supervisor_yml_path"
+    command printf '    insecure: true\n' >> "$supervisor_yml_path"
+    command printf '    insecure_skip_verify: true\n' >> "$supervisor_yml_path"
+    command printf 'capabilities:\n' >> "$supervisor_yml_path"
+    command printf '  accepts_remote_config: true\n' >> "$supervisor_yml_path"
+    command printf '  reports_remote_config: true\n' >> "$supervisor_yml_path"
+    command printf 'agent:\n' >> "$supervisor_yml_path"
+    command printf '  executable: "%s"\n' "$INSTALL_DIR/observiq-otel-collector" >> "$supervisor_yml_path"
+    command printf 'storage:\n' >> "$supervisor_yml_path"
+    command printf '  directory: "%s"\n' "$INSTALL_DIR/storage" >> "$supervisor_yml_path"
     succeeded
   fi
 }
--- a/scripts/install/install_macos.sh
+++ b/scripts/install/install_macos.sh
@@ -179,7 +179,10 @@ Usage:
       Defines the base of the download URL as '{base_url}/v{version}/observiq-otel-collector-v{version}-darwin-{os_arch}.tar.gz'.
       If not provided, this will default to '$DOWNLOAD_BASE'.
       Example: '-b http://my.domain.org/observiq-otel-collector/binaries' will be used as the base of the download URL.
-   
+    
+  $(fg_yellow '-f, --file')
+      Install Agent from a local file instead of downloading from a URL.
+
   $(fg_yellow '-e, --endpoint')
       Defines the endpoint of an OpAMP compatible management server for this agent install.
       This parameter may also be provided through the ENDPOINT environment variable.
@@ -369,7 +372,14 @@ setup_installation()
   increase_indent
 
   set_os_arch
-  set_download_urls
+
+  if [ -z "$package_path" ]; then
+    set_download_urls
+    out_file_path="$TMP_DIR/observiq-otel-collector.tar.gz"
+  else 
+    out_file_path="$package_path"
+  fi
+
   set_opamp_endpoint
   set_opamp_labels
   set_opamp_secret_key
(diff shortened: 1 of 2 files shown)


scripts/install/install_macos.sh
```diff
@@ -566,22 +572,34 @@ install_package()
...
-    [ -n "$OPAMP_LABELS" ] && command printf 'labels: "%s"\n' "$OPAMP_LABELS" >> "$manager_yml_path"
-    [ -n "$OPAMP_SECRET_KEY" ] && command printf 'secret_key: "%s"\n' "$OPAMP_SECRET_KEY" >> "$manager_yml_path"
+    command printf '' >> "$supervisor_yml_path"
+    chmod 0600 "$supervisor_yml_path"
+
+    command printf 'server:\n' > "$supervisor_yml_path"
+    command printf '  endpoint: "%s"\n' "$OPAMP_ENDPOINT" >> "$supervisor_yml_path"
+    command printf '  headers:\n' >> "$supervisor_yml_path"
+    [ -n "$OPAMP_SECRET_KEY" ] && command printf '    Authorization: "Secret-Key %s"\n' "$OPAMP_SECRET_KEY" >> "$supervisor_yml_path"
+    [ -n "$OPAMP_LABELS" ] && command printf '    X-Bindplane-Attribute: "%s"\n' "$OPAMP_LABELS" >> "$supervisor_yml_path"
```
BinaryFissionGames, line 591: Better way to accomplish this might be to add a snippet like this:
```yaml
agent:
  description:
    non_identifying_attributes:
      service.labels: "$labels"
```

I think this would make the install_id work (haven't tested yet but assuming there's still trouble with it)

dpaasman00: Tried this but still no luck

scripts/install/install_macos.sh
```diff
@@ -507,8 +520,8 @@ install_package()
   info "Copying artifacts to install directory..."
   increase_indent
 
-  # This find command gets a list of all artifacts paths except config.yaml, logging.yaml, or opentelemetry-java-contrib-jmx-metrics.jar
-  FILES=$(cd "$TMP_DIR/artifacts"; find "." -type f -not \( -name config.yaml -or -name logging.yaml -or -name opentelemetry-java-contrib-jmx-metrics.jar \))
+  # This find command gets a list of all artifacts paths except logging.yaml, or opentelemetry-java-contrib-jmx-metrics.jar
+  FILES=$(cd "$TMP_DIR/artifacts"; find "." -type f -not \( -name -or -name logging.yaml -or -name opentelemetry-java-contrib-jmx-metrics.jar \))
```
BinaryFissionGames, line 524: In think you could just remove the -name and -or here
```suggestion
  FILES=$(cd "$TMP_DIR/artifacts"; find "." -type f -not \( -name logging.yaml -or -name opentelemetry-java-contrib-jmx-metrics.jar \))
```

scripts/install/install_macos.sh
```diff
@@ -566,22 +572,37 @@ install_package()
...
+    command printf '  accepts_remote_config: true\n' >> "$supervisor_yml_path"
+    command printf '  reports_remote_config: true\n' >> "$supervisor_yml_path"
+    command printf 'agent:\n' >> "$supervisor_yml_path"
+    # TODO(dakota): Add logging config option when supervisor suppports it
+    command printf '  executable: "%s"\n' "$INSTALL_DIR/observiq-otel-collector" >> "$supervisor_yml_path"
+    command printf '  description:\n' >> "$supervisor_yml_path"
+    command printf '    non_identifying_attributes:\n' >> "$supervisor_yml_path"
+    [ -n "$OPAMP_LABELS" ] && command printf '      service.labels: "%s"\n' "$OPAMP_LABELS" >> "$supervisor_yml_path"
+    command printf 'storage:\n' >> "$supervisor_yml_path"
+    command printf '  directory: "%s"\n' "$INSTALL_DIR/storage" >> "$supervisor_yml_path"
```
BinaryFissionGames, line 605: Let's have this storage directory by something other than `storage`; My thinking is I don't want it to potentially conflict with other storage for OTel components (e.g. filestorage extension).

Maybe we could just have this go to a separate `supervisor_storage` directory?

scripts/install/install_macos.sh
```diff
@@ -592,11 +613,11 @@ display_results()
...
     banner 'Information'
     increase_indent
-    info "Agent Home:         $(fg_cyan "$INSTALL_DIR")$(reset)"
-    info "Agent Config:       $(fg_cyan "$INSTALL_DIR/config.yaml")$(reset)"
-    info "Start Command:      $(fg_cyan "sudo launchctl load /Library/LaunchDaemons/$SERVICE_NAME.plist")$(reset)"
-    info "Stop Command:       $(fg_cyan "sudo launchctl unload /Library/LaunchDaemons/$SERVICE_NAME.plist")$(reset)"
-    info "Logs Command:       $(fg_cyan "sudo tail -F $INSTALL_DIR/log/collector.log")$(reset)"
+    info "Agent Home:                    $(fg_cyan "$INSTALL_DIR")$(reset)"
+    info "Agent Config:                  $(fg_cyan "$INSTALL_DIR/effective.yaml")$(reset)"
+    info "Agent Logs Command:            $(fg_cyan "sudo tail -F $INSTALL_DIR/agent.log")$(reset)"
```
BinaryFissionGames, line 618: This has changed, probably will change again too 😅 
```suggestion
    info "Agent Config:                  $(fg_cyan "$INSTALL_DIR/supervisor_storage/effective.yaml")$(reset)"
    info "Agent Logs Command:            $(fg_cyan "sudo tail -F $INSTALL_DIR/supervisor_storage/agent.log")$(reset)"
```

scripts/install/install_macos.sh
```diff
@@ -631,14 +652,14 @@ uninstall()
   rm -f "/Library/LaunchDaemons/$SERVICE_NAME.plist" || error_exit "$LINENO" "Failed to remove service file /Library/LaunchDaemons/$SERVICE_NAME.plist"
   succeeded
 
-  info "Backing up config.yaml to config.bak.yaml"
-  cp "$INSTALL_DIR/config.yaml" "$INSTALL_DIR/config.bak.yaml" || error_exit "$LINENO" "Failed to backup config.yaml to config.bak.yaml"
+  info "Backing up effective.yaml to effective.bak.yaml"
+  cp "$INSTALL_DIR/effective.yaml" "$INSTALL_DIR/effective.bak.yaml" || error_exit "$LINENO" "Failed to backup effective.yaml to effective.bak.yaml"
   succeeded
```
BinaryFissionGames, line 657: Let's change the path for effective.yaml here to be under the supervisor storage dir, that recently had changed.

BinaryFissionGames: Actually, let's just remove this step. The effective.yaml is just given from BPOP, it's never something the user had configured right now. We can add it back in if we end up using it for standalone installs.

BinaryFissionGames (changes requested):
This is working pretty well for me right now, nice work.

scripts/install/install_macos.sh
BinaryFissionGames: The labels are working for me, but only on the unreleased versions (v1.68.0+) -- this is due to the recent changes to labels we are making BindPlane side. So I think this is good where it is right now.

scripts/install/install_macos.sh
```diff
@@ -24,6 +24,7 @@ PREREQS="printf sed uname tr find grep"
 TMP_DIR="${TMPDIR:-"/tmp/"}observiq-otel-collector" # Allow this to be overriden by cannonical TMPDIR env var
 INSTALL_DIR="/opt/observiq-otel-collector"
 MANAGEMENT_YML_PATH="$INSTALL_DIR/manager.yaml"
```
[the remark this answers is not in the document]
dpaasman00: shouldn't be lol

scripts/install/install_macos.sh
dpaasman00: 🤦

scripts/install/install_macos.sh
dpaasman00: sounds good

scripts/install/install_macos.sh
```diff
@@ -566,22 +571,38 @@ install_package()
...
-    [ -n "$OPAMP_SECRET_KEY" ] && command printf 'secret_key: "%s"\n' "$OPAMP_SECRET_KEY" >> "$manager_yml_path"
+  supervisor_yml_path="$1"
+  if [ ! -f "$supervisor_yml_path" ]; then
+    info "Creating supervisor config..."
+
+        # Note here: We create the file and change permissions of the file here BEFORE writing info to it.
+    # We do this because the file contains the secret key.
+    # We do not want the file readable by anyone other than root/obseriq-otel-collector.
+    command printf '' >> "$supervisor_yml_path"
+    chown observiq-otel-collector:observiq-otel-collector "$supervisor_yml_path"
```
BinaryFissionGames, line 584: I get this: 
`chown: observiq-otel-collector: illegal group name`

If we want this to have a separate group/user, we'll need to create the group/user first.
Also, if we're gonna do this, we should make everything in this directory belong to this user/group (chown -R)

We can just remove this line for now and come back to it later too, we've had everything belonging to root here for quite a while.

BinaryFissionGames (changes requested):

scripts/install/install_macos.sh
dpaasman00: That works for me. This was something I pulled over from the linux script, but that has the luxury of a pre-install script that sets/creates the group. I'll take it out

scripts/install/install_macos.sh
```diff
@@ -631,22 +628,9 @@ uninstall()
...
-  FILES=$(cd "$INSTALL_DIR"; find "." -not \( -name config.bak.yaml -or -name "." \))
-  for f in $FILES
-  do
-    rm -rf "${INSTALL_DIR:?}/$f" || error_exit "$LINENO" "Failed to remove artifact ${INSTALL_DIR:?}/$f"
-  done
-
-  # Copy the old config to a backup. This is similar to how RPM handles this.
-  # This way, the file is recoverable if the uninstall was somehow accidental,
-  # but if a new install occurs, the default config will still be used.
+  rm -rf /opt/observiq-otel-collector || error_exit "$LINENO" "Failed to remove /opt/observiq-otel-collector"
```
dpaasman00, line 633: Since we no longer care to keep any files, are you okay with this approach instead? I can change it back if not

BinaryFissionGames: Yep, seems good to me.

BinaryFissionGames (approved):
Working great for me.

PR #1769, opened 2024-07-23
feat(container): Minimal scratch based container image

jsirianni:
### Proposed Change

Added a new container build based on [scratch](https://hub.docker.com/_/scratch). Scratch images are empty, containing nothing. The final image will only contain files explicitly copied into the image during the build.

This change is security motivated. Users who do not require systemd (journald receiver) and Java (jmx receiver) can decide to use this image, which has a reduced attack surface. The final image contains a single binary (agent binary), does not contain a shell or any other utilizes that an attacker would look for.

The images are tagged with `major.minor.patch-minimal`. I opted to omit `latest` and did not bother with `major` and `major.minor`. The final images are somewhat small, ~ 240MB vs ~500 for the full image.

```
observiq/bindplane-agent-arm64                                               1.56.0-minimal      235MB
ghcr.io/observiq/bindplane-agent-arm64                                       1.56.0-minimal      235MB
us-central1-docker.pkg.dev/observiq-containers/agent/bindplane-agent-arm64   1.56.0-minimal      235MB
observiq/bindplane-agent-amd64                                               1.56.0-minimal      244MB
ghcr.io/observiq/bindplane-agent-amd64                                       1.56.0-minimal      244MB
us-central1-docker.pkg.dev/observiq-containers/agent/bindplane-agent-amd64   1.56.0-minimal      244MB
```

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -413,6 +413,21 @@ dockers:
       - release_deps/opentelemetry-java-contrib-jmx-metrics.jar
 
 docker_manifests:
+  - name_template: "observiq/bindplane-agent:{{ .Major }}.{{ .Minor }}.{{ .Patch }}-minimal"
+    image_templates:
+      - "observiq/bindplane-agent-amd64:{{ .Major }}.{{ .Minor }}.{{ .Patch }}-minimal"
+      - "observiq/bindplane-agent-arm64:{{ .Major }}.{{ .Minor }}.{{ .Patch }}-minimal"
+    skip_push: false
+  - name_template: "ghcr.io/observiq/bindplane-agent:{{ .Major }}.{{ .Minor }}.{{ .Patch }}-minimal"
+    image_templates:
+      - "ghcr.io/observiq/bindplane-agent-amd64:{{ .Major }}.{{ .Minor }}.{{ .Patch }}-minimal"
+      - "ghcr.io/observiq/bindplane-agent-arm64:{{ .Major }}.{{ .Minor }}.{{ .Patch }}-minimal"
+    skip_push: false
+  - name_template: "us-central1-docker.pkg.dev/observiq-containers/agent/bindplane-agent:{{ .Major }}.{{ .Minor }}.{{ .Patch }}-minimal"
+    image_templates:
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -369,6 +413,21 @@ dockers:
       - release_deps/opentelemetry-java-contrib-jmx-metrics.jar
 
 docker_manifests:
```
jsirianni, line 415: This section cannot be tested unless we issue a release. We should ensure the image_template names line up with the image names used in the `dockers` step.

jsirianni:
I tested against BindPlane using this command.

```bash
make release-test

docker run -it \
  -e OPAMP_ENDPOINT=wss://app.bindplane.com/v1/opamp \
  -e OPAMP_SECRET_KEY=<my secret key> \
  -e OPAMP_LABELS=configuration=test \
observiq/bindplane-agent-amd64:1.56.0-minimal
```

The agent appears in BindPlane, receives a config via rollout.

jsirianni:
When testing, I ended up using telemetry generator. Host metrics will not work, likely because the underlying utilities do not exist within the container.

.goreleaser.yml
```diff
@@ -222,6 +222,50 @@ nfpms:
...
+      - "--label=created={{.Date}}"
+      - "--label=title={{.ProjectName}}"
+      - "--label=revision={{.FullCommit}}"
+      - "--label=version={{.Version}}"
+      - "--platform=linux/amd64"
+    extra_files:
+      - plugins
+      - config/example.yaml
+      - config/logging.stdout.yaml
+      - LICENSE
```
dpaasman00, line 246: For these 'extra_files' I'd use the files that are present in the `release_deps` directory that's made by the release prep before hook. However I'm noticing that the other pre-existing docker builds don't use those files either and I'm not sure if there's a reason for that? @BinaryFissionGames Do you know?

dpaasman00 (commented):
One question

.goreleaser.yml
dpaasman00: Yea that works for me

dpaasman00 (approved):
Working for me

PR #1770, opened 2024-07-24
chore(docker): Use release_deps when copying files to container images

jsirianni:
### Proposed Change

- Copy plugins from release_deps
- Copy base config from release_deps
- New: Copy VERSION.txt from release_deps

The logging config and LICENSE will continue to be copied as is. The logging config was built specifically for the container build process.

This is a follow up PR for https://github.com/observIQ/bindplane-agent/pull/1769

Tested by deploying BindPlane managed agents and updating them to use the Ubuntu based image produced by this branch.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -240,8 +240,9 @@ dockers:
       - "--label=version={{.Version}}"
       - "--platform=linux/amd64"
     extra_files:
-      - plugins
-      - config/example.yaml
+      - release_deps/VERSION.txt
+      - release_deps/plugins
+      - release_deps/config.yaml
       - config/logging.stdout.yaml
       - LICENSE
   - id: scratch-arm64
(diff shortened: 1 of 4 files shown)


.goreleaser.yml
```diff
@@ -240,8 +240,9 @@ dockers:
       - "--label=version={{.Version}}"
       - "--platform=linux/amd64"
     extra_files:
-      - plugins
-      - config/example.yaml
+      - release_deps/VERSION.txt
+      - release_deps/plugins
+      - release_deps/config.yaml
       - config/logging.stdout.yaml
```
dpaasman00, line 246: Do you think it's worthwhile to copy over `logging.stdout.yaml` into the release_deps folder to follow the same principle or leave it as is? I don't know that I have a preference, just not sure what makes sense here

jsirianni: I am indifferent. It is not being modified or generated. It is only used by the container build process.

dpaasman00 (approved):

PR #1775, opened 2024-07-31
feat: Windows installation and packaging

dpaasman00:
### Proposed Change

Updates the windows installer to support supervisor and run it as a service

For testing, I ran a windows vm and RDP into it. The supervisor doesn't run well/correctly on windows atm so it will quit shortly after starting. This causes some weirdness with waiting for the service to start. Otherwise the UI wizard should reflect necessary changes, notably the secret key being required. 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


windows/install/generate-supervisor-yaml.bat
```diff
@@ -0,0 +1,71 @@
...
+set "reportsRemoteCfgField=  reports_remote_config: true"
+
+set "agentField=agent:"
+set "executablePathField=  executable: '%agentBinary%'"
+set "descriptionField=  description:"
+set "nonIdentifyingAttributesField=    non_identifying_attributes:"
+set "serviceLabelsField=      service.labels: "%labels%""
+
+set "storageField=storage:"
+set "directoryField=  directory: '%install_dir%storage'"
```
BinaryFissionGames, line 48: Let's make this supervisor_storage here too

windows/wix.json
```diff
@@ -9,21 +9,20 @@
...
     {
-      "path": "updater.exe"
+      "path": "observiq-otel-collector.exe"
     },
     {
-      "path": "config.yaml",
-      "never_overwrite": true
+      "path": "updater.exe"
     },
     {
```
BinaryFissionGames, line 27: logging.yaml should be removed from here

BinaryFissionGames (changes requested):
Few notes about folder structurre; Everything here should end up working, but we still need to add windows service support to the supervisor binary (I had forgotten about this, but if the binary doesn't implement windows specific service handling code, it will fail to start when run as a service).

When running as a non-service, stuff here seems to work great though.

BinaryFissionGames (approved):
This seems like it should work great once we resolve the service issue:
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34774

I'm OK with merging this as-is, since it seems like all the work we can do here is done. But we'll want to make sure to retest once that issue is solved.

PR #1785, opened 2024-08-06
fix: Remove length limit from component name

BinaryFissionGames:
### Proposed Change
Removes the length limit from the component name.

You can see the diff to v0.106.x here: https://github.com/open-telemetry/opentelemetry-collector/compare/release/v0.106.x..observIQ:opentelemetry-collector:component-length-restriction-remove-v0.106.1?expand=1

Tested this against bindplane with a long-named source from saved-sources. It fails on (unreleased) v1.57.0, but works with this change.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dpaasman00 (approved):
Yep, this fixes the issue for me

PR #1809, opened 2024-08-20
chore: Remove logging.yaml

dpaasman00:
### Proposed Change

Cleanup references to `logging.yaml`

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -110,9 +110,6 @@ archives:
       - src: release_deps/opentelemetry-java-contrib-jmx-metrics.jar
         dst: "."
         strip_parent: true
-      - src: release_deps/logging.yaml
-        dst: "."
-        strip_parent: true
       - src: release_deps/plugins/*
         dst: plugins
         strip_parent: true
@@ -158,14 +155,6 @@ nfpms:
           mode: 0755
           owner: observiq-otel-collector
           group: observiq-otel-collector
-      # TODO: (dakota) evaluate this file once supervisor has support for configuring agent logging
-      - src: release_deps/logging.yaml
-        dst: /opt/observiq-otel-collector/logging.yaml
-        file_info:
-          mode: 0640
-          owner: observiq-otel-collector
-          group: observiq-otel-collector
-        type: config|noreplace
       - src: LICENSE
         dst: /opt/observiq-otel-collector/LICENSE
         file_info:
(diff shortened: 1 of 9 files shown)


BinaryFissionGames (commented):
Seems like we could also remove the `config/logging.*` files

.goreleaser.yml
```diff
@@ -263,7 +252,6 @@ dockers:
     extra_files:
       - release_deps/VERSION.txt
       - release_deps/plugins
-      - config/logging.stdout.yaml
```
jsirianni, line 266: Will containers continue to log to stdout / error?

.goreleaser.yml
```diff
@@ -158,14 +155,6 @@ nfpms:
           mode: 0755
           owner: observiq-otel-collector
           group: observiq-otel-collector
-      # TODO: (dakota) evaluate this file once supervisor has support for configuring agent logging
-      - src: release_deps/logging.yaml
-        dst: /opt/observiq-otel-collector/logging.yaml
```
jsirianni, line 163: Will Windows / Linux agents log to a file? If so, what path?

.goreleaser.yml
```diff
@@ -627,14 +610,14 @@ docker_signs:
...
-      "--recursive"
-    ]
+        # Default options
+        "sign",
+        "--key=cosign.key",
+        "${artifact}",
+        "--yes",
+        # Additional options
+        "--recursive",
+      ]
```
jsirianni, line 620: PSA, we had to comment this out to fix the release (I have the ticket for turning it back on once we sort out our issues). I suspect this branch has not been rebased yet.

jsirianni (changes requested):

.goreleaser.yml
dpaasman00: Yea I haven't rebased yet

.goreleaser.yml
dpaasman00: The path is `<agent_home>/supervisor_storage/agent.log`. This is non-configurable by the supervisor - at least for now, unsure if this will change at all. The post install script messages about where to find agent logs were updated with this.

.goreleaser.yml
dpaasman00: Unsure about containers. All I know is the supervisor starts the agent as it's own subprocess and directs output from stdout to a file called `agent.log`. I'm not sure what that means for containers...

[Relevant code](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/cmd/opampsupervisor/supervisor/commander/commander.go#L74-L91)

jsirianni: I think this should be a blocker that we resolve before cutting over to v2. Logging to a file within the container will be a no go.

To be clear, it does not need to block this PR, but we do need to track this as an issue.

.goreleaser.yml
jsirianni: Cool, would be cool if it could be configured to go to the current path `/opt/observiq-otel-collector/log/collector.log` but its not a big deal.

Mostly curious.

jsirianni (approved):
Dakota and I spoke offline about supporting **container** stdout / stderr logging. The tentative plan is to push for Supervisor support in this area, probably by capturing  agent output and writing it to its own stdout / stderr.

PR #1813, opened 2024-08-24
fix: install_unix.sh breaks with status code 2 in non-interactive environments

dithmer:
### Proposed Change
The script "install_unix.sh" was not working in non-interactive shells, because there are tools used (stty and tput) for which a connected tty is mandatory. This is being fixed by a sh-compliant way of checking if the `$-` contains the `i`-flag.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/scripts/install/install_unix.sh
+++ b/scripts/install/install_unix.sh
@@ -35,33 +35,40 @@ SCRIPT_NAME="$0"
 INDENT_WIDTH='  '
 indent=""
 
+IS_INTERACTIVE="false"
+case "$-" in
+  *i*) IS_INTERACTIVE="true" ;;
+esac
+
 # out_file_path is the full path to the downloaded package (e.g. "/tmp/observiq-otel-collector_linux_amd64.deb")
 out_file_path="unknown"
 
-# Colors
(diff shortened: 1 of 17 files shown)


scripts/install/install_unix.sh
```diff
@@ -35,33 +35,40 @@ SCRIPT_NAME="$0"
 INDENT_WIDTH='  '
 indent=""
 
+IS_INTERACTIVE="false"
+case "$-" in
+  *i*) IS_INTERACTIVE="true" ;;
+esac
+
```
BinaryFissionGames, line 42: This seems to always be false in the environment I'm testing in 🤔 

For reference, I'm testing in a fresh Ubuntu Server 24.04 install. The exiting with status code 2 issue is fixed, but we don't get the coloring like we should.

BinaryFissionGames (changes requested):

Dylan-M:
I suggest closing this in favor of #1895

BinaryFissionGames:
I'm going to close this since #1895 has been merged and should solve this problem with an optional flag. Thanks for taking the time to open this PR, it's very much appreciated!

PR #1814, opened 2024-08-26
feat: Add okta receiver

colelaven:
### Proposed Change
This PR adds a new receiver to the bindplane-agent which collects logs from an Okta domain. 

##### Checklist
- [x] Changes are tested
- [x] CI has passed


receiver/oktareceiver/go.mod
```diff
@@ -0,0 +1,73 @@
+module github.com/observiq/bindplane-agent/receiver/oktareceiver
+
+go 1.23.0
```
BinaryFissionGames, line 3: Let's use go 1.21.9 or go 1.22.6 (everything is using 1.21.9 right now, but we'll be updating to 1.22 soon)

BinaryFissionGames (changes requested):

receiver/oktareceiver/factory.go
```diff
@@ -0,0 +1,56 @@
...
+	r, err := newOktaLogsReceiver(cfg, params.Logger, consumer)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create an Okta log receiver instance: %w", err)
+	}
+	return r, nil
+}
+
+func createDefaultConfig() component.Config {
+	c := &Config{
+		PollInterval: defaultPollInterval,
```
dpaasman00, line 53: I'd set the default start time here:
```suggestion
        c := &Config{
            PollInterval: defaultPollInterval,
            StartTime: time.Now().UTC().Add(-defaultPollInterval)
        }
```

Then the relevant part of `newOktaLogsReceiver()` can be simplified to:
```
startTime, err := time.Parse(OktaTimeFormat, cfg.StartTime)
if err != nil {
    return nil, err
}
```

receiver/oktareceiver/config.go
```diff
@@ -0,0 +1,102 @@
...
+		return errors.New("invalid timestamp: must be in the format YYYY-MM-DDTHH:MM:SS")
+	}
+
+	nowUTC := time.Now().UTC()
+	time180DaysAgo := nowUTC.AddDate(0, 0, -180)
+
+	// Check if the time is in the valid range the past 180 days and before now
+	if parsedTime.Before(time180DaysAgo) || parsedTime.After(nowUTC) {
+		return errors.New("invalid timestamp: must be within the past 180 days and not in the future")
+	}
```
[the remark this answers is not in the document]
colelaven: Ahh good catch on that edge case...
I agree it would be easiest to not have a custom start time, and just have it always poll starting when the receiver starts. I'll make that change

receiver/oktareceiver/logs.go
```diff
@@ -0,0 +1,237 @@
...
+	consumer  consumer.Logs
+	doneChan  chan bool
+	logger    *zap.Logger
+	nextURL   string
+	startTime time.Time
+	wg        *sync.WaitGroup
+}
+
+type httpClient interface {
+	Do(req *http.Request) (*http.Response, error)
```
dpaasman00, line 46: I'd add this function as well that can get called in `Shutdown()`
```suggestion
	Do(req *http.Request) (*http.Response, error)
    CloseIdleConnections()
```
More info: https://pkg.go.dev/net/http#Client.CloseIdleConnections

dpaasman00 (commented):
Couple comments in addition to Brandon's

receiver/oktareceiver/config.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if strings.Contains(c.Domain, "https://") || strings.Contains(c.Domain, "http://") {
+		return errInvalidDomain
+	}
+
+	if c.ApiToken == "" {
+		return errNoApiToken
+	}
+
+	if c.PollInterval != 0 && c.PollInterval < time.Second {
```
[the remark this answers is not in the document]
colelaven: I hadn't yet learned how bindplane sends the default value for optional pararmeters, and thought it would be sending over the zero value of the field. But I see now that it has a prepopulated default value that I have now set to "1m". I'll remove this case.

receiver/oktareceiver/logs.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != 200 {
+		r.logger.Error("okta logs endpoint returned non-200 statuscode: " + res.Status)
+		return logs
+	}
+
+	r.setNextLink(res)
```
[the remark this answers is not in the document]
colelaven: Great catch here as well,
I just pushed up some changes that fix this case where there are more than 1000 logs happening between polling intervals. I also added a limit parameter to the API query and set it to 1000 in order to minimize these edge cases happening. It was previously unset and the default is 100.

And actually we don't have to do 2 requests for most polls, we can just check if the # of logs returned is less than 1000 which means there are no "overflow" logs. And if there are 1000, we check if the 1000th log has a published timestamp after our initial Poll timestamp.

Another thing to note:
Okta API also mentions that with polling type requests, the logs are not _guaranteed_ to be in strictly ascending order based on published timestamp, however from my local testing hitting the okta API, I can see that they are still at least close to ascending order so I think this solution will work just fine.

There is an edge case where okta returns 1000 logs, and then 999th log has a published timestamp after our initial poll, but the 1000th log is out of order and has a timestamp before our initial poll, which in that case with the logic I just pushed, it would make another request to the log endpoint to get the next "page". 

Otherwise to make sure they are in strictly ascending order we could sort all the logs that come back if there are 1000 logs returned which I don't think would be worth the extra complexity.

receiver/oktareceiver/logs.go
```diff
@@ -0,0 +1,235 @@
...
+		if res.Body != nil {
+			defer res.Body.Close()
+		}
+
+		if res.StatusCode != http.StatusOK {
+			r.logger.Error("okta logs endpoint returned non-200 statuscode: " + res.Status)
+			break
+		}
+
+		r.setNextLink(res)
```
dpaasman00, line 146: Kinda nit, but I would move this `r.setNextLink(res)` call to the end of this for loop. The result of this func doesn't seem to have an impact on anything happening after it where it's placed now.

receiver/oktareceiver/logs.go
```diff
@@ -0,0 +1,235 @@
...
+	return logs
+}
+
+func (r *oktaLogsReceiver) Shutdown(_ context.Context) error {
+	r.logger.Debug("shutting down logs receiver")
+	r.client.CloseIdleConnections()
+	if r.cancel != nil {
+		r.cancel()
+	}
+	r.wg.Wait()
```
dpaasman00, line 215: I think the r.cancel() call should go first
```suggestion
	if r.cancel != nil {
		r.cancel()
	}
	r.client.CloseIdleConnections()
	r.wg.Wait()
```

receiver/oktareceiver/config.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+var (
+	defaultPollInterval = time.Minute
+
+	// OktaTimeFormat ISO 8601 Format
+	OktaTimeFormat = "2006-01-02T15:04:05Z"
+
+	// oktaMaxLimit maximum number of log objects returned in one call to Okta API
+	oktaMaxLimit = 1000
```
dpaasman00, line 33: I'd consider moving this over to logs.go since that's the only place it's referenced

dpaasman00 (changes requested):
Looking good, just a few minor change requests

receiver/oktareceiver/logs.go
```diff
@@ -0,0 +1,236 @@
...
+	logs := plog.NewLogs()
+
+	for _, logEvent := range logEvents {
+		resourceLogs := logs.ResourceLogs().AppendEmpty()
+		resourceLogs.ScopeLogs().AppendEmpty()
+		logRecord := resourceLogs.ScopeLogs().At(0).LogRecords().AppendEmpty()
+
+		// resource attributes
+		resourceAttributes := resourceLogs.Resource().Attributes()
+		resourceAttributes.PutStr("okta.domain", r.cfg.Domain)
```
BinaryFissionGames, line 182: Let's pull this out of the loop, each log record should be nested under the same resource and scope, since they have the same resource/scope attributes for each.

receiver/oktareceiver/logs.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		// attributes
+		logRecord.Attributes().PutStr("uuid", logEvent.Uuid)
+		logRecord.Attributes().PutStr("eventType", logEvent.EventType)
+		logRecord.Attributes().PutStr("displayMessage", logEvent.DisplayMessage)
+		logRecord.Attributes().PutStr("outcome.result", logEvent.Outcome.Result)
+		logRecord.Attributes().PutStr("actor.id", logEvent.Actor.Id)
+		logRecord.Attributes().PutStr("actor.alternateId", logEvent.Actor.AlternateId)
+		logRecord.Attributes().PutStr("actor.displayName", logEvent.Actor.DisplayName)
```
BinaryFissionGames, line 204: Could we add a nil check for actor here? Maybe in practice it never happens but the LogEvent structure allows for it.

receiver/oktareceiver/logs.go
```diff
@@ -0,0 +1,236 @@
...
+			r.logger.Error("unable to marshal logEvent", zap.Error(err))
+		} else {
+			logRecord.Body().SetStr(string(logEventBytes))
+		}
+
+		// attributes
+		logRecord.Attributes().PutStr("uuid", logEvent.Uuid)
+		logRecord.Attributes().PutStr("eventType", logEvent.EventType)
+		logRecord.Attributes().PutStr("displayMessage", logEvent.DisplayMessage)
+		logRecord.Attributes().PutStr("outcome.result", logEvent.Outcome.Result)
```
BinaryFissionGames, line 201: Same here, nil check for Outcome?

BinaryFissionGames (changes requested):

receiver/oktareceiver/logs.go
```diff
@@ -0,0 +1,236 @@
...
+			r.logger.Error("error reading response body", zap.Error(err))
+			break
+		}
+
+		var curLogs []*okta.LogEvent
+		err = json.Unmarshal(body, &curLogs)
+		if err != nil {
+			r.logger.Error("unable to unmarshal log events", zap.Error(err))
+			break
+		}
```
BinaryFissionGames, line 160: We don't necessarily have to do this here, but I was thinking about this and if we wanted to preserve the original JSON exactly, we could by unmarshalling into [json.RawMessage](https://pkg.go.dev/encoding/json#RawMessage) and later decoding that into okta.LogEvent when converting to a log record. This would let you keep the original representation of the log.

receiver/oktareceiver/logs.go
```diff
@@ -0,0 +1,236 @@
...
+	if logs.LogRecordCount() > 0 {
+		if err := r.consumer.ConsumeLogs(ctx, logs); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *oktaLogsReceiver) getLogs(ctx context.Context) []*okta.LogEvent {
+	var logs []*okta.LogEvent
```
BinaryFissionGames, line 106: Any reason for *okta.LogEvent instead of just okta.LogEvent?

receiver/oktareceiver/logs.go
colelaven: What would be the motivation to do that instead of what I wrote? So we would do the unmarshalling later in the process and an error wouldn't hold up the request cycle?

receiver/oktareceiver/logs.go
colelaven: From my understanding it would be less data to pass around between functions. Maybe my pointer knowledge is off

BinaryFissionGames: I wouldn't worry about copying between functions here, slices are already pointers so passing around a `[]*okta.LogEvent` has the same overhead as passing around a `[]okta.LogEvent`.

receiver/oktareceiver/logs.go
BinaryFissionGames: The motivation is just to preserve the original log byte-for-byte. I don't think we need to do it, to be clear. Just a thought I had.

receiver/oktareceiver/logs_test.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func jsonFileAsPlogs(filepath string) (plog.Logs, error) {
+	expectedFileBytes, err := os.ReadFile(filepath)
+	if err != nil {
+		return plog.Logs{}, nil
+	}
+	unmarshaler := &plog.JSONUnmarshaler{}
+	return unmarshaler.UnmarshalLogs(expectedFileBytes)
+}
```
BinaryFissionGames, line 220: nit: You can use this package to read/write logs for tests: 
https://pkg.go.dev/github.com/open-telemetry/opentelemetry-collector-contrib/pkg/golden

receiver/oktareceiver/logs_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+func jsonFileAsString(filePath string) string {
+	jsonBytes, err := os.ReadFile(filePath)
+	if err != nil {
+		log.Fatalf("Failed to read JSON file: %s", err)
+		return ""
+	}
+	return string(jsonBytes)
+}
```
BinaryFissionGames, line 266: nit: You can pass in the testing.T in function like these and use the `require` package, which makes the code a little terser.

BinaryFissionGames (commented):
Nice work, just a few nits in the tests, but it looks good.

receiver/oktareceiver/logs_test.go
colelaven: Ooo good tip, thats way cleaner

dpaasman00 (changes requested):
Last thing from me. Otherwise this is working well for me. Was able to connect to Okta and get logs.

BinaryFissionGames (approved):
seems good to me after Dakota's comment is addressed.

dpaasman00 (approved):

PR #1818, opened 2024-08-27
fix(chronicleforwardexporter): Escape newlines in body that is a string

kuiperda:
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/chronicleforwarderexporter/marshal.go
+++ b/exporter/chronicleforwarderexporter/marshal.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/observiq/bindplane-agent/expr"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/contexts/ottllog"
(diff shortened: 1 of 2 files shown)


exporter/chronicleforwarderexporter/marshal.go
```diff
@@ -54,6 +55,13 @@ func (m *marshaler) MarshalRawLogs(ctx context.Context, ld plog.Logs) ([]string,
 			scopeLog := resourceLog.ScopeLogs().At(j)
 			for k := 0; k < scopeLog.LogRecords().Len(); k++ {
 				logRecord := scopeLog.LogRecords().At(k)
+
+				// Escape any unescaped newlines (if body is a string)
+				logBody := logRecord.Body()
+				if logBody.Type() == pcommon.ValueTypeStr {
+					logBody.SetStr(strings.ReplaceAll(logBody.AsString(), "\n", "\\n"))
+				}
```
BinaryFissionGames, line 63: Can this issue turn up in any of these other places?
* attributes or resource attributes
* nested body fields
* a field specified with "RawLogField"

dpaasman00 (approved):
LGTM but get Brandon's too before merging

exporter/chronicleforwarderexporter/marshal.go
kuiperda: Good questions. 
- There is no issue with `\n` in attributes or resource attributes in SecOps. I think those fields are ignored, actually. 
- Nested body fields do not get parsed by SecOps and `\n` does not cut off the body:
image

- I am having trouble even sending anything with the RawLogField configured. Adding a `\n` to a field I *can* get to show up in SecOps does seem to break it, but there are fields I am trying to send without a `\n` that also don't show up at all. It's also unclear how to solve this, as escaping the newline doesn't fix the issue in the instance where we can send it. There are no errors in the agent, so this appears to be on the SecOps side. 

I don't think it's necessary to try to support this right now on our end, and it does not seem to be the usual use case (which is just sending body as a WEL string)

kuiperda: After a bit more testing it seems it can handle the `\n` if it's at the end of that field, but cannot handle any character inserted in the middle of the field.

kuiperda: After some more testing for the 3rd bullet, I can send `\n` if it's at the end of the field. Inserting any character before the end of the original value breaks it.

BinaryFissionGames (approved):

PR #1819, opened 2024-08-28
feat: Add clean install option to scripts

dpaasman00:
### Proposed Change

Adds functionality to macos and unix install scripts to ask about doing a clean install if a `supervisor.yaml` file is already located in the agent install dir. 

Tested on both debian and macos by trying to reinstall with different endpoints and secret keys (switching between dev and demo) and verifying correct behavior.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/install/install_macos.sh
+++ b/scripts/install/install_macos.sh
@@ -367,6 +367,8 @@ setup_installation() {
   set_opamp_labels
   set_opamp_secret_key
 
+  ask_clean_install
+
   success "Configuration complete!"
   decrease_indent
 }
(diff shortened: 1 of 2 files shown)


scripts/install/install_macos.sh
```diff
@@ -448,6 +450,31 @@ set_opamp_secret_key() {
...
+    *)
+      warn "Doing upgrade instead of clean install"
+      clean_install="false"
+      ;;
+    esac
+  else
+    warn "Previous supervisor config not found, doing clean install"
+    clean_install="true"
+  fi
+}
```
BinaryFissionGames, line 476: Do you think we could find a way to skip this question in non-interactive contexts? (e.g. add a flag to specify without reading from stdin)

dpaasman00: Yea, I can add a `-c` flag to indicate a clean install w/o doing the interactive

dpaasman00: Actually looks like we already use that flag, so I'll go with `-i` for install? Open to suggestions on this part

BinaryFissionGames: Probably fine as long as some flag exists 😆 There are a few "weird" choices for our short flags already.

I'm thinking we might want a way to also force a non-clean install? I'm wondering what you think that'd be useful. I think our automated tests assume you're able to non-interactively do an upgrade without specifying an endpoint.

dpaasman00: I can add it in pretty quick, so I can just do that, especially if it's something that's needed/wanted for automated testing.

BinaryFissionGames (approved):

PR #1820, opened 2024-08-28
chore: Disable confmap featuregate

dpaasman00:
### Proposed Change

Disable the `confmap.strictlyTypedInput` featuregate (needs to be done before v0.110.0 of OTel).

Ran into issues while updating OTel a few releases ago but no issues disabling now. I think there was some overlap with the environment var expansion feature gate stuff that was causing this one to fail too (issue [here](https://github.com/open-telemetry/opentelemetry-collector/issues/10552) goes into more detail). Now that the env var stuff is taken care disabling this featuregate seems fine.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/collector/featuregates.go
+++ b/collector/featuregates.go
@@ -31,9 +31,6 @@ func SetFeatureFlags() error {
 	if err := featuregate.GlobalRegistry().Set("filelog.mtimeSortType", true); err != nil {
 		return fmt.Errorf("failed to enable filelog.mtimeSortType: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


BinaryFissionGames (approved):

jsirianni (approved):
We have been running into issues with **old** versions of bindplane supporting the latest agent. This was expected, but did cause an uptick in support cases.

This change sounds safe. Just noting that we should be sure that any changes to the agent have been sorted out in recent versions of BindPlane.

PR #1824, opened 2024-08-29
chore: Custom .tar.gz and .zip unarchive functions

dpaasman00:
### Proposed Change

Adds custom functions for unpacking .tar.gz and .zip packages to the updater. Doing so to remove the `mholt/archiver/v3` dependency due to this [issue](https://github.com/mholt/archiver/pull/396). 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/opamp/observiq/observiq_downloadable_file_manager.go
+++ b/opamp/observiq/observiq_downloadable_file_manager.go
@@ -237,7 +237,7 @@ func extractTarGz(archivePath, extractPath string) error {
 			}
 
 		case tar.TypeReg:
-			if err := os.MkdirAll(filepath.Dir(outputPath), 0750); err != nil {
+			if err := os.MkdirAll(filepath.Dir(outputPath), 0700); err != nil {
 				return fmt.Errorf("create file's dir: %w", err)
 			}
 
@@ -286,7 +286,7 @@ func extractZip(archivePath, extractPath string) error {
 			continue
 		}
 
-		if err := os.MkdirAll(filepath.Dir(outputPath), 0750); err != nil {
+		if err := os.MkdirAll(filepath.Dir(outputPath), 0700); err != nil {
 			return fmt.Errorf("create file's dir: %w", err)
 		}
 
--- a/opamp/observiq/observiq_downloadable_file_manager.go
+++ b/opamp/observiq/observiq_downloadable_file_manager.go
@@ -238,19 +238,15 @@ func extractTarGz(archivePath, extractPath string) error {
 
 		case tar.TypeReg:
 			outputPathClean := filepath.Clean(outputPath)
-			outFile, err := os.Create(outputPathClean)
+			outFile, err := os.OpenFile(outputPathClean, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.FileMode(header.Mode))
 			if err != nil {
-				return fmt.Errorf("create file: %w", err)
+				return fmt.Errorf("open output file: %w", err)
 			}
 			defer outFile.Close()
 
 			_, err = io.CopyN(outFile, tarReader, maxArchiveObjectByteSize)
 			if err != nil && err != io.EOF {
 				return fmt.Errorf("write to file: %w", err)
-			}
-
-			if err := os.Chmod(outputPath, os.FileMode(header.Mode)); err != nil {
-				return fmt.Errorf("chmod on file: %w", err)
 			}
 
 		default:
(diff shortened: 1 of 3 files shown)


opamp/observiq/observiq_downloadable_file_manager.go
```diff
@@ -24,14 +27,15 @@ import (
...
+	"strings"
 
-	archiver "github.com/mholt/archiver/v3"
 	"github.com/observiq/bindplane-agent/opamp"
 	"github.com/open-telemetry/opamp-go/protobufs"
 	"go.uber.org/zap"
 )
 
 const extractFolder = "latest"
+const maxArchiveObjectByteSize = 1024
```
dpaasman00, line 38: Using this in conjunction with `io.CopyN` to resolve a gosec warning for a decompression bomb. This is the max size in bytes we will copy from a source file. Arbitrarily chose 1024, not sure what we want to do for this?

BinaryFissionGames: An object is a file, right? One of those files is going to be the collector, much bigger than this (if I'm understanding correctly). Let's set this to like 1GB.

dpaasman00: Yea no that's a good point, 1GB it is.

opamp/observiq/observiq_downloadable_file_manager.go
```diff
@@ -171,3 +175,146 @@ func (m DownloadableFileManager) CleanupArtifacts() {
...
+			outFile, err := os.Create(outputPathClean)
+			if err != nil {
+				return fmt.Errorf("create file: %w", err)
+			}
+			defer outFile.Close()
+
+			_, err = io.CopyN(outFile, tarReader, maxArchiveObjectByteSize)
+			if err != nil && err != io.EOF {
+				return fmt.Errorf("write to file: %w", err)
+			}
```
dpaasman00, line 250: As I understand it, CopyN returns an EOF error if the bytes to write (maxArchiveObjectBytesSize in this case) is larger than the source (tarReader here). I think it's safe to ignore this EOF error since we can't know for sure the exact sizes of these source files.

opamp/observiq/observiq_downloadable_file_manager.go
```diff
@@ -171,3 +175,146 @@ func (m DownloadableFileManager) CleanupArtifacts() {
...
+
+		switch header.Typeflag {
+		case tar.TypeDir:
+			if err := os.MkdirAll(outputPath, os.FileMode(header.Mode)); err != nil {
+				return fmt.Errorf("mkdir: %w", err)
+			}
+
+		case tar.TypeReg:
+			outputPathClean := filepath.Clean(outputPath)
+			outFile, err := os.Create(outputPathClean)
```
BinaryFissionGames, line 241: Why not use os.OpenFile like with zip here instead of os.Create + os.Chmod?

BinaryFissionGames (changes requested):

opamp/observiq/observiq_downloadable_file_manager.go
dpaasman00: No reason lol, I'll change that

BinaryFissionGames:
I'm getting this error when testing updating on my mac: 

`Package update failed: Failed to download and verify the supplied downloadable file: failed to extract file: extract tar.gz: create file: open tmp/latest/install/com.observiq.collector.plist: no such file or director`

BinaryFissionGames:
OK, I repulled, slightly different error message but same basic issue:
```
Failed to download and verify the supplied downloadable file: failed to extract file: extract tar.gz: open output file: open tmp/latest/install/com.observiq.collector.plist: no such file or directory
```

opamp/observiq/observiq_downloadable_file_manager.go
```diff
@@ -171,3 +175,142 @@ func (m DownloadableFileManager) CleanupArtifacts() {
...
+		}
+
+		switch header.Typeflag {
+		case tar.TypeDir:
+			if err := os.MkdirAll(outputPath, os.FileMode(header.Mode)); err != nil {
+				return fmt.Errorf("mkdir: %w", err)
+			}
+
+		case tar.TypeReg:
+			outputPathClean := filepath.Clean(outputPath)
```
BinaryFissionGames, line 240: It seems like we need to make the directory here if it doesn't exist, something like `os.MkdirAll(filepath.dir(outputPathClean), 0700)`

dpaasman00: Yeah, I think it's the `install` directory that is a part of the archive name that doesn't exist

BinaryFissionGames (approved):
Tested upgrades on mac, win, and linux, all working great.

PR #1830, opened 2024-08-30
Add marshal processor

kuiperda:
### Proposed Change

Add first iteration of Marshal processor to our Agent. (Formerly "Serialize Processor").

The intent is to allow sending data to security platforms that do their own parsing and do not want the body to be in OTLP format. Currently these use cases cannot easily use Bindplane for data reduction on those logs since they must retain the original format. 

This processor supports marshaling to JSON and KV.

XML support is going to be through an OTTL function that is being submitted to OTEL by Sam H. 

This processor makes assumptions about the format of the body. For now we are going to expect users to parse and flatten body fields correctly before using this processor. 

This is an initial draft that I expect to iterate on, but wanted to open it up for feedback. 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,110 @@
...
+		for j := 0; j < resourceLog.ScopeLogs().Len(); j++ {
+			scopeLog := resourceLog.ScopeLogs().At(j)
+			for k := 0; k < scopeLog.LogRecords().Len(); k++ {
+				logRecord := scopeLog.LogRecords().At(k)
+				logBody := logRecord.Body()
+				// If body is a string, skip that log
+				if logBody.Type() == pcommon.ValueTypeStr {
+					errors = append(errors, ErrStringBodyNotSupported)
+					continue
+				}
```
BinaryFissionGames, line 57: I'm curious on this, why are we only skipping string values? Is it OK for other non-map values to be converted? (e.g. bool, int, float)

processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,110 @@
...
+				case "XML":
+					return ld, fmt.Errorf("XML not yet supported")
+				case "KV":
+					jsonBody := logBody.AsString()
+					kvBody, err := convertJSONToKV(jsonBody)
+					if err != nil {
+						errors = append(errors, fmt.Errorf("Error converting to KV: %w", err))
+						continue
+					}
+					logBody.SetStr(kvBody)
```
BinaryFissionGames, line 71: This seems inefficient to me. Is there a reason we can't just iterate the key-values of the pcommon.Map instead of marshalling and unmarshalling JSON for every log record?

processor/marshalprocessor/config.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Validate validates the processor configuration
+func (cfg Config) Validate() error {
+	// Validate MarshalTo choice
+	switch strings.ToUpper(cfg.MarshalTo) {
+	case "JSON":
+		return nil
+	case "XML":
+		return nil
```
BinaryFissionGames, line 38: If XML is not implemented, can we make it an error to configure it?

processor/marshalprocessor/factory.go
```diff
@@ -0,0 +1,62 @@
...
+	"go.opentelemetry.io/collector/processor"
+	"go.opentelemetry.io/collector/processor/processorhelper"
+)
+
+// componentType is the value of the "type" key in configuration.
+var componentType = component.MustNewType("marshal")
+
+const (
+	stability = component.StabilityLevelAlpha
+)
```
BinaryFissionGames, line 31: I don't think we've done this for one of our processors yet, but let's start using mdatagen to generate these.

e.g. random example from an upstream processor: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/intervalprocessor/metadata.yaml

processor/marshalprocessor/processor_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	b, err := os.ReadFile(path)
+	require.NoError(t, err)
+
+	unmarshaller := plog.JSONUnmarshaler{}
+	l, err := unmarshaller.UnmarshalLogs(b)
+	require.NoError(t, err)
+
+	return l
+}
```
BinaryFissionGames, line 108: Let's use the golden package to write and read log payloads:
https://pkg.go.dev/github.com/open-telemetry/opentelemetry-collector-contrib/pkg/golden

And let's use plogtest for comparing exact log payloads:
https://pkg.go.dev/github.com/open-telemetry/opentelemetry-collector-contrib/pkg/pdatatest@v0.108.0/plogtest

processor/marshalprocessor/processor.go
BinaryFissionGames: Also, I'm not sure what happens here if you return an error + logs payload. Are all the logs dropped if you return an error for just one record, or do they get passed on anyways?

processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,110 @@
...
+		strValue := fmt.Sprintf("%v", value)
+		keyValuePairs = append(keyValuePairs, fmt.Sprintf("%s=%s", key, strValue))
+	}
+
+	// Ensure consistent order for testing
+	sort.Slice(keyValuePairs, func(i, j int) bool {
+		return keyValuePairs[i] < keyValuePairs[j]
+	})
+
+	kvBody := strings.Join(keyValuePairs, " ")
```
BinaryFissionGames, line 107: What happens if the key or value has a space in them already?

BinaryFissionGames (changes requested):

processor/marshalprocessor/processor.go
kuiperda: No, it should only allow a map. 

I am also unsure what happens, I assume we want to log errors for the records that can't be converted but allow the others through.

processor/marshalprocessor/processor.go
kuiperda: No, that's definitely a better way to do it.

processor/marshalprocessor/processor.go
dpaasman00: I don't think we should be returning these kinds of errors. We should only return an error if there's something incorrect happening. It looks like when the logsFunc gets called, the logs are dropped if there's an error.
https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/processorhelper/logs.go#L44-L56

processor/marshalprocessor/factory.go
kuiperda: Sounds good, took a pass on this

processor/marshalprocessor/processor.go
dpaasman00: If anything we can just log what happened

kuiperda: Yeah, makes sense

processor/marshalprocessor/processor.go
kuiperda: Great question... not sure if the format has a standard way to represent spaces in the key or value. I will look into it.

kuiperda: I think it will be an arbitrary escape/replacement, the only example of KV I have does not cover this case.

kuiperda: I have added support for configurable separators and quoting keys/vals that contain those separators. 
Also improved support for marshaling KV from nested fields at Dylan's suggestion.

kuiperda:
@BinaryFissionGames when you have some time, this is ready for re-review. Thanks!

processor/marshalprocessor/README.md
```diff
@@ -0,0 +1,178 @@
...
+processors:
+  marshal:
+    marshalTo: "KV"
+    kvSeparator: ","
+    kvPairSeparator: ":"
+service:
+  pipelines:
+    logs:
+      processors: [marshal]
+```
```
BinaryFissionGames, line 57: could we make these full configs with some sample exporter/receiver as well?

processor/marshalprocessor/README.md
```diff
@@ -0,0 +1,178 @@
+# Marshal Processor
+
+This processor is used to marshal parsed logs into JSON, XML, or KV format.
```
BinaryFissionGames, line 3: Should we remove XML here until it's supported?

processor/marshalprocessor/config.go
```diff
@@ -0,0 +1,61 @@
...
+		// Validate KV separators, which must be different from each other
+		if cfg.KVSeparator == cfg.KVPairSeparator && cfg.KVSeparator != 0 {
+			errs = multierr.Append(errs, errKVSeparatorsEqual)
+		}
+		if cfg.KVSeparator == 0 && cfg.KVPairSeparator == '=' {
+			errs = multierr.Append(errs, errKVSeparatorsEqual)
+		}
+		if cfg.KVPairSeparator == 0 && cfg.KVSeparator == ' ' {
+			errs = multierr.Append(errs, errKVSeparatorsEqual)
+		}
```
BinaryFissionGames, line 55: I think this could be simplified by having the default config set KVPairSeparator and KVSeparator, then you'd only need the top condition.

processor/marshalprocessor/go.mod
```diff
@@ -0,0 +1,67 @@
+module github.com/observiq/bindplane-agent/processor/marshalprocessor
+
+go 1.22.0
+
+toolchain go1.22.5
```
BinaryFissionGames, line 5: Let's set this to go 1.22.6 here to be consistent with other modules.

processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,136 @@
...
+	if mp.kvPairSeparator == 0 {
+		kvPairSeparator = " "
+	} else {
+		kvPairSeparator = string(mp.kvPairSeparator)
+	}
+	if mp.kvSeparator == 0 {
+		kvSeparator = "="
+	} else {
+		kvSeparator = string(mp.kvSeparator)
+	}
```
BinaryFissionGames, line 89: If we set the defaults for the config to be these default values, we can probably remove these if statements

BinaryFissionGames: We also might consider doing the string conversion up-front in the "newMarshalProcessor" func.

processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,136 @@
...
+
+	for k, v := range logBody.AsRaw() {
+		k = strings.ReplaceAll(k, "\"", "\\\"")
```
BinaryFissionGames, line 92: nit:  I figured I'd mention, this section might be easier to read if you use backticks instead of quotes for the strings, since they ignore escape sequences. e.g.:
```suggestion
		k = strings.ReplaceAll(k, `"`, `\"`)
```

processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,136 @@
...
+		if strings.Contains(k, kvPairSeparator) || strings.Contains(k, kvSeparator) {
+			k = "\"" + k + "\""
+		}
+
+		if valMap, ok := v.(map[string]interface{}); ok {
+			v = convertMapToString(valMap)
+		}
+
+		v = strings.ReplaceAll(fmt.Sprintf("%v", v), "\"", "\\\"")
```
BinaryFissionGames, line 101: I think it would make sense to assign to a new variable here instead of v, since we now know that `v` will be of type string after this point. That way we don't have to keep fmt.Sprintf'ing v in the next couple statements.

processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,136 @@
...
+		v := m[k]
+		switch val := v.(type) {
+		case map[string]interface{}:
+			v = convertMapToString(val)
+		default:
+			if str, ok := v.(string); ok {
+				v = strings.ReplaceAll(str, "\"", "\\\\\"")
+				if strings.ContainsAny(v.(string), ",[]=") {
+					v = fmt.Sprintf("\"%v\"", v)
+				}
```
BinaryFissionGames, line 130: This is interesting; Should the `,` and `=` here actually be the separators chosen by the user, or should these be hard-coded here?

processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+		}
+		kvPairs = append(kvPairs, fmt.Sprintf("%s=%v", k, v))
```
BinaryFissionGames, line 133: Same here, should these separators be the normal `=` and `,` for the sub kv-pairs, or should they be the user-defined ones?

processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,136 @@
...
+
+		v = strings.ReplaceAll(fmt.Sprintf("%v", v), "\"", "\\\"")
+		if strings.Contains(fmt.Sprintf("%v", v), kvPairSeparator) || strings.Contains(fmt.Sprintf("%v", v), kvSeparator) {
+			v = "\"" + fmt.Sprintf("%v", v) + "\""
+		}
+
+		kvStrings = append(kvStrings, fmt.Sprintf("%s%s%v", k, kvSeparator, v))
+	}
+
+	sort.Strings(kvStrings)
```
BinaryFissionGames, line 109: I'm not sure if this will necessarily sort by the keys in all cases. Could we do something similar to what we do in convertMapToString instead, sorting explicitly on the keys?

BinaryFissionGames (changes requested):

processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,136 @@
...
+
+	for k, v := range logBody.AsRaw() {
+		k = strings.ReplaceAll(k, "\"", "\\\"")
+		if strings.Contains(k, kvPairSeparator) || strings.Contains(k, kvSeparator) {
+			k = "\"" + k + "\""
+		}
+
+		if valMap, ok := v.(map[string]interface{}); ok {
+			v = convertMapToString(valMap)
+		}
```
BinaryFissionGames, line 99: Why do we necessarily need a separate function `convertMapToString` here? `convertToKV` already converts a map to KV format. 

Maybe instead of using logBody.AsRaw we could iterate over the pcommon values using `https://pkg.go.dev/go.opentelemetry.io/collector/pdata/pcommon#Map.Range` and recursively call `convertToKV` instead of having a separate function that seems very similar.

processor/marshalprocessor/processor.go
kuiperda: Good to know, thank you

processor/marshalprocessor/processor.go
kuiperda: Values in nested maps have to be escaped differently. But doing it recursively still makes sense.

processor/marshalprocessor/processor.go
kuiperda: Dylan gave me this example, I think they should be user-defined but allowed to be different from the normal KV separators.

processor/marshalprocessor/processor.go
kuiperda: @BinaryFissionGames I think this will properly sort by keys, when might it not? The unit tests are consistently passing with deeply nested values being sorted in the same order each run.

kuiperda: I guess the quoting of keys will change the order, and rare cases where the separator is found in the key.

kuiperda:
@BinaryFissionGames This is ready for re-review when you have some time. Thanks!

processor/marshalprocessor/README.md
```diff
@@ -0,0 +1,192 @@
...
+"body": {
+    "stringValue": {
+        bindplane-otel-attributes-baba=you
+        bindplane-otel-attributes-host=myhost
+        name=test
+        nested-n1=1
+        nested-n2=2
+        severity=155
+    }
+}
```
BinaryFissionGames, line 191: I think I'd expect this output to all be on one line

processor/marshalprocessor/processor.go
```diff
@@ -0,0 +1,143 @@
...
+		var vStr string
+		switch kv.val.Type() {
+		case pcommon.ValueTypeMap:
+			vStr = mp.convertMapToKV(kv.val.Map(), true)
+			vStr = `[` + vStr + `]`
+			if !inNestedValue && (mp.kvSeparator == mp.mapKVSeparator || mp.kvPairSeparator == mp.mapKVPairSeparator) {
+				vStr = `"` + vStr + `"`
+			}
+		default:
+			vStr = mp.escapeAndQuoteKV(kv.val.AsString(), inNestedValue)
```
BinaryFissionGames, line 113: Does supporting array values here work? Or is that a caveat here?

kuiperda: Good point. Right now they are stringified which looks a bit messy. Ultimately I don't think we need to support this differently- Talking with Dylan, he's never seen it before and doesn't have any suggestions for the format. We will also recommend flattening the body before using this processor. 

I'll add a test so this expectation is covered, though.

processor/marshalprocessor/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Configuration
+
+| Field              | Type   | Default | Description                                   |
+| ------------------ | ------ | ------- | --------------------------------------------- |
+| marshalTo          | string | ""      | The format to marshal into. Can be JSON or KV |
+| kvSeparator        | rune   | "="     | The separator between key and value           |
+| kvPairSeparator    | rune   | " "     | The separator between KV pairs                |
+| mapKVSeparator     | rune   | "="     | The separator between nested KV pairs         |
+| mapKVPairSeparator | rune   | ","     | The separator between nested KV pairs         |
```
BinaryFissionGames, line 36: Testing this out, looks like these field names should be in `snake_case`

BinaryFissionGames: This also applies to the config samples below (e.g. marshalTo should actually be `marshal_to`)

BinaryFissionGames (commented):
One last thing I see in the README, otherwise this is looking great to me.

kuiperda:
Thanks for the thorough review on this, it is much improved from where it started.

BinaryFissionGames (approved):

PR #1843, opened 2024-09-09
fix: Polish release workflow

dpaasman00:
### Proposed Change

Preparing for v2, update release workflow with necessary steps

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,8 @@ jobs:
           go-version: "1.21"
           check-latest: true
       # Needed until supervisor binary is released
+      - name: Install Tools
+        run: make install-tools
       - name: Build Supervisor Binary
         run: make release-prep
       - name: Build Windows Binaries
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -233,7 +233,7 @@ release-prep:
 	@rm -rf release_deps
 	@mkdir release_deps
 	@echo 'v$(CURR_VERSION)' > release_deps/VERSION.txt
-	./buildscripts/download-dependencies.sh release_deps
+	bash ./buildscripts/download-dependencies.sh release_deps
```
BinaryFissionGames, line 236: Can we use just sh here if necessary?
```suggestion
	sh ./buildscripts/download-dependencies.sh release_deps
```

.github/workflows/release.yml
```diff
@@ -16,6 +16,8 @@ jobs:
           go-version: "1.21"
           check-latest: true
       # Needed until supervisor binary is released
+      - name: Install Tools
+        run: make install-tools
```
BinaryFissionGames, line 20: I think this install should come before this comment

Makefile
BinaryFissionGames: We need to use bash for list support, we can revisit making this script work with posix shells in a future PR.

BinaryFissionGames (approved):

PR #1844, opened 2024-09-09
feat: Add GitHub Receiver

justinianvoss22:
### Proposed Change
GitHub's audit logs are used to keep track of events in an enterprise, organization, or user events. In order to access these logs, one must have a GitHub Enterprise Cloud account.
(https://docs.github.com/en/enterprise-cloud@latest/admin/monitoring-activity-in-your-enterprise/reviewing-audit-logs-for-your-enterprise/audit-log-events-for-your-enterprise)

Because there is an extensive amount of logs that GitHub provides, I first implemented a solution with Enterprise logs. I would like feedback on the receiver logic and what I can do better. I will continue to work on implementing logic that supports Organization and User logs. For now, this is the first iteration of the githubreceiver. 

I still have yet to hear back from the people on opentelemetry-collector-contrib on their thoughts on my implementation of audit logs. Their implementation uses web hooks and metrics, both of which I am not using yet. 

##### Checklist
- [x]  Changes are tested
- [ ] CI has passed


receiver/githubreceiver/README.md
```diff
@@ -0,0 +1,56 @@
...
+| Name          | string        |         | `true`   | The name of the user, organization or enterprise.                                                                                                                                                                                                                                                 |
+| PollInterval  | time.Duration |         | `false`  | The rate at which the receiver will poll for logs. An alternative to webhooks.                                                                                                                                                                                                                    |
+| WebhookConfig | WebhookConfig |         | `false`  | Webhooks (not configured yet) that are used when an event triggers on an enterprise, organization, or user. An alternative to polling.                                                                                                                                                            |
```
colelaven, line 34: These field names should be the same as you have them in your config below (access_token not AccessToken, etc)

receiver/githubreceiver/README.md
```diff
@@ -0,0 +1,56 @@
...
+- Access to an admin account for any enterprise, organization, or repo required for audit logs.
+
+## Configuration
+
+| Field         | Type          | Default | Requried | Description                                                                                                                                                                                                                                                                                       |
+| ------------- | ------------- | ------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| AccessToken   | string        |         | `true`   | Access token is required for audit log generation. Grants access to enterprise and organization if access token shows admin role. https://docs.github.com/en/enterprise-cloud@latest/apps/creating-github-apps/(authenticating-with-a-github-app/generating-a-user-access-token-for-a-github-app) |
```
colelaven, line 30: Use a configopaque.String for sensitive fields instead of just a normal string. 
https://github.com/open-telemetry/opentelemetry-collector/blob/e375da45118c3b82edb6adf556c8ecbc765e7aa1/config/configopaque/opaque.go#L12
Shoutout Brandon for teaching me about this

receiver/githubreceiver/logs.go
```diff
@@ -0,0 +1,315 @@
...
+	logs := r.processLogEvents(observedTime, logEvents)
+	if logs.LogRecordCount() > 0 {
+		if err := r.consumer.ConsumeLogs(ctx, logs); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *gitHubLogsReceiver) getLogs(ctx context.Context) ([]gitHubEnterpriseLog) {
```
colelaven: nit: you don't need to add parenthesis around the return type if you are returning just 1 thing.

receiver/githubreceiver/logs.go
```diff
@@ -0,0 +1,315 @@
...
+	return allLogs
+}
+
+
+
+func (r *gitHubLogsReceiver) processLogEvents(observedTime pcommon.Timestamp, logEvents []gitHubEnterpriseLog) plog.Logs {
+	logs := plog.NewLogs()
+	resourceLogs := logs.ResourceLogs().AppendEmpty()
+	resourceLogs.ScopeLogs().AppendEmpty()
+	resourceLogs.Resource().Attributes().PutStr("access_token", r.cfg.AccessToken)
```
colelaven: We probably don't want to expose the access token in the log object as it is a sensitive value.

colelaven: The log type & the name of the user/org/enterprise would be good resource attributes to add though

receiver/githubreceiver/logs.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	r.logger.Error("unable to get next link")
+	return page
+}
+
+func millisecondsToTime(ms int64) time.Time {
+    seconds := ms / 1000
+    nanoseconds := (ms % 1000) * 1000000
+    return time.Unix(seconds, nanoseconds)
+}
```
colelaven: This may be a bit simpler way to convert milliseconds to Time, I was just curious and did some googling/chatGPTing. Not saying you have to change it though just found it interesting.

```
func millisecondsToTime(ms int64) time.Time {
    return time.Unix(0, ms*int64(time.Millisecond))
}
```

colelaven (changes requested):

colelaven:
You will also need to make some small changes in the docs/receivers.md file and factories/receivers.go similar to this PR: https://github.com/observIQ/bindplane-agent/pull/1814/files

But I see this is a draft pr and you are waiting on a few other parts & open source so no biggie just wanted to make sure you were aware

PR #1855, opened 2024-09-12
fix: Call shutdown on storage extension

dpaasman00:
### Proposed Change

Shutdown should be calling shutdown on the storage extension. When it doesn't, the next start up fails because it cannot connect to the file extension. 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):
LGTM

PR #1877, opened 2024-09-24
feat: add ottl condition field to sampling processor

colelaven:
### Proposed Change
Adds a 'condition' field to the Sampling Processor, where only telemetry that passes the condition will be sampled from.

## Functionality remains intact when no condition is inputted
image
image

## Sampling only runs on telemetry that passes the given condition
image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


processor/samplingprocessor/processor.go
```diff
@@ -18,79 +18,201 @@ import (
...
-func (sp *samplingProcessor) processTraces(_ context.Context, td ptrace.Traces) (ptrace.Traces, error) {
+func (sp *metricsSamplingProcessor) sampleFunc() bool {
+	//#nosec G404 -- randomly generated number is not used for security purposes. It's ok if it's weak
+	return rand.Float64() <= sp.dropCutOffRatio
+}
+
+func (sp *tracesSamplingProcessor) sampleFunc() bool {
+	//#nosec G404 -- randomly generated number is not used for security purposes. It's ok if it's weak
+	return rand.Float64() <= sp.dropCutOffRatio
+}
```
dpaasman00, line 86: To reduce a bit of code dup, why don't we refactor to one sampleFunc:
```
func sampleFunc(dropCutOffRatio float64) bool {
    return rand.Float64 <= dropCutOffRatio
}
```

And then replace the existing calls with `sampleFunc(sp.dropCutOffRatio)`

processor/samplingprocessor/processor.go
```diff
@@ -18,79 +18,201 @@ import (
...
+
+func (sp *tracesSamplingProcessor) processTraces(ctx context.Context, td ptrace.Traces) (ptrace.Traces, error) {
 	switch {
-	case sp.dropCutOffRatio == 1.0: // Drop everything
-		return ptrace.NewTraces(), nil
 	case sp.dropCutOffRatio == 0.0: // Drop nothing
 		return td, nil
-	default: // Drop based on ratio
+	default: // Drop based on ratio and condition
```
dpaasman00, line 92: For these process funcs, can we refactor the switch statement to an if conditional for simplicity
```
if sp.dropCutOffRatio == 0.0 {
    return td, nil
}
....
```

dpaasman00 (changes requested):
Looks good overall, just a few style related things

dpaasman00 (approved):
Working well for me

processor/samplingprocessor/processor.go
```diff
@@ -18,79 +18,192 @@ import (
...
+					metricMatch, err := sp.condition.Match(ctx, metricCtx)
+					anyMatch = err == nil && metricMatch
+				}
+
+				return anyMatch && sampleFunc(sp.dropCutOffRatio)
+			})
 		}
-		return md, nil
+	}
+	return md, nil
```
BinaryFissionGames, line 170: Because we filter whole metrics, not datapoints, we should use the metric context and not the datapoint context here. That probably requires extending the `github.com/observiq/bindplane-agent/expr` package, but it'll make more sense to do it that way.

BinaryFissionGames (changes requested):

processor/samplingprocessor/processor.go
colelaven: Unfortunate timing on the merge 😳 great catch, I just opened a new PR to address your feedback: #1878

PR #1878, opened 2024-09-24
fix: add ottl metric condition to expr package for use in sampling processor

colelaven:
### Proposed Change
Adds OTTL Metric Condition, Statement, and Expression to the expr package. Changes the sampling processor to use a metric based condition instead of datapoint based condition.

Followup to: #1877 

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/processor/samplingprocessor/config.go
+++ b/processor/samplingprocessor/config.go
@@ -36,9 +36,5 @@ func (cfg Config) Validate() error {
 		return errInvalidDropRatio
 	}
 
-	if cfg.Condition == "" {
-		cfg.Condition = "true"
-	}
-
 	return nil
 }
--- a/processor/samplingprocessor/config.go
+++ b/processor/samplingprocessor/config.go
@@ -36,5 +36,9 @@ func (cfg Config) Validate() error {
 		return errInvalidDropRatio
 	}
 
+	if cfg.Condition == "" {
+		cfg.Condition = "true"
+	}
+
 	return nil
 }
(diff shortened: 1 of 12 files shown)


processor/samplingprocessor/go.mod
```diff
@@ -50,3 +50,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/observiq/bindplane-agent/expr => ../../expr
```
BinaryFissionGames, line 54: We'll also want to change the version of `github.com/observiq/bindplane-agent/expr` to v1.61.0 up top; 

You could do this by just running `make update-modules NEW_VERSION=v1.61.0`.

BinaryFissionGames (changes requested):
Looks good, just one comment on the go.mod

BinaryFissionGames (approved):

processor/samplingprocessor/config.go
```diff
@@ -36,5 +36,9 @@ func (cfg Config) Validate() error {
 		return errInvalidDropRatio
 	}
 
+	if cfg.Condition == "" {
+		cfg.Condition = "true"
+	}
```
BinaryFissionGames, line 41: Do we ever hit this case? We already set the default to this in createDefaultConfig. IMO we shouldn't modify the config in a "Validate" function, I wouldn't expect validation to mutate the config.

BinaryFissionGames (changes requested):

processor/samplingprocessor/config.go
colelaven: Good point, I wasn't sure how the condition field would be set if it isn't passed in from the config, but I see now with some testing that it gets set to the default. Change has been pushed 👍

BinaryFissionGames (approved):

PR #1879, opened 2024-09-25
fix: Sampling Processor not removing higher level objs

dpaasman00:
### Proposed Change

While testing the Sampling Processor I noticed that in BindPlane the 'Filter Debug Logs' processor recommendation was being recommended despite no debug logs being present.
Screenshot 2024-09-24 at 11 01 49 AM

After debugging, I discovered this was occurring because this processor only deleted log records, leaving behind the parent scope and resource level objects. This was causing these objects to be initialized with default values when they were technically empty.

This change cleans up these objects if they are effectively empty. So resource spans/metrics/logs are removed if there are no scope entries, and scope spans/metrics/logs are removed if there are no spans/metrics/logRecords present. This successfully removes the filter debug recommendation.
image

To test, add the following replace statements to the transform agent `go.mod`, and then configure local bindplane to use this build of the transform agent (or just use a remote bta with this build).

``` 
replace github.com/observiq/bindplane-agent/processor/samplingprocessor => github.com/observiq/bindplane-agent/processor/samplingprocessor v1.60.1-0.20240925175304-a8f9052d43bf
replace github.com/observiq/bindplane-agent/expr => github.com/observiq/bindplane-agent/expr v1.60.1-0.20240925175304-a8f9052d43bf
```

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


processor/samplingprocessor/processor.go
```diff
@@ -111,7 +111,13 @@ func (sp *tracesSamplingProcessor) processTraces(ctx context.Context, td ptrace.
 				return err == nil && match && sampleFunc(sp.dropCutOffRatio)
 			})
 		}
+		td.ResourceSpans().At(i).ScopeSpans().RemoveIf(func(ss ptrace.ScopeSpans) bool {
+			return ss.Spans().Len() == 0
+		})
```
BinaryFissionGames, line 116: Could we add some go test that verifies this new behavior?

BinaryFissionGames (changes requested):

processor/samplingprocessor/processor.go
dpaasman00: Added some tests, had to use `require.NotEqual(t, 0, logRecords)` instead of NotEmpty. For whatever reason it couldn't detect an empty pmetric slice. I verified the tests expose the old behavior by commenting out the fixes

BinaryFissionGames (approved):

PR #1880, opened 2024-09-25
feat: add ottl condition field to dedupe logs processor

colelaven:
### Proposed Change

Adds OTTL Condition field to Deduplicate Logs Processor

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


processor/logdeduplicationprocessor/factory_test.go
```diff
@@ -46,12 +47,28 @@ func TestCreateLogsProcessor(t *testing.T) {
...
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			f := NewFactory()
-			p, err := f.CreateLogsProcessor(context.Background(), processor.Settings{}, tc.cfg, nil)
+			set := processor.Settings{
+				TelemetrySettings: component.TelemetrySettings{
+					Logger: zap.NewNop(),
+				},
+			}
```
BinaryFissionGames, line 70: nit: could use [processortest.NewNopSettings](https://pkg.go.dev/go.opentelemetry.io/collector/processor@v0.110.0/processortest#NewNopSettings)

BinaryFissionGames:
I just realized this is a processor we moved upstream recently. Could you open an issue for implementing this upstream? I think ideally we'd be upstreaming any changes we make to this processor. We'll need to switch over to the upstream version at some point.

colelaven:
Per Brandon's comment, I will make this change upstream instead.

PR #1887, opened 2024-09-27
feat: Allow metric conversion functions on datapoint and metric context

BinaryFissionGames:
### Proposed Change
* Allow datapoint and metric context to convert between sum and gauge

You can see the diff here: https://github.com/open-telemetry/opentelemetry-collector-contrib/compare/v0.110.0...observIQ:opentelemetry-collector-contrib:fix/v0.110.0-transform-processor-interop?expand=1

Tested with this sample config:
```yaml
receivers:
  hostmetrics:
    collection_interval: 60s
    scrapers:
      load:
        metrics: null

processors:
  transform:
    metric_statements:
      - context: metric
        statements:
          - convert_gauge_to_sum("cumulative", false)
          - convert_sum_to_gauge()
      - context: datapoint
        statements:
          - convert_gauge_to_sum("cumulative", false)
          - convert_sum_to_gauge()

exporters:
  debug:
    verbosity: detailed

service:
  pipelines:
    logs:
      receivers: [hostmetrics]
      processors: [transform]
      exporters: [debug]
```

Here, both contexts have the convert functions, so you can use them in either. We need this for BPOP to be able to upgrade the agent properly.

We plan to remove this in v1.63.0, at which point we should only support using the metric context for these functions.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/collector/featuregates.go
+++ b/collector/featuregates.go
@@ -31,10 +31,6 @@ func SetFeatureFlags() error {
 	if err := featuregate.GlobalRegistry().Set("filelog.mtimeSortType", true); err != nil {
 		return fmt.Errorf("failed to enable filelog.mtimeSortType: %w", err)
 	}
-	// Temporary while we get some interoperability into BPOP
-	if err := featuregate.GlobalRegistry().Set("processor.transform.ConvertBetweenSumAndGaugeMetricContext", false); err != nil {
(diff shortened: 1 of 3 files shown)


dpaasman00 (approved):
Replace working for me

PR #1895, opened 2024-10-02
feat(install): Add quiet flag to allow use in non-interactive shells

Dylan-M:
### Proposed Change
Add a new quiet flag that allows running this script in a non-interactive shell, including performing a prerequisite check that prevents running when interactive features are required.

##### Checklist
- [X] Changes are tested
- [x] CI has passed

### DIFF
--- a/scripts/install/install_unix.sh
+++ b/scripts/install/install_unix.sh
@@ -801,8 +801,6 @@ uninstall_package()
 
 uninstall()
 {
-  bindplane_banner
-
   set_package_type
   banner "Uninstalling BindPlane Agent"
   increase_indent
--- a/scripts/install/install_unix.sh
+++ b/scripts/install/install_unix.sh
@@ -34,34 +34,37 @@ PREREQS="curl printf $SVC_PRE sed uname cut"
 SCRIPT_NAME="$0"
 INDENT_WIDTH='  '
 indent=""
+non_interactive=false
 
 # out_file_path is the full path to the downloaded package (e.g. "/tmp/observiq-otel-collector_linux_amd64.deb")
 out_file_path="unknown"
 
 # Colors
-num_colors=$(tput colors 2>/dev/null)
-if test -n "$num_colors" && test "$num_colors" -ge 8; then
-  bold="$(tput bold)"
-  underline="$(tput smul)"
-  # standout can be bold or reversed colors dependent on terminal
-  standout="$(tput smso)"
-  reset="$(tput sgr0)"
-  bg_black="$(tput setab 0)"
-  bg_blue="$(tput setab 4)"
-  bg_cyan="$(tput setab 6)"
-  bg_green="$(tput setab 2)"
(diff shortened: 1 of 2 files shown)


scripts/install/install_unix.sh
```diff
@@ -34,34 +34,37 @@ PREREQS="curl printf $SVC_PRE sed uname cut"
...
-  bg_yellow="$(tput setab 3)"
-  fg_black="$(tput setaf 0)"
-  fg_blue="$(tput setaf 4)"
-  fg_cyan="$(tput setaf 6)"
-  fg_green="$(tput setaf 2)"
-  fg_magenta="$(tput setaf 5)"
-  fg_red="$(tput setaf 1)"
-  fg_white="$(tput setaf 7)"
-  fg_yellow="$(tput setaf 3)"
+if [ "$non_interactive" = "true" ]; then
```
dithmer, line 43: I really looked at this multiple times now. But as far as I get it, this should be inversed right?

In case of non-interactive mode, there should be no colors, but in interactive mode. So this conditional needs to be 
`"$non_interactive" = "false"`?

Same for the MacOS script.

dithmer (changes requested):

scripts/install/install_unix.sh
```diff
@@ -823,6 +847,8 @@ main()
   if [ $# -ge 1 ]; then
     while [ -n "$1" ]; do
       case "$1" in
+        -q|--quiet)
```
dithmer, line 850: In my opinion, a quiet flag should not just remove the color of the ouput, but shrink the ouput at all completely or at least to a minimum, shouldn't it?

Dylan-M: A future revision of the scripts can account for full quiet where all non-error messages are suppressed. That will require a little refactoring that I'm not ready for due to the mess it will make of another open PR. I'll adjust description to account for that.

Dylan-M: Scratch that, I forgot we were overriding printf with our own function of it. I changed that function instead.

jsirianni (changes requested):
edit: --quiet flag, not proxy. Clarified w/ Dylan offline.

I get this error when passing only the proxy flag.

```
  ERROR (scripts/install/install_unix.sh:546): The proxy password must be set via the command line argument -P, if called non-interactively!
    546 The proxy password must be set via the command line argument -P, if called non-interactively!
```

Dylan-M:
> I get this error when passing only the proxy flag.
> 
> ```
>   ERROR (scripts/install/install_unix.sh:546): The proxy password must be set via the command line argument -P, if called non-interactively!
>     546 The proxy password must be set via the command line argument -P, if called non-interactively!
> ```

Exactly, if you're non-interactive (quiet mode), we can't prompt for the password. That is why it errors. Unless you're saying it drops the error when you're NOT passing -q?

PR #1900, opened 2024-10-07
fix: Update manifests to work at runtime

dpaasman00:
### Proposed Change

Update manifests:
- remove tencent exporter from bindplane manifest
- list out all providers in minimal manifest

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/manifests/minimal/manifest.yaml
+++ b/manifests/minimal/manifest.yaml
@@ -1,6 +1,6 @@
 dist:
   module: github.com/observiq/bindplane-agent-ocb
-  name: observiq-distro
+  name: com.observiq.collector
   description: ObservIQ's minimal distro for OpenTelemetry Collector
   version: "v0.0.1"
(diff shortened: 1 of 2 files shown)


manifests/minimal/manifest.yaml
```diff
@@ -1,6 +1,6 @@
 dist:
   module: github.com/observiq/bindplane-agent-ocb
-  name: observiq-distro
+  name: com.observiq.collector
```
BinaryFissionGames, line 3: This seems pretty weird for the name of the collector, also not sure how it will work with windows where it adds a `.exe` to the end of the executable name. Is it possible to just set `service.name` in the supervisor config instead?

BinaryFissionGames (changes requested):

BinaryFissionGames (approved):

PR #1907, opened 2024-10-09
feat: Update to v2

dpaasman00:
### Proposed Change

Reworks updater to update an installation from v1 to v2.
- Reads `endpoint`, `secret_key`, and `agent_id` values from existing manager.yaml and uses them to create basic supervisor.yaml
- Uses healthcheck extension on agent to determine if startup is successful once service is restarted
- Cleans up various files from v1 install that are no longer needed in a v2 install

Note:
- This does not apply the previously used config, users will need to reapply a config from BP
- Only the values listed above are carried over from the `manager.yaml`

### Testing
Test following this guide: https://docs.google.com/document/d/1ZaSXnJIu-rjctiihBdBLvHymuaB7GFIW8I56biN5v8Q/edit?tab=t.0

1. Run `VERSION=v2.0.0 make release-test` in this repo
2. Run the following script as `./build-agent-upgrade.sh version=v2.0.0` from the root of this repo:
```bash
#!/bin/sh
VERSION="$1"

ref_name=$(git rev-parse HEAD | cut -b 1-8)

mkdir -p ./artifacts
mkdir -p ./tmp
cp ./scripts/install/*.sh ./artifacts
cp ./dist/*tar.gz ./artifacts
cp ./dist/*.rpm ./artifacts
cp ./dist/*.deb ./artifacts
cp ./dist/*SHA256SUMS ./artifacts
tar -czvf "./tmp/observiq-otel-collector-${VERSION}-SNAPSHOT-${ref_name}-artifacts.tar.gz" -C ./artifacts .

rm -rf ./artifacts
```
3. Make sure you have a v1.64.0 or older Agent running and connected to local BP.
4. Restart local BP instance after changing/adding top level config key `offline: true`.
5. Apply the agent-upgrade to BP using `bindplane upload agent-upgrade ~/bindplane-agent/tmp/observiq-otel-collector-...`.
6. Select the agent in the BP agent table and select 'Upgrade Outdated Agent'.
7. Verify Agent shuts down and comes back as a v2 agent with the same ID and is running.

##### Checklist
- [ ] Changes are tested
- [x] CI has passed


updater/internal/updater/updater.go
```diff
@@ -145,3 +151,33 @@ func (u *Updater) removeTmpDir() {
 		u.logger.Error("failed to remove temporary directory", zap.Error(err))
 	}
 }
+
+// cleanup removes unnecessary files now that agent has updated from v1 to v2.
+// Update is already successfully completed at this point, so we don't want to return errors.
+func (u *Updater) cleanup() {
+	files := []string{filepath.Join("log", "collector.log"), "config.yaml", "manager.yaml", "updater", "config.bak.yaml", "logging.yaml", "package_statuses.json"}
```
dpaasman00, line 158: These are the files we delete once a v1 to v2 update is successful, let me know if you'd rather keep any of these around instead of deleting. Conversely, I currently keep around `log/updater.log` so if you want that deleted instead let me know

updater/internal/install/install.go
```diff
@@ -223,3 +243,216 @@ func skipConfigFiles(path string) bool {
...
+		return fmt.Errorf("read in agent id")
+	}
+	agentULID, err := ulid.Parse(agentID)
+	if err != nil {
+		return fmt.Errorf("parse agent id into a ULID: %w", err)
+	}
+	ulidBytes, err := agentULID.MarshalBinary()
+	if err != nil {
+		return fmt.Errorf("marshal ulid to bytes: %w", err)
+	}
```
BinaryFissionGames, line 416: With #1907 merged, we'll want to check if it's a UUID as well.

Side note, it does suck a bit that this will cause the agent to report itself with a new ID. I don't think there's a way around it though.

BinaryFissionGames (changes requested):

dpaasman00:
Moving away from updating from v1 to v2. Changes to how the supervisor runs the collector prevents accurately verifying collector has started up.

PR #1908, opened 2024-10-09
feat: Use configurable supervisor logging

dpaasman00:
### Proposed Change

Configure default supervisor config with some logging options
- Info level logging
- Log to a file at `supervisor_storage/supervisor.log`

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (changes requested):
@dpaasman00 is the supervisor log rotated by default? I suspect it is but want to be sure.

jsirianni (approved):

PR #1909, opened 2024-10-10
chore: Don't require OpAMP endpoint when installing & starting

dpaasman00:
### Proposed Change

Want to be able to install and start supervisor without needing to provide an opamp endpoint or secret key. Neither are no longer required but since the supervisor needs an endpoint to actually start, use a default of `ws://localhost:3001/v1/opamp` when one isn't provided

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/install/install_macos.sh
+++ b/scripts/install/install_macos.sh
@@ -28,6 +28,9 @@ SCRIPT_NAME="$0"
 INDENT_WIDTH='  '
 indent=""
 
+# Default Supervisor Config Hash
+DEFAULT_SUPERVISOR_CFG_HASH="ac4e6001f1b19d371bba6a2797ba0a55d7ca73151ba6908040598ca275c0efca"
+
 # Colors
 num_colors=$(tput colors 2>/dev/null)
 if test -n "$num_colors" && test "$num_colors" -ge 8; then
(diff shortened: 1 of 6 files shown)


scripts/install/install_macos.sh
```diff
@@ -594,6 +591,13 @@ create_supervisor_config() {
 
   info "Creating supervisor config..."
 
+  if [ -z "$OPAMP_ENDPOINT" ]; then
+    OPAMP_ENDPOINT="ws://localhost:3000/v1/opamp"
+    increase_indent
+    info "No OpAMP endpoint specified, starting agent using 'ws://localhost:3001/v1/opamp' as endpoint."
```
jsirianni, line 597: Line 595 has port `3000`, not `3001`. Is this a problem? Looks like it deployed correctly for me, with 3001.

scripts/install/install_unix.sh
```diff
@@ -718,6 +713,15 @@ create_supervisor_config() {
     return
   fi
 
+  info "Creating supervisor config..."
+
+  if [ -z "$OPAMP_ENDPOINT" ]; then
+    OPAMP_ENDPOINT="ws://localhost:3000/v1/opamp"
```
jsirianni, line 719: Same as previous comment

jsirianni (changes requested):
Windows seems okay, some macos and unix questions before I move on.

scripts/install/install_macos.sh
dpaasman00: Yea good catch, that should be `3001`, I was just trying to use the default for a local BP server as the default OpAmp endpoint.

jsirianni (changes requested):
I tested RHEL and Debian, both are missing default supervisor configs when installing the package directly.

What do you think about including a default config in the packages, and using the install script to overwrite it?

I think we should assume users might install the agent without the script, especially if they do not care about OpAMP.

I know this would not work right away, as the script does not deploy the supervisor config if it already exists. We would need the ability to update it.

Happy to discuss further.

scripts/install/install_macos.sh
```diff
@@ -465,21 +468,28 @@ ask_clean_install() {
...
-    read -r clean_install_response
-    clean_install_response=$(echo "$clean_install_response" | tr '[:upper:]' '[:lower:]')
-    case $clean_install_response in
-    y | yes)
-      increase_indent
-      success "Doing clean install!"
-      decrease_indent
+    # Check for default config file hash
+    cfg_file_hash=$(sha256sum "$SUPERVISOR_YML_PATH" | awk '{print $1}')
+    if [ "$cfg_file_hash" == "$DEFAULT_SUPERVISOR_CFG_HASH"]; then
```
jsirianni, line 473: Shellcheck fails here. Will also need to be fixed on the linux script.

```
In scripts/install/install_macos.sh line 473:
    if [ "$cfg_file_hash" == "$DEFAULT_SUPERVISOR_CFG_HASH"]; then
    ^-- SC1009 (info): The mentioned syntax error was in this if expression.
       ^-- SC1073 (error): Couldn't parse this test expression. Fix to allow more checks.
                                                            ^-- SC1020 (error): You need a space before the ].
                                                            ^-- SC1072 (error): Missing space before ]. Fix any mentioned problems and try again.
```

jsirianni (changes requested):

scripts/install/install_macos.sh
dpaasman00: Also changed those conditionals to use `=` instead of `==` since shellcheck was giving me warnings for that: https://www.shellcheck.net/wiki/SC3014

jsirianni (approved):

PR #1915, opened 2024-10-15
fix: Fix panic when exporting tracing to GCP

BinaryFissionGames:
### Proposed Change
There's a panic caused by us updating to v1.24.2 of the `github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace` package. A method was added to the interface, which this module uses, but the other module, which implements this interface, _embeds_ the interface. This means that the implementor always trivially implements the interface, even when it doesn't actually implement a method. When this unimplemented method is called, a nil panic happens because the embedded interface is always set to nil.

TL;DR v1.24.2 is incompatible with v0.111.0 of the GCP exporter, we need to use v1.24.1.

Here's a sample config I used to reproduce/test (this would panic immediately on startup):
```yaml
receivers:
  telemetrygeneratorreceiver:
    payloads_per_second: 1
    generators:
      - type: otlp
        additional_config:
          telemetry_type: "traces"
          otlp_json: '{"resourceSpans":[{"resource":{"attributes":[{"key":"service.name","value":{"stringValue":"shopping-cart"}},{"key":"host.name","value":{"stringValue":"web-server-1"}},{"key":"duplicateKey","value":{"stringValue":"duplicateValueFromResource"}}]},"scopeSpans":[{"scope":{"name":"exampleTracer","version":"1.0"},"spans":[{"traceId":"0102030405060708090a0b0c0d0e0f10","spanId":"0102030405060708","name":"CheckoutProcess","kind":"SPAN_KIND_INTERNAL","startTimeUnixNano":"1646735053200000000","endTimeUnixNano":"1646735055200000000","attributes":[{"key":"http.method","value":{"stringValue":"GET"}},{"key":"duplicateKey","value":{"stringValue":"duplicateValueFromSpan"}},{"key":"user.id","value":{"intValue":12345}}],"status":{"code":"STATUS_CODE_OK"}}]}]}]}'

exporters:
  googlecloud/otel-agent-dev:
    credentials_file: "<redacted>"
    log:
      compression: gzip
      resource_filters:
        - regex: .*
    metric:
      compression: gzip
    project: "<redacted>"

service:
  pipelines:
    traces:
      receivers:
        - telemetrygeneratorreceiver
      exporters:
        - googlecloud/otel-agent-dev
```

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -495,7 +495,7 @@ require (
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp v1.24.1 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector v0.48.1 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector/googlemanagedprometheus v0.48.1 // indirect
-	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace v1.24.2 // indirect
+	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace v1.24.1 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/internal/resourcemapping v0.48.2 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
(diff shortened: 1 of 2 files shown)


dpaasman00 (approved):
Works well

PR #1927, opened 2024-10-24
feat: Add AES confmap provider

shazlehu:
### Proposed Change

Adds the AES confmap provider for obfuscating sensitive config values.

You can test that the provider works using this [PR stack](https://bindplane-ee-3621-bbolt.stage.bindplane.com) and setting a key under the advanced config options. When you inspect the config on the disc, any sensitive parameters should be replaced with `${aes:<base 64 encoded, encrypted string>}`

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dpaasman00 (approved):
Works well for me

PR #1939, opened 2024-10-30
feat(chronicleexporter): Support dynamic namespace and ingestion labels 

justinianvoss22:
### Proposed Change
- Update chronicle exporter to be able to take namespace and ingestion labels from attributes if present.
- Make logType not required

### Testing
- Unit testing
- Testing agent on local BP

##### Checklist
- [ x ] Changes are tested
- [ x ] CI has passed

### DIFF
--- a/exporter/chronicleexporter/README.md
+++ b/exporter/chronicleexporter/README.md
@@ -49,8 +49,12 @@ currently supported log types are:
 - sql_server
 
 If the `attributes["chronicle_log_type"]` field is present in the log, we will use its value in the payload instead of the automatic detection or the `log_type` in the config.
-If the `attributes["chronicle_namespace"]` field is present in the log, we will use its value in the payload.
-If there are nested fields in `attributes["chronicle_ingestion_label"]`, we will use the values in the payload.
+
+### Namespace and Ingestion Labels
+
+If the `attributes["chronicle_namespace"]` field is present in the log, we will use its value in the payload instead of the `namespace` in the config.
+
+If there are nested fields in `attributes["chronicle_ingestion_label"]`, we will use the values in the payload instead of the `ingestion_labels` in the config.
 
 ## Credentials
 
--- a/exporter/chronicleexporter/README.md
+++ b/exporter/chronicleexporter/README.md
@@ -22,19 +22,19 @@ This exporter facilitates the sending of logs to Chronicle, which is a security 
 
 The exporter can be configured using the following fields:
 
-| Field                  | Type              | Default                                | Required | Description                                                                                 |
-| ---------------------- | ----------------- | -------------------------------------- | -------- | ------------------------------------------------------------------------------------------- |
-| `endpoint`             | string            | `malachiteingestion-pa.googleapis.com` | `false`  | The Endpoint for sending to chronicle.                                                      |
-| `creds_file_path`      | string            |                                        | `true`   | The file path to the Google credentials JSON file.                                          |
-| `creds`                | string            |                                        | `true`   | The Google credentials JSON.                                                                |
-| `log_type`             | string            |                                        | `true`   | The type of log that will be sent.                                                          |
-| `raw_log_field`        | string            |                                        | `false`  | The field name for raw logs.                                                                |
-| `customer_id`          | string            |                                        | `false`  | The customer ID used for sending logs.                                                      |
-| `override_log_type`    | bool              | `true`                                 | `false`  | Whether or not to override the `log_type` in the config with `attributes["log_type"]`       |
-| `namespace`            | string            |                                        | `false`  | User-configured environment namespace to identify the data domain the logs originated from. |
-| `compression`          | string            | `none`                                 | `false`  | The compression type to use when sending logs. valid values are `none` and `gzip`           |
-| `ingestion_labels`     | map[string]string |                                        | `false`  | Key-value pairs of labels to be applied to the logs when sent to chronicle.                 |
-| `collect_agent_metrics` | bool              | `true`                                 | `false`  | Enables collecting metrics about the agent's process and log ingestion metrics          |
+| Field                   | Type              | Default                                | Required | Description                                                                                 |
+| ----------------------- | ----------------- | -------------------------------------- | -------- | ------------------------------------------------------------------------------------------- |
+| `endpoint`              | string            | `malachiteingestion-pa.googleapis.com` | `false`  | The Endpoint for sending to chronicle.                                                      |
+| `creds_file_path`       | string            |                                        | `true`   | The file path to the Google credentials JSON file.                                          |
+| `creds`                 | string            |                                        | `true`   | The Google credentials JSON.                                                                |
+| `log_type`              | string            |                                        | `false`  | The type of log that will be sent.                                                          |
+| `raw_log_field`         | string            |                                        | `false`  | The field name for raw logs.                                                                |
+| `customer_id`           | string            |                                        | `false`  | The customer ID used for sending logs.                                                      |
+| `override_log_type`     | bool              | `true`                                 | `false`  | Whether or not to override the `log_type` in the config with `attributes["log_type"]`       |
+| `namespace`             | string            |                                        | `false`  | User-configured environment namespace to identify the data domain the logs originated from. |
+| `compression`           | string            | `none`                                 | `false`  | The compression type to use when sending logs. valid values are `none` and `gzip`           |
+| `ingestion_labels`      | map[string]string |                                        | `false`  | Key-value pairs of labels to be applied to the logs when sent to chronicle.                 |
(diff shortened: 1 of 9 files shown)


jsirianni:
I updated the PR name, it will help w/ release notes. Feel free to tweak it if its incorrect.

kuiperda:
Took a quick high level pass and it makes sense to me, will let Dakota review though.

exporter/chronicleexporter/marshal.go
```diff
@@ -50,7 +51,6 @@ type logMarshaler interface {
 	MarshalRawLogs(ctx context.Context, ld plog.Logs) ([]*api.BatchCreateLogsRequest, error)
 	MarshalRawLogsForHTTP(ctx context.Context, ld plog.Logs) (map[string]*api.ImportLogsRequest, error)
 }
-
 type protoMarshaler struct {
```
dpaasman00, line 54: Since we're using dynamic labels, the `labels` object on this struct is no longer used and should be removed.

exporter/chronicleexporter/marshal.go
```diff
@@ -73,64 +72,89 @@ func newProtoMarshaler(cfg Config, teleSettings component.TelemetrySettings, lab
...
+					namespaceMap[logType] = namespace
+				}
+				if len(ingestionLabels) != 0 {
+					if ingestionLabelsMap[logType] == nil {
+						ingestionLabelsMap[logType] = make([]*api.Label, 0)
+					}
+					for _, v := range ingestionLabels {
+						ingestionLabelsMap[logType] = append(ingestionLabelsMap[logType], v)
+					}
+				}
```
dpaasman00, line 122: I think some more thought needs to go into how these labels get added. Currently we append all labels present on a log record to the overarching map, regardless of if they're already present. This causes repeat entries if there are multiple log records with the same labels. You can add this test case to the test `TestProtoMarshaler_MarshalRawLogs` in `marshal_test.go` test to see this behavior:
```go
{
			name: "Multiple log records with attributes",
			cfg: Config{
				CustomerID:      uuid.New().String(),
				LogType:         "WINEVTLOG",
				Namespace:       "test",
				RawLogField:     "body",
				OverrideLogType: false,
			},
			labels: []*api.Label{},
			logRecords: func() plog.Logs {
				logs := plog.NewLogs()
				record1 := logs.ResourceLogs().AppendEmpty().ScopeLogs().AppendEmpty().LogRecords().AppendEmpty()
				record1.Body().SetStr("First log message")
				record1.Attributes().FromRaw(map[string]any{`chronicle_ingestion_label["key1"]`: "value1", `chronicle_ingestion_label["key2"]`: "value2"})
				record2 := logs.ResourceLogs().AppendEmpty().ScopeLogs().AppendEmpty().LogRecords().AppendEmpty()
				record2.Body().SetStr("Second log message")
				record2.Attributes().FromRaw(map[string]any{`chronicle_ingestion_label["key1"]`: "value1", `chronicle_ingestion_label["key2"]`: "value2"})
				return logs
			},
			expectations: func(t *testing.T, requests []*api.BatchCreateLogsRequest) {
				require.Len(t, requests, 1, "Expected a single batch request")
				batch := requests[0].Batch
				// verify batch source labels
				require.Len(t, batch.Source.Labels, 2)
				require.Len(t, batch.Entries, 2, "Expected two log entries in the batch")
				// Verifying the first log entry data
				require.Equal(t, "First log message", string(batch.Entries[0].Data))
				// Verifying the second log entry data
				require.Equal(t, "Second log message", string(batch.Entries[1].Data))
			},
		},
```

exporter/chronicleexporter/marshal_test.go
dpaasman00, line 1: I think there are some more test cases that can be added to thoroughly test this new functionality and potential edge cases, such as the multiple log records example I gave in an earlier comment.

exporter/chronicleexporter/marshal.go
dpaasman00, line 1: This is a general, very nit comment. Unless you feel strongly about it, there are several deleted new lines in this PR that I think should be restored to help with readability. Just take a look through the diff here to find them.

exporter/chronicleexporter/marshal.go
```diff
@@ -61,7 +61,6 @@ type protoMarshaler struct {
 }
 
 func newProtoMarshaler(cfg Config, teleSettings component.TelemetrySettings, labels []*api.Label, customerID []byte) (*protoMarshaler, error) {
-
 	return &protoMarshaler{
 		startTime:    time.Now(),
```
dpaasman00, line 65: Contrary to the deleted newlines comment I left, this is a valid deleted new line and should stay deleted.

dpaasman00 (changes requested):
First pass and generally looks good. Somethings to address however.

exporter/chronicleexporter/marshal.go
```diff
@@ -177,6 +215,35 @@ func (m *protoMarshaler) getLogType(ctx context.Context, logRecord plog.LogRecor
...
+	}
+	ingestionLabels, err := m.getRawNestedFields(chronicleIngestionLabelsPrefix, logRecord)
+	if err != nil {
+		return configLabels, fmt.Errorf("get chronicle ingestion labels: %w", err)
+	}
+	if len(ingestionLabels) != 0 {
+		return ingestionLabels, nil
+	}
+	return configLabels, nil
+}
```
dpaasman00, line 245: Again, the return value isn't used when there's an error, so return an empty slice and simplify this code a bit
```suggestion
func (m *protoMarshaler) getIngestionLabels(logRecord plog.LogRecord) ([]*api.Label, error) {
	// check for labels in attributes["chronicle_ingestion_labels"]
	ingestionLabels, err := m.getRawNestedFields(chronicleIngestionLabelsPrefix, logRecord)
	if err != nil {
		return []*api.Label{}, fmt.Errorf("get chronicle ingestion labels: %w", err)
	}
	if len(ingestionLabels) != 0 {
		return ingestionLabels, nil
	}

	// use labels defined in config if needed
	configLabels := make([]*api.Label, 0)
	for key, value := range m.cfg.IngestionLabels {
		configLabels = append(configLabels, &api.Label{
			Key:   key,
			Value: value,
		})
	}

	return configLabels, nil
}
```

exporter/chronicleexporter/marshal.go
```diff
@@ -207,10 +274,32 @@ func (m *protoMarshaler) getRawField(ctx context.Context, field string, logRecor
...
+			nestedFields = append(nestedFields, &api.Label{
+				Key:   key,
+				Value: value.AsString(),
+			})
+		}
+		return true
+	})
+	if len(nestedFields) == 0 {
+		return nil, nil
+	}
```
dpaasman00, line 290: No need to return nil instead of an empty `[]*api.Label`

```suggestion
```

exporter/chronicleexporter/marshal.go
dpaasman00, line 1: I know your PR doesn't touch the `getLogType()` function, but in the spirit of adding 2 similar funcs, can you update `getLogType()` to return an empty string in the 2 cases it returns an error, because the returned string does not get used when there's an error

exporter/chronicleexporter/README.md
```diff
@@ -48,8 +48,9 @@ currently supported log types are:
 - windows_event.system
 - sql_server
 
-
 If the `attributes["chronicle_log_type"]` field is present in the log, we will use its value in the payload instead of the automatic detection or the `log_type` in the config.
+If the `attributes["chronicle_namespace"]` field is present in the log, we will use its value in the payload.
+If there are nested fields in `attributes["chronicle_ingestion_label"]`, we will use the values in the payload.
```
dpaasman00, line 53: These should be broken out into a separate section from the "Log Type" section they're in currently, something like "Namespace & Ingestion Labels". Also state that if the attribute fields are present, they are used regardless of the value in the config.

exporter/chronicleexporter/marshal.go
```diff
@@ -113,14 +114,54 @@ func (m *protoMarshaler) extractRawLogs(ctx context.Context, ld plog.Logs) (map[
 					Data:           []byte(rawLog),
 				}
 				entries[logType] = append(entries[logType], entry)
+				// each logType maps to exactly 1 namespace value
+				if _, ok := namespaceMap[logType]; !ok {
+					namespaceMap[logType] = namespace
+				}
```
dpaasman00, line 120: This should also check if the namespace is empty before adding
```suggestion
                         if namespace != "" {
				// each logType maps to exactly 1 namespace value
				if _, ok := namespaceMap[logType]; !ok {
					namespaceMap[logType] = namespace
				}
			}
```

dpaasman00 (changes requested):
Couple more thoughts, otherwise this is looking good. Going to test next.

dpaasman00 (approved):
This has been working well for me.

PR #1940, opened 2024-11-04
Add AIX Support to the collector

Dylan-M:
- `feat(AIX):` = Enable building and deploying the collector for AIX

### Proposed Change
Add AIX support to the collector.

##### Checklist
- [ ] Changes are tested
- [x] CI has passed


mrsillydog:
@claude review

updater/internal/path/path_unix.go
mrsillydog, line 1: I think this and the two files below are all in the wrong place. `updater` is not supposed to be a thing on V2.

PR #1946, opened 2024-11-06
feat: Support manager config environment variable expansion

mrsillydog:
### Proposed Change
- Move from using standard os and yaml methods for reading and unmarshalling manager configuration files to the opentelemetry confmap library
- Add the envprovider factory to allow for environment variable usage within the manager configuration files.
- Minor cleanup of unused variables and attributes within the modified test file

##### Checklist
- [X] Changes are tested
- [X] CI has passed


opamp/config_test.go
```diff
@@ -188,15 +188,17 @@ func TestToTLS(t *testing.T) {
 				}
 
 				expectedConfig := tls.Config{
-					MinVersion: tls.VersionTLS12,
+					// MinVersion: tls.VersionTLS12,
```
jsirianni, line 191: Was there an issue w/ this test locally, commented out temporarily?

jsirianni (changes requested):
Working great. One question in tests.

opamp/config_test.go
mrsillydog: My IDE simply flagged the sections I've commented out as unused - they're set, but then never referenced again (in this particular test, the only `require.Equal` call just references `expectedConfig.Certificates`, and the rest of `expectedConfig` is ignored). My take was that commenting out the unused aspects was a middle ground between simply removing all the unused sections (which might be valuable in the future, if this test is ever improved to account for them) and letting the scope of this ticket creep to actually improve the test now. Happy to do either of them if your take is different from mine.

jsirianni: We synced n this offline :+1:

jsirianni (approved):
Looks great assuming CI passes :+1:

PR #1965, opened 2024-11-13
feat: SSAPI Pagination (BPS-276)

Caleb-Hurshman:
### Proposed Change

Paginate Splunk search result responses. Add an `event_batch_size` config option to control how many events (search results) to fetch at a time.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/splunksearchapireceiver/config.go
```diff
@@ -32,6 +32,7 @@ type Config struct {
 	Username                string        `mapstructure:"splunk_username"`
 	Password                string        `mapstructure:"splunk_password"`
 	Searches                []Search      `mapstructure:"searches"`
+	EventBatchSize          int           `mapstructure:"event_batch_size"`
```
schmikei, line 35: Should this be on the search configuration itself?

receiver/splunksearchapireceiver/api.go
```diff
@@ -118,6 +119,7 @@ func (ssapir *splunksearchapireceiver) getSearchResults(config *Config, sid stri
 	if err != nil {
 		return SearchResults{}, fmt.Errorf("failed to unmarshal search job results: %v", err)
 	}
+	fmt.Println("Init offset: ", searchResults.InitOffset)
```
schmikei, line 122:
```suggestion
```

receiver/splunksearchapireceiver/model.go
```diff
@@ -49,7 +49,8 @@ type List struct {
 
 // SearchResults struct to represent the JSON response from Splunk search results endpoint
 type SearchResults struct {
```
schmikei, line 51: I wonder could we link to documentation on these endpoints in comments to at least capture the full set of fields that may be useful?

receiver/splunksearchapireceiver/receiver.go
```diff
@@ -67,66 +67,83 @@ func (ssapir *splunksearchapireceiver) runQueries(ctx context.Context) error {
...
+				// convert log timestamp to ISO8601 (UTC() makes RFC3339 into ISO8601)
+				logTimestamp, err := time.Parse(time.RFC3339, splunkLog.Time)
+				if err != nil {
+					ssapir.logger.Error("error parsing log timestamp", zap.Error(err))
+					break
+				}
+				if logTimestamp.UTC().After(latestTime.UTC()) {
+					ssapir.logger.Info("skipping log entry - timestamp after latestTime", zap.Time("time", logTimestamp.UTC()), zap.Time("latestTime", latestTime.UTC()))
+					// logger will only log up to 10 times for a given code block, known weird behavior
+					// TODO: Consider breaking, assuming all logs are in order
```
schmikei, line 107: Fair thing to look in to; we'd need to make sure that the events being retrieved from Splunk are in order, which I'm not confident to say one way or another on quite yet

receiver/splunksearchapireceiver/receiver.go
```diff
@@ -67,66 +67,83 @@ func (ssapir *splunksearchapireceiver) runQueries(ctx context.Context) error {
...
-		ssapir.logger.Info("fetching search results")
-		results, err := ssapir.getSplunkSearchResults(searchID)
-		if err != nil {
-			ssapir.logger.Error("error fetching search results", zap.Error(err))
-		}
-		ssapir.logger.Info("search results fetched", zap.Int("num_results", len(results.Results)))
+		var resultCountTracker = 0 // track number of results exported
+		var offset = 0             // offset for pagination
+		var limitReached = false
+		for {
```
schmikei, line 73: Let's set up some time to vscode share a solution on polling; feel like having an infinite for loop in this case could actually be improved if we start looking at context.Context

receiver/splunksearchapireceiver/config.go
Caleb-Hurshman: I guess. If this is something that a user would want to change for different queries, then sure.

receiver/splunksearchapireceiver/receiver.go
Caleb-Hurshman: Took some time to look into this, the pattern I'm seeing is that events are returned latest to earliest, which makes sense based on how the Splunk search UI works. So it would be correct to break once a log is earlier than the earliest time.

schmikei (approved):

PR #1968, opened 2024-11-14
feat: SSAPI pass timestamps in Splunk query (BPS-283)

Caleb-Hurshman:
### Proposed Change

Instead of passing the query from the config as-is, use the earliest and latest time config fields to specify a `starttime` and `endtime` on the created query. This will eliminate unnecessary API calls and remove a lot of the post-processing done by the receiver.

Example output:
```
{"level":"info","ts":"2024-11-14T14:34:26.854-0500","caller":"splunksearchapireceiver/receiver.go:155","msg":"creating search","kind":"receiver","name":"splunksearchapi","data_type":"logs","query":"search index=otel starttime=\"2024-11-14T00:00:00.000Z\" endtime=\"2024-11-14T20:00:00.000Z\" timeformat=\"%Y-%m-%dT%H:%M:%S\""}
{"level":"info","ts":"2024-11-14T14:34:32.094-0500","caller":"splunksearchapireceiver/receiver.go:142","msg":"search completed","kind":"receiver","name":"splunksearchapi","data_type":"logs"}
{"level":"info","ts":"2024-11-14T14:34:32.094-0500","caller":"splunksearchapireceiver/receiver.go:67","msg":"fetching search results","kind":"receiver","name":"splunksearchapi","data_type":"logs"}
{"level":"info","ts":"2024-11-14T14:34:32.144-0500","caller":"splunksearchapireceiver/receiver.go:72","msg":"search results fetched","kind":"receiver","name":"splunksearchapi","data_type":"logs","num_results":20}
```
Screenshot 2024-11-14 at 3 06 52 PM

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/splunksearchapireceiver/client.go
+++ b/receiver/splunksearchapireceiver/client.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 
 	"go.opentelemetry.io/collector/component"
 	"go.uber.org/zap"
@@ -59,7 +60,7 @@ func newSplunkSearchAPIClient(ctx context.Context, settings component.TelemetryS
 func (c defaultSplunkSearchAPIClient) CreateSearchJob(search string) (CreateJobResponse, error) {
 	endpoint := fmt.Sprintf("%s/services/search/jobs", c.endpoint)
 
-	reqBody := fmt.Sprintf(`search=%s`, search)
+	reqBody := fmt.Sprintf(`search=%s`, url.QueryEscape(search))
 	req, err := http.NewRequest("POST", endpoint, bytes.NewBuffer([]byte(reqBody)))
 	if err != nil {
 		return CreateJobResponse{}, err
--- a/receiver/splunksearchapireceiver/config.go
+++ b/receiver/splunksearchapireceiver/config.go
@@ -72,6 +72,10 @@ func (cfg *Config) Validate() error {
 			return errNonStandaloneSearchQuery
 		}
 
+		if strings.Contains(search.Query, "earliest=") || strings.Contains(search.Query, "latest=") {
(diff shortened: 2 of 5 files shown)


receiver/splunksearchapireceiver/client.go
```diff
@@ -59,7 +60,7 @@ func newSplunkSearchAPIClient(ctx context.Context, settings component.TelemetryS
 func (c defaultSplunkSearchAPIClient) CreateSearchJob(search string) (CreateJobResponse, error) {
 	endpoint := fmt.Sprintf("%s/services/search/jobs", c.endpoint)
 
-	reqBody := fmt.Sprintf(`search=%s`, search)
+	reqBody := fmt.Sprintf(`search=%s`, url.QueryEscape(search))
```
schmikei, line 63: Are we sure we want to be url.QueryEscaping in this circumstance? I thought that was for query parameters; Do you have a case that caused this to be necessary?

Caleb-Hurshman: @schmikei from the example in the [API docs](https://docs.splunk.com/Documentation/Splunk/9.3.1/RESTREF/RESTsearch#search.2Fjobs):
```
curl -k -u admin:pass https://localhost:8089/servicesNS/admin/search/search/jobs --data-urlencode search="search index=_internal source=*/metrics.log"
```
That's the only example that uses --data-urlencode, but since we'll be adding the time fields to every search, there's not a case where we wouldn't need it to get those right (unless there's some other way)

schmikei: Looked into this a bit further and yep it seems like this is the correct approach when using the Splunk search REST API https://community.splunk.com/t5/Splunk-Enterprise/rest-api-search-not-working-when-using-cs-uri-stem-cs-uri-query/m-p/532759

schmikei (approved):

PR #1969, opened 2024-11-15
feat: SSAPI Checkpointing (BPS-277)

Caleb-Hurshman:
### Proposed Change

Add checkpointing capabilities to the SSAPI receiver. 
The receiver now requires a storage extension to be configured in order to run.
Store the offset of the last successfully exported log as the checkpoint value.
Upon starting the receiver, load a checkpoint from the storage extension in case a previous run failed.

**1st run (Empty storage extn)**:
```
{"level":"info","ts":"2024-11-15T11:33:03.831-0500","caller":"splunksearchapireceiver/receiver.go:260","msg":"no checkpoint found","kind":"receiver","name":"splunksearchapi","data_type":"logs"}
{"level":"info","ts":"2024-11-15T11:33:09.094-0500","caller":"splunksearchapireceiver/receiver.go:111","msg":"search results fetched","kind":"receiver","name":"splunksearchapi","data_type":"logs","num_results":8}
offset 8
{"level":"info","ts":"2024-11-15T11:33:09.096-0500","caller":"splunksearchapireceiver/receiver.go:106","msg":"fetching search results","kind":"receiver","name":"splunksearchapi","data_type":"logs"}
{"level":"info","ts":"2024-11-15T11:33:09.136-0500","caller":"splunksearchapireceiver/receiver.go:111","msg":"search results fetched","kind":"receiver","name":"splunksearchapi","data_type":"logs","num_results":8}
offset 16
{"level":"info","ts":"2024-11-15T11:33:09.136-0500","caller":"splunksearchapireceiver/receiver.go:106","msg":"fetching search results","kind":"receiver","name":"splunksearchapi","data_type":"logs"}
{"level":"info","ts":"2024-11-15T11:33:09.173-0500","caller":"splunksearchapireceiver/receiver.go:111","msg":"search results fetched","kind":"receiver","name":"splunksearchapi","data_type":"logs","num_results":4}
offset 20
{"level":"info","ts":"2024-11-15T11:33:09.174-0500","caller":"splunksearchapireceiver/receiver.go:185","msg":"all search results exported","kind":"receiver","name":"splunksearchapi","data_type":"logs","query":"search index=otel earliest=-1d latest=now","total results":20}
```

**2nd run (Storage extension not cleared out to simulate an existing offset checkpoint)**:
```
{"level":"info","ts":"2024-11-15T11:33:35.808-0500","caller":"splunksearchapireceiver/receiver.go:72","msg":"found offset checkpoint in storage extension","kind":"receiver","name":"splunksearchapi","data_type":"logs","offset":20}
{"level":"info","ts":"2024-11-15T11:33:41.161-0500","caller":"splunksearchapireceiver/receiver.go:111","msg":"search results fetched","kind":"receiver","name":"splunksearchapi","data_type":"logs","num_results":0}
offset 20
{"level":"info","ts":"2024-11-15T11:33:41.162-0500","caller":"splunksearchapireceiver/receiver.go:185","msg":"all search results exported","kind":"receiver","name":"splunksearchapi","data_type":"logs","query":"search index=otel earliest=-1d latest=now","total results":0}
```

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/splunksearchapireceiver/receiver.go
```diff
@@ -201,3 +241,26 @@ func (ssapir *splunksearchapireceiver) getSplunkSearchResults(sid string, offset
...
+	marshalBytes, err := ssapir.storageClient.Get(ctx, eventStorageKey)
+	if err != nil {
+		ssapir.logger.Error("failed to read checkpoint", zap.Error(err))
+		return
+	}
+	if marshalBytes == nil {
+		ssapir.logger.Info("no checkpoint found")
+		return
+	}
+	if err = json.Unmarshal(marshalBytes, ssapir.checkpointRecord); err != nil {
```
schmikei, line 263: shouldn't we bomb out if we fail to unmarshal a checkpoint?

receiver/splunksearchapireceiver/receiver.go
```diff
@@ -126,23 +158,31 @@ func (ssapir *splunksearchapireceiver) runQueries(ctx context.Context) error {
...
+				// error from down the pipeline, freak out
 				ssapir.logger.Error("error consuming logs", zap.Error(err))
 			}
+			// last batch of logs has been successfully exported
+			exportedEvents += logs.ResourceLogs().Len()
+			offset += len(results.Results)
+
+			// update checkpoint
+			ssapir.checkpointRecord.Offset = offset
+			fmt.Println("offset", offset)
```
schmikei, line 170: Probably don't need this println merged
```suggestion
```

receiver/splunksearchapireceiver/receiver.go
```diff
@@ -48,12 +58,34 @@ func (ssapir *splunksearchapireceiver) Start(ctx context.Context, host component
...
+	}
+
 	go ssapir.runQueries(ctx)
 	return nil
 }
 
-func (ssapir *splunksearchapireceiver) Shutdown(_ context.Context) error {
-	return nil
+func (ssapir *splunksearchapireceiver) Shutdown(ctx context.Context) error {
+	ssapir.logger.Info("shutting down logs receiver")
```
schmikei, line 81:
```suggestion
	ssapir.logger.Debug("shutting down logs receiver")
```

receiver/splunksearchapireceiver/receiver.go
```diff
@@ -126,23 +158,31 @@ func (ssapir *splunksearchapireceiver) runQueries(ctx context.Context) error {
...
 			if len(results.Results) < search.EventBatchSize {
-				exportedEvents += len(results.Results)
 				break
 			}
-			exportedEvents += logs.ResourceLogs().Len()
-			offset += len(results.Results)
+
 		}
-		ssapir.logger.Info("search results exported", zap.String("query", search.Query), zap.Int("total results", exportedEvents))
+		ssapir.logger.Info("all search results exported", zap.String("query", search.Query), zap.Int("total results", exportedEvents))
```
schmikei, line 185:
```suggestion
		ssapir.logger.Debug("all search results exported", zap.String("query", search.Query), zap.Int("total results", exportedEvents))
```

receiver/splunksearchapireceiver/model.go
```diff
@@ -58,3 +58,8 @@ type SearchResultsResponse struct {
 		Time string `json:"_time"`
 	} `json:"results"`
 }
+
+// EventRecord struct stores the offset of the last event exported successfully
+type EventRecord struct {
+	Offset int `json:"offset"`
```
schmikei, line 64: Curious last time I've used a storage extension; we've used a timestamp to manage the last processed event time... Do you think that'd be more bulletproof than using the paginated offset?

Caleb-Hurshman: Using the offset instead of the timestamp was a piece of feedback from Dan on the spec. Looking back at his comment, he thought it would be easier to use the offset. Whether it's as bulletproof is a good question, but I assumed he wouldn't sacrifice the reliability of the receiver for ease. Curious to hear more of your perspective.

receiver/splunksearchapireceiver/receiver.go
schmikei: We shouldn't Fatal because that would stop the collector process entirely. I'd prefer seeing the error propogated back up. 

We do this so that other components can have clean Shutdown calls.

schmikei:
Yeah I'm thinking there's a couple scenarios where we need to think about:

1. Test a fail condition (shut off network connection mid processing) change configuration to do a different search; does it still use an offset that makes sense? Seems like we need to start tracking this by search
2. Test a network failure from Splunk on retrieving paginated results (does our checkpoint pretty much work as we expect)

At least those are the two main things I'm curious if this solution fully solves at the moment.

receiver/splunksearchapireceiver/config.go
```diff
@@ -60,6 +62,10 @@ func (cfg *Config) Validate() error {
 		return errors.New("at least one search must be provided")
 	}
 
+	if cfg.StorageID == nil {
+		return errors.New("storage configuration must be provided")
```
schmikei, line 66:
```suggestion
		return errors.New("storage configuration is required for this receiver")
```

Caleb-Hurshman:
> 2. Test a network failure from Splunk on retrieving paginated results (does our checkpoint pretty much work as we expect)

How would you test this?

schmikei:
> > 2. Test a network failure from Splunk on retrieving paginated results (does our checkpoint pretty much work as we expect)
> 
> How would you test this?

My recommendation would be to use https://pkg.go.dev/net/http/httptest and simulate one in code as a test; otherwise you can use a non-routable IP address: https://stackoverflow.com/questions/100841/artificially-create-a-connection-timeout-error

schmikei (commented):
There's one clarificationI want to understand when the `ConsumeLogs` call breaks;

receiver/splunksearchapireceiver/receiver.go
```diff
@@ -125,23 +169,30 @@ func (ssapir *splunksearchapireceiver) runQueries(ctx context.Context) error {
 			// pass logs, wait for exporter to confirm successful export to GCP
 			err = ssapir.logsConsumer.ConsumeLogs(ctx, logs)
 			if err != nil {
-				// Error from down the pipeline, freak out
+				// error from down the pipeline, freak out
 				ssapir.logger.Error("error consuming logs", zap.Error(err))
 			}
+			// last batch of logs has been successfully exported
+			exportedEvents += logs.ResourceLogs().Len()
```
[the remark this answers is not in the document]
Caleb-Hurshman: I agree that the process should break if `ConsumeLogs` errors, but I don't think the checkpoint should be updated. 
In my head, the flow is like this:
1. Previous batch consumed successfully (new offset written, which is the offset of the last log exported +1)
2. Next batch obtained from Splunk using the offset
3. Batch fails to export, `ConsumeLogs` errors and the receiver process exits (no partial writes to GCP by default, no checkpoint written)
4. User addresses some issue with their GCP instance
5. User restarts the receiver, which grabs the last written offset, essentially restarting from step 2

Feel free to add on if I'm missing something.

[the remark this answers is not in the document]
schmikei: Yeah I think my brain was on the thread that if some of the batch gets processed, but some others get dropped; I think that gets hairy for checkpointing for what is likely a pretty big edge case...

receiver/splunksearchapireceiver/go.mod
```diff
@@ -3,34 +3,65 @@ module github.com/open-telemetry/opentelemtry-collector-contrib/receiver/splunks
...
 
 require (
+	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza v0.113.0
 	github.com/stretchr/testify v1.9.0
 	go.opentelemetry.io/collector/component v0.113.0
+	go.opentelemetry.io/collector/confmap v1.20.0
 	go.opentelemetry.io/collector/consumer v0.113.0
+	go.opentelemetry.io/collector/consumer/consumertest v0.113.0
+	go.opentelemetry.io/collector/extension/experimental/storage v0.113.0
+	go.opentelemetry.io/collector/filter v0.114.0
```
schmikei, line 13: is this supposed to v0.114.0? I'd imagine a merge from main might smooth out some of these dependency conflicts?

receiver/splunksearchapireceiver/receiver.go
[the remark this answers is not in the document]
Caleb-Hurshman: Makes sense. It was my impression that the GCP exporter wouldn't partially write a batch, but I might be off on that. I don't know how to handle that case right now, but that's something I wanted to revisit when we get a chance to test. It depends on how much information the error returns. If there's no info on which log(s) failed, then we're kinda hosed.

receiver/splunksearchapireceiver/go.mod
Caleb-Hurshman: I'm not seeing any conflicts personally. Tidy-ing doesn't update these deps. Should I rebase the feature branch to the latest release?

schmikei: Perhaps @dpaasman00 would know better; but I'd imagine that we probably want `go.opentelemetry.io/collector/filter v0.113.0` rather than v0.114.0 but since this is not being merged into main for a little while; I'd imagine its not worth worrying about too much until we want to start entertaining a main merge.

schmikei (approved):
Think its good for the feature branch, may need some further testing.

Nice work 🚀

PR #1970, opened 2024-11-15
fix: Add `-r` flag back to unix install script

dpaasman00:
### Proposed Change

Flag got accidentally removed during rebase so adding it back in. Verified -r is present in macos script. Probably trigger another release afterwards.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/install/install_unix.sh
+++ b/scripts/install/install_unix.sh
@@ -937,6 +937,10 @@ main() {
         base_url=$2
         shift 2
         ;;
+      -r | --uninstall)
+        uninstall
+        exit 0
+        ;;
       -h | --help)
         usage
         exit 0


scripts/install/install_unix.sh
```diff
@@ -937,6 +937,10 @@ main() {
         base_url=$2
         shift 2
         ;;
+      -r | --uninstall)
```
mrsillydog, line 940: Should this be described in the usage() function as well?

dpaasman00: I'll do a follow up PR to branch `release/v1.65.0` bc it's not in there either and should be. After the next release this branch will get rebased and include the change.

mrsillydog (approved):

PR #1982, opened 2024-11-20
feat: SSAPI Unit Tests (BPS-272)

Caleb-Hurshman:
### Proposed Change

Add more unit tests for the Splunk Search API Receiver.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


receiver/splunksearchapireceiver/receiver_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	jobResponse.Content.Dict.Keys[0].Value = "RUNNING"
+	done = ssapireceiver.isSearchCompleted(jobResponse)
+	require.False(t, done)
+
+	done = ssapireceiver.isSearchCompleted(emptyResponse)
+	require.False(t, done)
+}
+
+func TestInitCheckpoint(t *testing.T) {
```
schmikei, line 89: Are these supposed to be implemented?

Caleb-Hurshman: Somehow those changes didn't get added in my last commit. Should be good now.

receiver/splunksearchapireceiver/model.go
```diff
@@ -20,13 +20,16 @@ type CreateJobResponse struct {
...
 
-// JobStatusResponse struct to represent the XML response from Splunk job status endpoint
+// SearchStatusResponse struct to represent the XML response from Splunk job status endpoint
 // https://docs.splunk.com/Documentation/Splunk/9.3.1/RESTREF/RESTsearch#search.2Fjobs.2F.7Bsearch_id.7D
-type JobStatusResponse struct {
-	Content struct {
-		Type string `xml:"type,attr"`
-		Dict Dict   `xml:"dict"`
-	} `xml:"content"`
+type SearchStatusResponse struct {
```
schmikei, line 25:
```suggestion
type SearchJobStatusResponse struct {
```

nit: I think this is more indicative of a name. not required just something I think makes it more readable

receiver/splunksearchapireceiver/model.go
```diff
@@ -20,13 +20,16 @@ type CreateJobResponse struct {
...
+	Content Content `xml:"content"`
+}
+
+// Content struct to represent <content> elements
+type Content struct {
```
schmikei, line 30:
```suggestion
type SearchJobContent struct {
```

Along similar veins

receiver/splunksearchapireceiver/testdata/logs/testPollJobStatus/input.xml
```diff
@@ -0,0 +1,212 @@
...
+        </s:dict>
+      </s:key>
+      <s:key name="searchProviders">
+        <s:list>
+          <s:item>mbp15.splunk.com</s:item>
+        </s:list>
+      </s:key>
+    </s:dict>
+  </content>
+</entry>
```
schmikei, line 212:
```suggestion
</entry>

```
nit

receiver/splunksearchapireceiver/go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	go.opentelemetry.io/collector/extension/experimental/storage v0.113.0
 	go.opentelemetry.io/collector/pdata v1.19.0
 	go.opentelemetry.io/collector/receiver v0.113.0
+	go.opentelemetry.io/collector/receiver/receivertest v0.113.0
```
schmikei, line 14: Did this not rmake any go.sum changes? I'd imagine it would. Would you mind re-running `go mod tidy` to just double check?

receiver/splunksearchapireceiver/client_test.go
```diff
@@ -0,0 +1,142 @@
...
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	server     = newMockServer()
```
schmikei, line 28: nit I think I'd like to see these defined within the test functions themselves so other test files can use the names `server` and `testClient` without worrying about variable shadowing.

receiver/splunksearchapireceiver/testdata/logs/testPollJobStatus/input.xml
```diff
@@ -0,0 +1,212 @@
...
+  <link href="/services/search/jobs/mysearch_02151949/control" rel="control"/>
+  <author>
+    <name>admin</name>
+  </author>
+  <content type="text/xml">
+    <s:dict>
+      <s:key name="cursorTime">1969-12-31T16:00:00.000-08:00</s:key>
+      <s:key name="delegate"></s:key>
+      <s:key name="diskUsage">2174976</s:key>
+      <s:key name="dispatchState">DONE</s:key>
```
schmikei, line 25: I think having a dispatchState as not done will be another important test for any future changes.

receiver/splunksearchapireceiver/client.go
```diff
@@ -60,6 +61,10 @@ func newSplunkSearchAPIClient(ctx context.Context, settings component.TelemetryS
 func (c defaultSplunkSearchAPIClient) CreateSearchJob(search string) (CreateJobResponse, error) {
 	endpoint := fmt.Sprintf("%s/services/search/jobs", c.endpoint)
 
+	if !strings.Contains(search, "starttime=") || !strings.Contains(search, "endtime=") || !strings.Contains(search, "timeformat=") {
```
schmikei, line 64: Might be overkill but in splunk, are these case-insensitive? I.e. I believe you  can do `Starttime="<whatever>` and its still valid

For the sake of correctness I'm curious if we need to make these error checks non case-sensitive? 

I know that we're inputting these parameters so its not too big of a deal, so maybe a comment would suffice in this situation

receiver/splunksearchapireceiver/go.mod
Caleb-Hurshman: Apparently not 🤷‍♂️

receiver/splunksearchapireceiver/client.go
Caleb-Hurshman: Good catch, I didn't know that about Splunk.

PR #1991, opened 2024-11-22
fix: Set CURR_VERSION for new v2 release-prep calls

mrsillydog:
### Proposed Change
Set CURR_VERSION flag so that `make release-prep` provides the correct version in VERSION.txt.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dpaasman00 (approved):
Fixes Windows and still working well on other platforms.

PR #2022, opened 2024-12-03
Minimize copying and complexity

djaglowski:


schmikei (approved):
The only thing I'm thinking about is if we can improve the remove logic (may not be a nice way without making a copy)/ordering of results. However I think the efficiency is probably worth more in this and those could probably come later

Thank you Dan!

PR #2023, opened 2024-12-03
feat: SSAPI Exporter Integration Test (BPS-279)

Caleb-Hurshman:
### Proposed Change
Add a new integration test to simulate a failure from the exporter that would travel back up the pipeline while blocking after calling `ConsumeLogs`. 

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/splunksearchapireceiver/receiver.go
+++ b/receiver/splunksearchapireceiver/receiver.go
@@ -111,6 +111,9 @@ func (ssapir *splunksearchapireceiver) Shutdown(ctx context.Context) error {
 
 func (ssapir *splunksearchapireceiver) runQueries(ctx context.Context) error {
 	for _, search := range ssapir.config.Searches {
+		// set current search query
+		ssapir.checkpointRecord.Search = search.Query
+
 		// create search in Splunk
 		searchID, err := ssapir.createSplunkSearch(search)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


Caleb-Hurshman:
Having trouble reproducing these test failures on my branch. LMK if you have any ideas @schmikei

receiver/splunksearchapireceiver/receiver.go
```diff
@@ -111,6 +111,9 @@ func (ssapir *splunksearchapireceiver) Shutdown(ctx context.Context) error {
 
 func (ssapir *splunksearchapireceiver) runQueries(ctx context.Context) error {
```
StefanKurek, line 112: Unless I'm missing something (which I may be. I haven't looked behind the failing test and this function), I believe your test failures are related to this function.

It looks like the logic here is to make your query, wait for it, and then start retrieving the search results and processing them in a loop until the # of search results are less than the max possible returned in a single page. If that's the case you quit the infinite loop, otherwise you try to get another page of results.

Now what happens if it was your last page of results, but the number of results was the max number that can fit within a page (as is with the failing test)? It seems like you try to get search results but get an error because there are none. 

Hopefully I'm right or at least half right and this puts you on the right track to fixing the CI issues.

StefanKurek (changes requested):

schmikei (approved):

PR #2025, opened 2024-12-03
chore: Rename to bindplane-agent

dpaasman00:
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (changes requested):
1. Container images fail with:

> could not stat agent::executable path: stat /collector/observiq-otel-collector: no such file or directory

Probably need to update the binary path in the Dockerfiles.

2. Linux packages fail with conflict error.

> Error: Transaction test error:
  file /opt/opentelemetry-java-contrib-jmx-metrics.jar from install of bindplane-agent-1.66.0~SNAPSHOT_09452249-1.x86_64 conflicts with file from package observiq-otel-collector-2.0.0~alpha.5-1.x86_64

We probably want to remove the conflict from Goreleaser's nfpm config so we can allow users to deploy v2 side by side v1.

PR #2033, opened 2024-12-09
feat: Enforce request maximum size and number of logs

mrsillydog:
Proposed Change
Google SecOps (Chronicle) limits API log ingestion in two different ways - first, in uncompressed request size, and secondly in number of logs per request. This PR changes the implementation of the chronicle exporter to respect these limits by checking against them before sending, and then splitting up the batched request logging data if either limit is exceeded. These limits are configurable on the SecOps backend, and therefore now are also configurable in the exporter.

Possible concerns:

- Is proto.Size() an accurate measurement for the HTTP protocol workflow? It was confirmed as accurate for the GRPC protocol workflow.
- What is a reasonable default size limit for HTTP and GRPC? The documentation lists it as 1MB for the GRPC endpoint, undocumented for the HTTP endpoint. Manual testing confirmed our configured limit is 5 MiB via GRPC, but substantially higher (unclear exactly what the limit is) via the v1alpha HTTP endpoint.

##### Checklist
- [X] Changes are tested
- [X] CI has passed


exporter/chronicleexporter/marshal.go
```diff
@@ -411,25 +412,61 @@ func (m *protoMarshaler) constructPayloads(rawLogs map[string][]*api.LogEntry, n
...
+		// re-enforce max size restriction on each half
+		enforcedRequest := m.enforceMaximumsGRPCRequest(request)
+		enforcedOtherHalfRequest := m.enforceMaximumsGRPCRequest(otherHalfRequest)
+
+		return append(enforcedRequest, enforcedOtherHalfRequest...)
+	}
+
+	return []*api.BatchCreateLogsRequest{
+		request,
+	}
```
djaglowski, line 450: readability nit: in Go it's typical to handle the simple case and return early. This allows the code to stay flat and makes it easier to keep track of assumptions
```suggestion
	if size <= m.cfg.BatchRequestSizeLimitGRPC && len(request.Batch.Entries) <= m.cfg.BatchLogCountLimitGRPC {
		return []*api.BatchCreateLogsRequest{request}
	}
	
	if len(request.Batch.Entries) == 1 {
		m.teleSettings.Logger.Error("Single entry exceeds max request size. Dropping entry", zap.Int("size", size))
		return []*api.BatchCreateLogsRequest{}
	}

	// split request into two
	entries := request.Batch.Entries
	mid := len(entries) / 2
	leftHalf := entries[:mid]
	rightHalf := entries[mid:]

	request.Batch.Entries = leftHalf
	otherHalfRequest := m.buildGRPCRequest(rightHalf, request.Batch.LogType, request.Batch.Source.Namespace, request.Batch.Source.Labels)
	// re-enforce max size restriction on each half
	enforcedRequest := m.enforceMaximumsGRPCRequest(request)
	enforcedOtherHalfRequest := m.enforceMaximumsGRPCRequest(otherHalfRequest)

	return append(enforcedRequest, enforcedOtherHalfRequest...)
```

exporter/chronicleexporter/marshal.go
```diff
@@ -482,26 +519,60 @@ func buildForwarderString(cfg Config) string {
...
+			payloads[logType] = m.enforceMaximumsHTTPRequest(request)
 		}
 	}
 	return payloads
 }
+
+func (m *protoMarshaler) enforceMaximumsHTTPRequest(request *api.ImportLogsRequest) []*api.ImportLogsRequest {
+	size := proto.Size(request)
+	logs := request.GetInlineSource().Logs
+	if size > m.cfg.BatchRequestSizeLimitHTTP || len(logs) > m.cfg.BatchLogCountLimitHTTP {
```
djaglowski, line 538: I'm still working out how size is measured for HTTP requests, but I don't think it's `proto.Size`, and therefore I don't think it's a cheap value to calculate. Therefore, I would suggest handling # of logs per request first, and then checking size afterwards. Splitting logic can be the shared, if the condition can be abstracted.

mrsillydog: If it's likely not cheap to calculate, maybe it makes sense to do the batch-splitting based on the API response rather than pre-calculating it during payload construction, as we discussed?

djaglowski (approved):

exporter/chronicleexporter/marshal.go
mrsillydog: For posterity, Dan and I discussed this further offline and decided that 1) proto.Size was as reliable as any other measurement method, so with cautious configuration it would work 2) trying to detect when a "too large" error was not feasible with the current state of the v1alpha HTTP API, so any enhancement in that sphere will have to wait for improved API usability.

exporter/chronicleexporter/config.go
```diff
@@ -85,6 +85,26 @@ type Config struct {
...
+
+	// BatchLogCountLimitHTTP is the maximum number of logs that can be sent in a single batch to Chronicle via the HTTP protocol
+	// This field is defaulted to 1000, as that is the default Chronicle backend limit.
+	// All batched logs beyond the backend limit will be dropped. Any batches with more logs than this limit will be split into multiple batches
+	BatchLogCountLimitHTTP int `mapstructure:"batch_log_count_limit_grpc"`
+
+	// BatchRequestSizeLimitHTTP is the maximum batch request size, in bytes, that can be sent to Chronicle via the HTTP protocol
+	// This field is defaulted to 1048576 as that is the default Chronicle backend limit
+	// Setting this option to a value above the Chronicle backend limit may result in rejected log batch requests
+	BatchRequestSizeLimitHTTP int `mapstructure:"batch_request_size_limit_grpc"`
```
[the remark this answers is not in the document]
dpaasman00: Also, can you update the exporter's Readme with these new parameters, along with their defaults? Maybe a quick blurb as well about how this works, specifically when logs will be dropped.

exporter/chronicleexporter/marshal.go
```diff
@@ -482,26 +519,60 @@ func buildForwarderString(cfg Config) string {
...
+	logs := request.GetInlineSource().Logs
+	if size <= m.cfg.BatchRequestSizeLimitHTTP && len(logs) <= m.cfg.BatchLogCountLimitHTTP {
+		return []*api.ImportLogsRequest{
+			request,
+		}
+	}
+
+	if len(logs) < 2 {
+		m.teleSettings.Logger.Error("Single entry exceeds max request size. Dropping entry", zap.Int("size", size))
+		return []*api.ImportLogsRequest{}
```
dpaasman00, line 546: I'm not sure I'm a fan of this, but I'm assuming this is the current behavior of secops where it will just drop an entry if it's too big?

dpaasman00 (changes requested):
Documentation requests and questions but that's it. Looking good.

exporter/chronicleexporter/marshal.go
mrsillydog: Yeah if one entry were too big for SecOps, it would just reject the request. Not sure what the alternative to this approach would be - this is what we did for Stanza as well sending to Google Cloud https://github.com/observIQ/stanza/blob/main/operator/builtin/output/googlecloud/request.go#L42

exporter/chronicleexporter/config.go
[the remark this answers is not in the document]
mrsillydog: Ticket has been created in Linear, BPOP-1402.

exporter/chronicleexporter/marshal.go
dpaasman00: Cool, if logs are going to be dropped regardless by secops then this is fine with me.

dpaasman00 (approved):
LGTM

PR #2034, opened 2024-12-09
fix: Delete empty values iterates through nested arrays

justinianvoss22:
### Proposed Change
- Added implementation for iterating through slices in telemetry
- Cleans value depending on value in array, whether empty, map, slice,  or str.
- Add unit tests to handle

###
Testing
- Unit testing
- Test agent locally in bp to see proper processor behavior

### Logs sent in
image
### Result seen in recent telemetry
image

### DIFF
--- a/processor/removeemptyvaluesprocessor/processor.go
+++ b/processor/removeemptyvaluesprocessor/processor.go
@@ -165,13 +165,11 @@ func (evp *emptyValueProcessor) processMetrics(_ context.Context, md pmetric.Met
 	return md, nil
 }
 
-// cleanMap removes empty values from the map, as defined by the config.
 func cleanMap(m pcommon.Map, c Config, excludeKeys map[string]struct{}) {
 	m.RemoveIf(func(s string, v pcommon.Value) bool {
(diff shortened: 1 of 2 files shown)


dpaasman00 (changes requested):
Couple nits otherwise this is working well for me...

These attributes:
image

Get converted to this after the processor:
Screenshot 2024-12-10 at 1 31 53 PM

justinianvoss22:
> Couple nits otherwise this is working well for me...
> 
> These attributes: image
> 
> Get converted to this after the processor: Screenshot 2024-12-10 at 1 31 53 PM

Pretty sure we want to keep this behavior. As an array, not sure how it would be possible to make the array layout array[0], array[1], array[3]. Because it is an array, I think moving up that value to the position of the deleted one makes sense.

dpaasman00:
Yea the behavior is correct, I was just posting that there showing it working.

dpaasman00 (changes requested):
Doesn't look like my comments were addressed, I think you  just forgot to push a commit.

dpaasman00 (approved):

PR #2043, opened 2024-12-11
chore: Rename to `bindplane-otel-collector`

dpaasman00:
### Proposed Change

Renaming the BindPlane Agent to BindPlane OTel Collector for v2. Need to rename repository to `bindplane-otel-collector` and update current versions to use this new name.

This renames module paths and links. After this is merged, the repository name will be renamed to `bindplane-otel-collector` and the `release/v1.67.0` branch rebased.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (approved):
We might want to add a set of container image tags for `bindplane-otel-collector`. I think this should be a follow up PR, to avoid dirtying up this PR.

antonblock (approved):

PR #2048, opened 2024-12-12
chore: Localize chronicle exporter's metrics concerns

djaglowski:
Follows #2047, another step towards #2044

This PR moves everything specific to the `hostMetricsReporter` into the same place and allows the struct to be managed with start/shutdown methods that correspond roughly to a typical exporter. (This will also help if we ever want to turn this into a proper `exporter.Metrics`.) Additionally, the `hostMetricsReporter` is only instantiated and used if `CollectAgentMetrics`, whereas before it was always instantiated and used to track metrics, even if they were not sent anywhere.

### DIFF
--- a/exporter/chronicleexporter/hostmetrics.go
+++ b/exporter/chronicleexporter/hostmetrics.go
@@ -15,6 +15,7 @@
 package chronicleexporter
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"sync"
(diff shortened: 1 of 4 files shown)


exporter/chronicleexporter/hostmetrics.go
```diff
@@ -38,19 +47,63 @@ type hostMetricsReporter struct {
...
+func (hmr *hostMetricsReporter) start() error {
+	ticker := time.NewTicker(5 * time.Minute)
+	defer ticker.Stop()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	hmr.cancel = cancel
+	hmr.wg.Add(1)
+	defer hmr.wg.Done()
+
+	for {
```
djaglowski, line 93: This needs to be in a goroutine

mrsillydog (approved):

PR #2049, opened 2024-12-12
chore: Add new tests for chronicle exporter with http and grpc servers

djaglowski:
Follows #2048, another step towards #2044

This PR mostly just adds tests, but a few minor changes are made to allow specific aspects of the code to be swapped out in those tests.

Regardless of protocol, a fake `tokenSource` function can be replaced, allowing us to bypass the loading of google credentials. Likewise, the endpoint is changed to point to the mock server. When using grpc, the dial opts are also changed to indicate no authentication. However, even with these changes the clients are real and actually exercised by the tests.

When using http, the exporter correctly returns a permanent error on at least some error codes. More should be done here, but it is a start and can be validated by tests.

Finally, in writing these tests, I found a few error messages which should be simplified.

Overall, test coverage for the exporter is improved from about to 61% to 72%.

### DIFF
--- a/exporter/chronicleexporter/exporter.go
+++ b/exporter/chronicleexporter/exporter.go
@@ -18,11 +18,9 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
-	"os"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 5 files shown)


mrsillydog (commented):
Looks like some tests are failing, otherwise LGTM

djaglowski:
Looks like it was an unrelated flaky test. Passed on a second run though.

mrsillydog (approved):

PR #2081, opened 2024-12-30
Feat(anomaly-connector) Log Volume Anomaly Connector

BominRahmani:
### Proposed Change

The following PR introduces the Log volume Anomaly connector.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


connector/loganomalyconnector/go.mod
```diff
@@ -0,0 +1,33 @@
+module github.com/observiq/bindplane-agent/connector/loganomalyconnector
+
+go 1.22.6
```
antonblock, line 3:
```suggestion
go 1.22.7
```

connector/loganomalyconnector/go.mod
```diff
@@ -0,0 +1,33 @@
+module github.com/observiq/bindplane-agent/connector/loganomalyconnector
```
antonblock, line 1:
```suggestion
module github.com/observiq/bindplane-otel-collector/connector/loganomalyconnector
```

antonblock (changes requested):
Two quick mod changes, haven't tested yet. Please add some tests for the config and Detector

antonblock (dismissed):

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	d.currentBucket.count += logCount
+
+	now := time.Now()
+	if now.Sub(d.lastSampleTime) >= d.config.SampleInterval {
+		d.takeSample(now)
+	}
+
+	return d.nextConsumer.ConsumeLogs(ctx, ld)
```
djaglowski, line 120:
```suggestion
	return nil
```

Connectors should not acts as passthroughs _and_ data generators.

Logs coming in and anomalies detected are two separate data streams. Users should have the ability to manage each data stream separately. This would be mixing data streams. By _not_ passing through, users can decide for themselves if they want to mix the streams or not.

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// countLogs counts the number of log records in the input
+func (d *Detector) countLogs(ld plog.Logs) int64 {
```
djaglowski, line 124: This function isn't needed. Just use `ld.LogRecordCount()`

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,145 @@
...
+	if d.currentBucket.start.IsZero() {
+		d.currentBucket.start = time.Now()
+	}
+
+	d.currentBucket.count += logCount
+
+	now := time.Now()
+	if now.Sub(d.lastSampleTime) >= d.config.SampleInterval {
+		d.takeSample(now)
+	}
```
djaglowski, line 118: Using samples and rates seems like a unnecessary precision loss. Why not just add accumulate counts? You can easily convert counts to rates immediately before the calculate. (If that's even necessary, as the formulae appear to apply just as well to a series of counts as a series of rates)

connector/loganomalyconnector/anomaly.go
```diff
@@ -0,0 +1,285 @@
...
+	cutoffTime := time.Now().Add(-d.config.MaxWindowAge)
+	idx := sort.Search(len(d.rateHistory), func(i int) bool {
+		return d.rateHistory[i].timestamp.After(cutoffTime)
+	})
+	if idx > 0 {
+		d.rateHistory = d.rateHistory[idx:]
+	}
+
+	// in the case we have more logs than specified for our buffer
+	if len(d.rateHistory) > d.config.EmergencyMaxSize {
```
djaglowski, line 100: If we're maintaining a maximum window age and buckets of fixed durations, how would this condition come about?

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,145 @@
...
+// The logs are then forwarded to the next consumer in the pipeline.
+func (d *Detector) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
+	d.stateLock.Lock()
+	defer d.stateLock.Unlock()
+
+	logCount := d.countLogs(ld)
+
+	if d.currentBucket.start.IsZero() {
+		d.currentBucket.start = time.Now()
+	}
```
djaglowski, line 111: This should be handled in the `Start` method

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,145 @@
...
+
+	d.currentBucket.count += logCount
+
+	now := time.Now()
+	if now.Sub(d.lastSampleTime) >= d.config.SampleInterval {
```
djaglowski, line 116: Why do we have to do a time check before calling `takeSample`, but also as the first thing we do within `takeSample`? What is the relationship between these?

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,145 @@
...
+	return count
+}
+
+// checkAndUpdateMetrics runs periodically to check for anomalies even when no logs are received
+func (d *Detector) checkAndUpdateAnomalies() {
+	d.stateLock.Lock()
+	defer d.stateLock.Unlock()
+
+	now := time.Now()
+	if now.Sub(d.lastSampleTime) >= d.config.SampleInterval {
```
djaglowski, line 142: Same here - why do we have to check the time before and after calling `takeSample`?

connector/loganomalyconnector/factory.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/connector"
+	"go.opentelemetry.io/collector/consumer"
+)
+
+// NewFactory returns a new factory for the Metrics Generation processor.
+func NewFactory() connector.Factory {
+	return connector.NewFactory(
+		component.MustNewType("loganomaly"),
```
djaglowski, line 30: Two observations about the name:
1. It's only detecting throughput anomalies and not looking at any content within the logs.
2. The same throughput analysis can apply just as well to other signals.

So maybe it shoudl be called `throughputanomaly`?

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}()
+
+	return nil
+}
+
+// Shutdown stops the detector's operations by canceling its context.
+// It cleans up any resources and stops the background sampling process.
+func (d *Detector) Shutdown(_ context.Context) error {
+	d.cancel()
```
djaglowski, line 91: and wait?

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,145 @@
...
+	return nil
+}
+
+// Capabilities returns the consumer capabilities of the detector.
+// It indicates that this detector does not mutate the data it processes.
+func (d *Detector) Capabilities() consumer.Capabilities {
+	return consumer.Capabilities{MutatesData: false}
+}
```
djaglowski, line 99: nit: this can be omitted since it just returns the default

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,145 @@
...
+
+// ConsumeLogs processes incoming log data, counting the logs and maintaining time-based sampling buckets.
+// The logs are then forwarded to the next consumer in the pipeline.
+func (d *Detector) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
+	d.stateLock.Lock()
+	defer d.stateLock.Unlock()
```
djaglowski, line 105: Is it really necessary to mux every call? This seems like a lot of overhead. Maybe push into a channel and have a single worker consuming?

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,145 @@
...
+type Detector struct {
+	ctx    context.Context
+	cancel context.CancelFunc
+	logger *zap.Logger
+
+	stateLock sync.Mutex
+	config    *Config
+
+	// Rolling window of rate samples
+	rateHistory []Sample
```
djaglowski, line 41: I think there is a notable mathematical inaccuracy in the way this is modeled. The windows are inconsistently sized based when logs happen to arrive, so a rate representing one window weighs more heavily than a rate representing a shorter window. The effect is exaggerated when you consider that windows are more likely to be longer when there is minimal activity.

This should be a lot simpler to manage anyways. Psuedocode:

```go

counts []int
lastWindowsEndTime time.Time

func (d *Detector) ConsumeLogs(ctx context.Context, ld plog.Logs) error {

      // may run more than once when no log has arrived recently
      for now.Sub(lastWindowEndTime).Minutes() > 1 {

          // dequeue the oldest count
          counts = append(counts[1:], 0)
          lastWindowsEndTime = lastWindowsEndTime.Add(time.Minute)
     }

    counts[len(counts)-1] += ld.LogRecordCount()
}

```

connector/loganomalyconnector/connector.go
djaglowski: A previous comment of mine explaining this in more detail: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/18381#issuecomment-1424910925

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,147 @@
...
+func (d *Detector) processLogBatch(batch logBatch) {
+	now := time.Now()
+
+	// Update windows that have elapsed since last update
+	for now.Sub(d.lastWindowEndTime).Minutes() >= 1 {
+		// Shift windows and add empty window
+		copy(d.counts, d.counts[1:])
+		d.counts[len(d.counts)-1] = 0
+		d.lastWindowEndTime = d.lastWindowEndTime.Add(time.Minute)
+	}
```
djaglowski, line 101: I think you need to separate this into 1) drop windows earlier than now-X, and 2) appending windows until the now. The reason is that this assumes we always want to drop a window, regardless of how many are currently present.

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Add counts to current window
+	var logCount int
+	if batch.logs != (plog.Logs{}) { // Check if logs is not empty
+		logCount = batch.logs.LogRecordCount()
+	}
```
djaglowski, line 107: When would this happen?

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,147 @@
...
+
+	counts            []int64
+	lastWindowEndTime time.Time
+
+	logChan      chan logBatch
+	nextConsumer consumer.Logs
+}
+
+func newDetector(config *Config, logger *zap.Logger, nextConsumer consumer.Logs) *Detector {
+	ctx, cancel := context.WithCancel(context.Background())
```
djaglowski, line 54: This should be done in `Start`

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+}
+
+// Start begins the anomaly detection process, sampling at intervals specified in the config.
+// It launches a background goroutine that periodically checks for and updates anomalies.
+func (d *Detector) Start(_ context.Context, _ component.Host) error {
+
+	ticker := time.NewTicker(d.config.SampleInterval)
+
+	go func() {
```
djaglowski, line 76: We should use a wait group and wait for it in Shutdown

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Add counts to current window
+	var logCount int
+	if batch.logs != (plog.Logs{}) { // Check if logs is not empty
+		logCount = batch.logs.LogRecordCount()
+	}
+	d.counts[len(d.counts)-1] += int64(logCount)
+
+	// Check for anomalies using the fixed window counts
+	if anomaly := d.checkForAnomaly(); anomaly != nil {
```
djaglowski, line 111: Should we really be checking for anomalies here? We're already doing it on a timer. Why make it event driven as well?

connector/loganomalyconnector/connector.go
```diff
@@ -0,0 +1,147 @@
...
+	return nil
+}
+
+func (d *Detector) processLogBatch(batch logBatch) {
+	now := time.Now()
+
+	// Update windows that have elapsed since last update
+	for now.Sub(d.lastWindowEndTime).Minutes() >= 1 {
+		// Shift windows and add empty window
+		copy(d.counts, d.counts[1:])
```
djaglowski, line 98: Since we're reading the counts in a separate goroutine, we need to be threadsafe. Maybe it best to use accumulate log counts onto an atomic int, and have the ticker trigger all updates to the count slice.

connector/loganomalyconnector/anomaly.go
```diff
@@ -0,0 +1,206 @@
...
+	attrs.PutDouble("anomaly.current_count", float64(anomaly.currentCount))
+	attrs.PutDouble("anomaly.z_score", anomaly.zScore)
+	attrs.PutDouble("anomaly.mad_score", anomaly.madScore)
+	attrs.PutDouble("anomaly.percentage_diff", anomaly.percentageDiff)
+	attrs.PutDouble("baseline.mean", anomaly.baselineStats.mean)
+	attrs.PutDouble("baseline.std_dev", anomaly.baselineStats.stdDev)
+	attrs.PutDouble("baseline.median", anomaly.baselineStats.median)
+	attrs.PutDouble("baseline.mad", anomaly.baselineStats.mad)
+
+	d.logger.Info("Log anomaly detected",
```
djaglowski, line 190: I'm not convinced it makes sense to emit the log two different ways.

connector/throughputanomalyconnector/config.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"go.opentelemetry.io/collector/component"
+)
+
+var _ component.Config = (*Config)(nil)
+
+// Config defines the configuration parameters for the log anomaly detector connector.
+type Config struct {
+	// How often to take measurements
+	SampleInterval time.Duration `mapstructure:"sample_interval"`
```
djaglowski, line 29: We aren't really sampling anymore right? But this is used to control how often we perform analysis on the counts, so maybe there's a better name.

Related, but probably for a future enhancement - should this be a "WindowSize" setting so that we're analyzing each window once?

connector/loganomalyconnector/connector.go
BominRahmani: It was a weird thing that was happening when i was running my unit test cases, although i don't think it'll ever happen in an actual environment. I can remove

djaglowski (approved):

PR #2021, opened 2024-12-02
feat: unroll processor

schmikei:
### Proposed Change
Still need to write some documentation on this but the idea is to create new log entries out of arrays. The start of this is working exclusively with the body of a log entry.

Sample config

```yaml
receivers:
  filelog:
    include: [ ./test.json ]
    start_at: beginning
  filelog/2:
    include: [ ./test2.json ]
    start_at: beginning
processors:
  transform:
    log_statements:
      - context: log
        statements:
          - set(body, Split(body, "\\n"))
          # - unroll(body)
  transform/split-by-comma:
    log_statements:
      - context: log
        statements:
          - set(body, Split(body, ","))
  transform/set-attr:
    log_statements:
      - context: log
        statements:
          - set(attributes["owner"], "Keith")
  unroll:
  unroll/body:
    unroll_key: ""

exporters:
  file:
    path: ./test/output.json
  debug:
    verbosity: detailed
  googlecloud/otel-agent-dev:
    log:
        compression: gzip
        default_log_name: Keiths-Macbook-Pro.local
        resource_filters:
            - regex: .*
    metric:
        compression: gzip
    project: otel-agent-dev


service:
  pipelines:
    # logs:
    #   receivers: [filelog]
    #   processors: [transform, unroll]
    #   exporters: [debug, file, googlecloud/otel-agent-dev]
    logs/2:
      receivers: [filelog/2]
      processors: [transform/split-by-comma, unroll/body, transform/set-attr]
      exporters: [debug, file, googlecloud/otel-agent-dev]
```

This for example can take in a log file like this

```
1,2,3,4,5,6,7,8
```

And it becomes multiple log events:

image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

PR #2086, opened 2025-01-06
chore: add unroll processor to v2 manifest

schmikei:
### Proposed Change
With #2021 merged, it seemed like it missed its way into the v2 manifests. This PR adds it back so that new processor types can use it with v2 installations!

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dpaasman00 (approved):
Can you throw it in the list of processors available in this manifest's readme as well?

PR #2098, opened 2025-01-07
fix: refactor azureeventhubrehydrationreceiver to stream blobs as to not lock up on larger environments (BPOP-831)

schmikei:
### Proposed Change

- Streaming results of the blobs rather than doing a lump sum for the requests of the blobs. This would cause the collector to stall for large rehydration efforts using azure eventhub.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


schmikei:
Opening up for review; but going to see if I can add some more tests/benchmarks as I go before merging anything

receiver/azureblobrehydrationreceiver/config.go
```diff
@@ -45,19 +48,19 @@ type Config struct {
 	// Default value of false
 	DeleteOnRead bool `mapstructure:"delete_on_read"`
 
-	// PollInterval is the interval at which the Azure API is scanned for blobs.
-	// Default value of 1m
-	PollInterval time.Duration `mapstructure:"poll_interval"`
-
-	// PollTimeout is the timeout for the Azure API to scan for blobs.
-	PollTimeout time.Duration `mapstructure:"poll_timeout"`
```
dpaasman00, line 53: I don't think these parameters should be removed quite yet. They should go through a deprecation process. However this receiver is unique in that it's a rehydration receiver and is only ran for short periods of time.

dpaasman00 (changes requested):
Some nits and a context thing

receiver/azureblobrehydrationreceiver/README.md
```diff
@@ -40,70 +41,4 @@ This is not a traditional receiver that continually produces data but rather reh
...
-This configuration enables the `delete_on_read` functionality which will delete a blob from Azure after it has been successfully rehydrated into OTLP data and sent onto the next component in the pipeline. 
-
-```yaml
-azureblobrehydration:
-    connection_string: "DefaultEndpointsProtocol=https;AccountName=storage_account_name;AccountKey=storage_account_key;EndpointSuffix=core.windows.net"
-    container: "my-container"
-    starting_time: 2023-10-01T13:00
-    ending_time: 2023-10-01T14:30
-    delete_on_read: true
-```
```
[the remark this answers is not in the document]
schmikei: Ope good catch, have no idea how that got deleted, apologies about that!

receiver/azureblobrehydrationreceiver/config.go
```diff
@@ -45,19 +48,38 @@ type Config struct {
...
 
 // Validate validates the config
 func (c *Config) Validate() error {
+	if c.PollInterval != 0 {
+		return errors.New("poll_interval is no longer supported and batch_size/page_size should be used instead")
+	}
+
+	if c.PollTimeout != 0 {
+		return errors.New("poll_timeout is no longer supported and batch_size/page_size should be used instead")
+	}
```
jsirianni, line 77: This is a breaking change. Can we log warnings and ignore these options instead of failing? 

```json
{
  "level":"fatal",
  "ts":"2025-01-14T10:50:55.721-0500",
  "caller":"collector/main.go:121",
  "msg":"RunService returned error",
  "error":"failed to start service: error during OpAmp connection: collector failed to start: invalid configuration: receivers::azureblobrehydration/source1_01JHJQHCEX5NHT9HN2RS9FCRHX: poll_interval is no longer supported and batch_size/page_size should be used instead"}
```

I get that error when using Bindplane's Azure Blob source.

```yaml
receivers:
    azureblobrehydration/source1_01JHJQHCEX5NHT9HN2RS9FCRHX:
        connection_string: redacted
        container: test
        delete_on_read: false
        ending_time: 2025-01-15T00:00
        poll_interval: 1m
        starting_time: 2025-01-01T00:00
```

This will allow us some time to get the Bindplane source updated, and provide users with upgrade compatibility. The source docs could include some disclaimers around agent version and support config options.

jsirianni (changes requested):
Working great. I did run into upgrade issues, left a comment around that.

jsirianni (changes requested):
Still working great, however, I am having some performance issues.

Without a batch processor, I seem to get five logs per request. I get a burst of requests and then a ~7 second delay before the next "burst"

```
{"level":"info","ts":"2025-01-14T16:01:09.776-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:10.919-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:10.999-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:11.080-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:11.161-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:11.241-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:11.322-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:11.402-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:11.484-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:11.563-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:11.644-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}

DELAY

{"level":"info","ts":"2025-01-14T16:01:19.787-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:20.934-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:21.015-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:21.097-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:21.179-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:21.265-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:21.351-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:21.432-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:21.512-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:21.592-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
{"level":"info","ts":"2025-01-14T16:01:21.675-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":5,"log records":5}
```

When I use a batch processor with a 2 second send interval, the logs seem to be more steady.

```
{"level":"info","ts":"2025-01-14T16:01:40.796-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":60,"log records":60}
{"level":"info","ts":"2025-01-14T16:01:42.798-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":105,"log records":105}
{"level":"info","ts":"2025-01-14T16:01:44.799-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":105,"log records":105}
{"level":"info","ts":"2025-01-14T16:01:46.800-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":120,"log records":120}
{"level":"info","ts":"2025-01-14T16:01:48.801-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":125,"log records":125}
{"level":"info","ts":"2025-01-14T16:01:50.803-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":100,"log records":100}
{"level":"info","ts":"2025-01-14T16:01:52.805-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":125,"log records":125}
{"level":"info","ts":"2025-01-14T16:01:54.806-0500","msg":"Logs","kind":"exporter","data_type":"logs","name":"debug/debug","resource logs":115,"log records":115}
```

receiver/azureblobrehydrationreceiver/config.go
dpaasman00: Yea that works for me!

receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -109,170 +120,185 @@ func newRehydrationReceiver(id component.ID, logger *zap.Logger, cfg *Config) (*
...
 		id:              id,
 		cfg:             cfg,
 		azureClient:     azureClient,
-		doneChan:        make(chan struct{}),
 		checkpointStore: rehydration.NewNopStorage(),
 		startingTime:    startingTime,
 		endingTime:      endingTime,
-		ctx:             ctx,
-		cancelFunc:      cancel,
+		blobChan:        make(chan *azureblob.BlobResults),
```
jsirianni, line 131: I think we should consider making this a buffered channel. I think it would be okay for the receiver to slow to a crawl if the channel is full and we are waiting on blobs to be processed and sent down the pipeline.

receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -109,170 +120,185 @@ func newRehydrationReceiver(id component.ID, logger *zap.Logger, cfg *Config) (*
...
+				emptyPolls = 0
+				continue
 			}
 
-			numBlobsRehydrated := r.rehydrateBlobs(checkpoint, marker)
-			emptyBlobCounter = checkBlobCount(numBlobsRehydrated, emptyBlobCounter)
+			emptyPolls++
+			if emptyPolls == emptyPollLimit {
+				r.logger.Warn("No blobs processed for 3 consecutive polls, assuming no more blobs to process")
+				return
```
jsirianni, line 222: I am still curious about this. It feels awkward to me to have the receiver stop processing logs despite continuing to have the process run.

I think it would be nice to progressively backoff. Starting with some short interval in the seconds, maxing out in the minutes. If someone leaves the agent running too long, and it is backing off to retrying every 5m or 10m, I would not expect their Azure bill to be impacted much.

jsirianni (changes requested):
General question: With the concurrency changes, we should make sure the shutdown logic allows the receiver to process all in flight batches before shutting down or persisting the marker to storage. 

It may already be doing this, just want to be sure.

receiver/azureblobrehydrationreceiver/config.go
```diff
@@ -25,6 +25,9 @@ import (
 
 // Config is the configuration for the azure blob rehydration receiver
 type Config struct {
+	// BatchSize is the number of blobs to process entering the pipeline in a single batch. (default 30)
+	BatchSize int `mapstructure:"batch_size"`
```
jsirianni, line 29: We might want to comment about concurrency here, and mention that each blob will be responsible for a connection to azure. Funny enough, I ran out of file handles (limited to 1024 by default) when configuring this to 10,000. It was working for the most part.

```suggestion
	// BatchSize is the number of blobs to process entering the pipeline in a single batch. (default 30)
	BatchSize int `mapstructure:"batch_size"`
```

jsirianni (changes requested):

schmikei:
> General question: With the concurrency changes, we should make sure the shutdown logic allows the receiver to process all in flight batches before shutting down or persisting the marker to storage.
> 
> It may already be doing this, just want to be sure.

It should be doing this because of the wait group but I'll write a test to make sure!

receiver/azureblobrehydrationreceiver/config.go
schmikei: It is mentioned in the readme, but yep I can update it here as well!

receiver/azureblobrehydrationreceiver/receiver.go
schmikei: I would agree generally that maybe a backoff approach could be taken, I was mostly just trying to keep the old functionality of the receiver stopping requests if nothing is being processed as a cost saving measure. I think Corbin was focusing on that for this use case rather than thinking a backoff approach should be taken...

@dpaasman00 do you have any thoughts on this as you've looked at s3 rehydration?

receiver/azureblobrehydrationreceiver/receiver.go
schmikei: Good idea 👍

receiver/azureblobrehydrationreceiver/receiver.go
jsirianni: Okay, Im okay with keeping it the same but would like to hear Ryan's take on it.

jsirianni: It could certainly be a future change.

receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -90,9 +101,13 @@ func newTracesReceiver(id component.ID, logger *zap.Logger, cfg *Config, nextCon
 	return r, nil
 }
 
+// factor of buffered channel size
+// number of blobs to process at a time is blobChanSize * batchSize
+const blobChanSize = 5
```
jsirianni, line 106: Cool, this seems like a good size, its performing well for me. In the future maybe we can expose it. I don't see a need to right now.

jsirianni (approved):
This is working well for me. Id like @dpaasman00 to approve as well, he knows more of the collector inner workings.

schmikei:
@dpaasman00 was noticing some shutdown hanging but with my most recent commit I think this is ready for another look if you have a moment!

dpaasman00 (changes requested):
Some comments, nothing major

receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -109,170 +124,202 @@ func newRehydrationReceiver(id component.ID, logger *zap.Logger, cfg *Config) (*
...
+			}()
 		}
 	}
 
-	return
+	r.wg.Wait()
+
+	if err := r.makeCheckpoint(ctx); err != nil {
+		r.logger.Error("Error while saving checkpoint", zap.Error(err))
+	}
```
dpaasman00, line 309: Thoughts on checkpointing after every blob that gets processed? I think it would help simplify some of the concurrency logic and general logic, I also don't think it'd be that costly either... thoughts?

receiver/azureblobrehydrationreceiver/receiver.go
dpaasman00: It looks like this functionality was removed, but I think that's okay since we're now using a better pagination process. We shouldn't need to worry about getting empty polls anymore.

receiver/azureblobrehydrationreceiver/receiver.go
schmikei: Can do, albeit I don't think that the rehydration timestamps are guaranteed to be in order at least based off this screenshot. I'm skeptical that checkpointing was working as designed before-hand to be honest.

image

receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -109,170 +124,202 @@ func newRehydrationReceiver(id component.ID, logger *zap.Logger, cfg *Config) (*
...
 
-	var marker *string
+	var errs error
+	if err := r.makeCheckpoint(shutdownCtx); err != nil {
+		r.logger.Error("Error while saving checkpoint", zap.Error(err))
+		err = errors.Join(err, err)
+	}
+
+	return errors.Join(errs, r.checkpointStore.Close(shutdownCtx))
```
dpaasman00, line 203: Generally try to wrap errors, and since we're combining the errors from these 2 calls I don't think we need to log the error from the first one since it'll get reported once shutdown is done. Thoughts?
```suggestion
	var errs error
	if err := r.makeCheckpoint(shutdownCtx); err != nil {
		errs = errors.Join(errs, fmt.Errorf("make checkpoint: %w", err))
	}
	if err := r.checkpointStore.Close(shutdownCtx); err != nil {
	         errs = errors.Join(errs, fmt.Errorf("close checkpoint store: %w", err))
	}
        return errs
```

dpaasman00: If you decide to checkpoint after every blob though, I don't think we need to worry about checkpointing in shutdown, so the first error block could get removed, but that's if the checkpointing logic changes

receiver/azureblobrehydrationreceiver/receiver.go
dpaasman00, line 1: Errant const var `emptyPollLimit` still present in this file from previous logic

receiver/azureblobrehydrationreceiver/README.md
```diff
@@ -13,25 +13,32 @@ This is not a traditional receiver that continually produces data but rather reh
...
 ## How it works
-1. The receiver polls blob storage for all blobs in the specified container.
+1. The receiver polls blob storage for pages of blobs in the specified container.
 2. The receiver will parse each blob's path to determine if it matches a path created by the [Azure Blob Exporter](../../exporter/azureblobexporter/README.md#blob-path).
 3. If the blob path is from the exporter, the receiver will parse the timestamp represented by the path.
 4. If the timestamp is within the configured range the receiver will download the blob and parse its contents into OTLP data.
 
     a. The receiver will process both uncompressed JSON blobs and blobs compressed with gzip.
 
+> Note: There is no current way of specifying a time range to rehydrate so any blobs outside fo the time range still  need to be retrieved from the API in order to filter via the `starting_time` and `ending_time` configuration.
```
dpaasman00, line 23:
```suggestion
> Note: There is no current way of specifying a time range to rehydrate so any blobs outside of the time range still need to be retrieved from the API in order to filter via the `starting_time` and `ending_time` configuration.
```

receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -109,170 +124,204 @@ func newRehydrationReceiver(id component.ID, logger *zap.Logger, cfg *Config) (*
...
+	select {
+	case <-done:
+	case <-shutdownCtx.Done():
+		return fmt.Errorf("shutdown timeout: %w", shutdownCtx.Err())
+	}
 
-	var marker *string
+	var errs error
+	if err := r.makeCheckpoint(shutdownCtx); err != nil {
+		r.logger.Error("Error while saving checkpoint", zap.Error(err))
```
dpaasman00, line 199: I think this log can get removed since we're in shutdown and the error will get reported. Pretty nit tho so up to you

receiver/azureblobrehydrationreceiver/receiver.go
```diff
@@ -109,170 +124,204 @@ func newRehydrationReceiver(id component.ID, logger *zap.Logger, cfg *Config) (*
...
-	if err != nil {
-		r.logger.Error("Failed to list blobs", zap.Error(err))
-		return
-	}
-
-	marker = nextMarker
-
+func (r *rehydrationReceiver) rehydrateBlobs(ctx context.Context, blobs []*azureblob.BlobInfo) (numProcessedBlobs int) {
 	// Go through each blob and parse it's path to determine if we should consume it or not
+	r.logger.Debug("received a batch of blobs, parsing through them to determine if they should be rehydrated", zap.Int("num_blobs", len(blobs)))
```
dpaasman00, line 252:
```suggestion
	r.logger.Debug("Received a batch of blobs, parsing through them to determine if they should be rehydrated", zap.Int("num_blobs", len(blobs)))
```

dpaasman00 (approved):
Left a few more comments about nits, but this looks good. Merge when you're happy with it

PR #2099, opened 2025-01-08
feat: Make SSAPI receiver storage optional

Caleb-Hurshman:
### Proposed Change

Removes the storage extension as a requirement. Mostly for debugging/testing purposes, should never be disabled by the user for actual use of the receiver.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/receiver/splunksearchapireceiver/receiver.go
+++ b/receiver/splunksearchapireceiver/receiver.go
@@ -247,7 +247,7 @@ func (ssapir *splunksearchapireceiver) pollSearchCompletion(ctx context.Context,
 }
 
 func (ssapir *splunksearchapireceiver) createSplunkSearch(search Search) (string, error) {
-	timeFormat := "%Y-%m-%dT%H:%M:%S"
+	timeFormat := "%Y-%m-%dT%H:%M"
 	searchQuery := fmt.Sprintf("%s starttime=\"%s\" endtime=\"%s\" timeformat=\"%s\"", search.Query, search.EarliestTime, search.LatestTime, timeFormat)
 	ssapir.logger.Info("creating search", zap.String("query", searchQuery))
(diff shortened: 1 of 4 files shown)


receiver/splunksearchapireceiver/receiver.go
```diff
@@ -247,7 +247,7 @@ func (ssapir *splunksearchapireceiver) pollSearchCompletion(ctx context.Context,
 }
 
 func (ssapir *splunksearchapireceiver) createSplunkSearch(search Search) (string, error) {
-	timeFormat := "%Y-%m-%dT%H:%M:%S"
+	timeFormat := "%Y-%m-%dT%H:%M"
```
dpaasman00, line 250: What's going on here with this change? Also can you update the readme?

dpaasman00 (changes requested):

receiver/splunksearchapireceiver/receiver.go
Caleb-Hurshman: The timestamps from bindplane are using `rehydration.TimeFormat`, which isn't precise down to the second. If Splunk thinks there should be seconds, it will fail to parse the timestamps from BP that don't have them. This change should've been made with the last PR for this receiver, but it slipped through the cracks.

dpaasman00: I gotcha, if you're able to update the readme that the storage extension is not required that'd be great

Caleb-Hurshman: Sorry, got distracted with a different PR. It's not listed as required anymore, but it's still a prereq for any user. Basically, this change is for the dev side of things, but no user should run the receiver without a storage extn.

dpaasman00 (approved):

PR #2100, opened 2025-01-10
chore: Add SSAPI to v2 manifest

dpaasman00:
### Proposed Change

- Updates builder and supervisor to v0.116.0 to match other otel versions
- Adds SSAPI to V2 manifest
- Removes errant opamp file that was added during rebase (opamp folder is deleted, but recent PR in v1 added a file to the folder, so it got added back in during rebase)

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


schmikei:
Seems like we need to re run `make add-license`? Otherwise I think this looks good

schmikei (approved):

PR #2103, opened 2025-01-17
fix(awss3rehydrationreceiver): Stream objects from S3 to improve performance

dpaasman00:
### Proposed Change

Refactor the S3 receiver to stream objects from S3 instead of polling. The previous poll logic was missing correct pagination and didn't handle large buckets well. This change uses an S3 paginator to stream objects. There should also be a performance improvement after making it more concurrent as well.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/awss3rehydrationreceiver/reciever.go
```diff
@@ -288,31 +339,35 @@ func (r *rehydrationReceiver) processObject(object *aws.ObjectInfo) error {
...
+	if r.lastObjectName == "" || r.lastObjectTime == nil {
+		return nil
 	}
+
+	// update && store checkpoint
+	r.checkpoint.UpdateCheckpoint(*r.lastObjectTime, r.lastObjectName)
+	if err := r.checkpointStore.SaveCheckpoint(ctx, r.checkpointKey, r.checkpoint); err != nil {
+		return fmt.Errorf("save checkpoint: %w", err)
+	}
+	return nil
```
schmikei, line 372:
```suggestion
	if r.lastObjectName == "" || r.lastObjectTime == nil {
		return nil
	}
	r.checkpointMutex.Lock()
	defer r.checkpointMutex.Unlock()

	// update && store checkpoint
	r.checkpoint.UpdateCheckpoint(*r.lastObjectTime, r.lastObjectName)
	if err := r.checkpointStore.SaveCheckpoint(ctx, r.checkpointKey, r.checkpoint); err != nil {
		return fmt.Errorf("save checkpoint: %w", err)
	}
	return nil
```

Nit: We probably don't want to lock the mutex if we are doing an early return in this case.

receiver/awss3rehydrationreceiver/reciever.go
```diff
@@ -109,168 +122,206 @@ func newRehydrationReceiver(id component.ID, logger *zap.Logger, cfg *Config) (*
...
-		doneChan:        make(chan struct{}),
 		checkpointStore: rehydration.NewNopStorage(),
 		startingTime:    startingTime,
 		endingTime:      endingTime,
-		ctx:             ctx,
-		cancelFunc:      cancel,
+		doneChan:        make(chan struct{}),
+		wg:              &sync.WaitGroup{},
+		checkpointMutex: &sync.Mutex{},
+		objectChan:      make(chan *aws.ObjectResults),
```
schmikei, line 136: Sort of up to you since I think its somewhat minor but do we want to make this a buffered chan size? https://github.com/observIQ/bindplane-otel-collector/pull/2098#discussion_r1917316671

receiver/awss3rehydrationreceiver/reciever.go
dpaasman00: Yea that's fair, there's no consequential race condition if we end up returning early

receiver/awss3rehydrationreceiver/reciever.go
dpaasman00: Talked offline

schmikei (approved):

PR #2105, opened 2025-01-17
fix: dont buffer on azurerehydration receiver

schmikei:
I believe we want to block on the channel read as the doneChan gets closed when the stream is over. 

Thanks to @dpaasman00 for catching this.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/azureblobrehydrationreceiver/receiver.go
+++ b/receiver/azureblobrehydrationreceiver/receiver.go
@@ -101,10 +101,6 @@ func newTracesReceiver(id component.ID, logger *zap.Logger, cfg *Config, nextCon
 	return r, nil
 }
 
-// factor of buffered channel size
-// number of blobs to process at a time is blobChanSize * batchSize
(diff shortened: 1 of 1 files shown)


receiver/azureblobrehydrationreceiver/receiver.go
dpaasman00, line 1: Should it still check if the output is ok when it pulls from the channel? Doesn't matter to me just wondering

dpaasman00 (approved):
Looks good, just 1 question for you, otherwise merge when happy

receiver/azureblobrehydrationreceiver/receiver.go
schmikei: I think its probably safer in this case if ever the blob_client doesn't close the channel properly in the future as to not panic on closed read. Just think in principle it's a "better safe than sorry" kind of check imo

PR #2128, opened 2025-01-29
fix: Remove chan close in rehydration receiver shutdowns

dpaasman00:
### Proposed Change

Closing the done channel in shutdown was causing the function to block and not shutdown properly. I believe this is because we are calling the ctx cancel func which is going to cause the go routine to return before it can read from the done chan.

Also acquire the lock before closing the checkpoint store.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/awss3rehydrationreceiver/reciever.go
+++ b/receiver/awss3rehydrationreceiver/reciever.go
@@ -158,7 +158,6 @@ func (r *rehydrationReceiver) Shutdown(ctx context.Context) error {
 	if r.cancelFunc != nil {
 		r.cancelFunc()
 	}
-	close(r.doneChan)
 
 	// wait for any in-progress object rehydrations to finish
 	done := make(chan struct{})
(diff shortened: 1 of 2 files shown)


schmikei:
For some reason I'm getting this error even with your change:

```json
{"level":"error","ts":"2025-01-29T09:40:02.778-0500","caller":"azureblobrehydrationreceiver/receiver.go:305","msg":"Error while saving checkpoint","kind":"receiver","name":"azureblobrehydration","data_type":"logs","error":"database not open","stacktrace":"github.com/observiq/bindplane-otel-collector/receiver/azureblobrehydrationreceiver.(*rehydrationReceiver).rehydrateBlobs\n\t/Users/keithschmitt/workspace/bindplane-agent/receiver/azureblobrehydrationreceiver/receiver.go:305\ngithub.com/observiq/bindplane-otel-collector/receiver/azureblobrehydrationreceiver.(*rehydrationReceiver).streamRehydrateBlobs\n\t/Users/keithschmitt/workspace/bindplane-agent/receiver/azureblobrehydrationreceiver/receiver.go:239"}
```

🤔

schmikei (approved):

PR #2131, opened 2025-01-30
feat: Use timeout for Dial in secops forwarder

dpaasman00:
### Proposed Change

Call the `Dial` functions with a timeout to exit early in the case of a bad IP address. 

Uses the timeout already present in the configuration from the squashed `confignet.AddrConfig`. This way when BP starts adding this param to the exporter config it is backwards compatible

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/chronicleforwarderexporter/exporter.go
+++ b/exporter/chronicleforwarderexporter/exporter.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/exporter"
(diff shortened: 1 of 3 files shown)


exporter/chronicleforwarderexporter/exporter.go
```diff
@@ -52,11 +53,13 @@ type forwarderClient struct {
...
 func (fc *forwarderClient) Dial(network string, address string) (net.Conn, error) {
-	return net.Dial(network, address)
+	return net.DialTimeout(network, address, 10*time.Second)
 }
 
 func (fc *forwarderClient) DialWithTLS(network string, addr string, config *tls.Config) (*tls.Conn, error) {
-	return tls.Dial(network, addr, config)
+	d := new(net.Dialer)
+	d.Timeout = 10 * time.Second
+	return tls.DialWithDialer(d, network, addr, config)
```
dpaasman00, line 62: This is meant to be identical to what `tls.Dial` normally does except we're setting a timeout instead now: https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/crypto/tls/tls.go;l=173

exporter/chronicleforwarderexporter/exporter.go
```diff
@@ -49,14 +50,17 @@ type chronicleForwarderClient interface {
...
 func (fc *forwarderClient) Dial(network string, address string) (net.Conn, error) {
-	return net.Dial(network, address)
+	return net.DialTimeout(network, address, fc.timeout)
 }
 
 func (fc *forwarderClient) DialWithTLS(network string, addr string, config *tls.Config) (*tls.Conn, error) {
-	return tls.Dial(network, addr, config)
+	d := new(net.Dialer)
+	d.Timeout = fc.timeout
+	return tls.DialWithDialer(d, network, addr, config)
```
dpaasman00, line 63: This is meant to be identical to what `tls.Dial` normally does except we're setting a timeout instead now: https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/crypto/tls/tls.go;l=173

mrsillydog (approved):

PR #2133, opened 2025-01-31
feat: Use ReportHeartbeat OpAmp capability

dpaasman00:
### Proposed Change

Updates opamp-go dependency to v0.19.0 so we can set ReportsHeartbeat Capability and set default with empty interval to disable by default. The idea is if the agent connects to an older BP version, the heartbeats will cause BP to hit the store constantly, so disable heartbeats by default. Newer BP versions won't hit the store anymore and will set this interval on the agent during the first connection.

Also runs `make generate` with updated Mockery

https://linear.app/bindplane/issue/BPOP-164/implement-heartbeat-in-bindplane-agent

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Caleb-Hurshman (approved):
LGTM ✅

PR #2134, opened 2025-01-31
chore: Add script for generating available components resource for v1.x collectors

mrsillydog:
### Proposed Change

This script generates a YAML available components resource file for Bindplane, so that its database can be seeded on startup with the available components for non-supervisor collectors.

This script should be run every release of the BDOT collector - an addition to the release workflow will be coming soon.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


mrsillydog:
After talking with Dakota, this needs a little more refinement - agent version and type need to be on the spec.

mrsillydog:
This needs updating to work based on the release tags not the branches.

mrsillydog:
Should be good to go - this isn't a perfect script in regards to past releases of the collector, but should be good going forward. I've manually modified the relevant resources in Bindplane that this script doesn't work with.

dpaasman00 (approved):

PR #2167, opened 2025-02-14
feat: Allow for custom endpoints in chronicleexporter/http_exporter

shazlehu:
Removes the requirement for Config.Location to be set so that users can use a custom endpoint. If Location is set, we construct the endpoint for backwards compatibility.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/chronicleexporter/config.go
+++ b/exporter/chronicleexporter/config.go
@@ -137,8 +137,8 @@ func (cfg *Config) Validate() error {
 	}
 
 	if cfg.Protocol == protocolHTTPS {
-		if cfg.Location == "" {
-			return errors.New("location is required when protocol is https")
+		if cfg.Endpoint == "" {
+			return errors.New("endpoint is required when protocol is https")
 		}
 		if cfg.Project == "" {
 			return errors.New("project is required when protocol is https")
--- a/exporter/chronicleexporter/http_exporter.go
+++ b/exporter/chronicleexporter/http_exporter.go
@@ -156,6 +156,9 @@ func (exp *httpExporter) uploadToChronicleHTTP(ctx context.Context, logs *api.Im
 // URL for the request: https://{region}-chronicle.googleapis.com/{version}/projects/{project}/location/{region}/instances/{customerID}
 // Override for testing
 var httpEndpoint = func(cfg *Config, logType string) string {
+	if cfg.Location == "" {
+		return fmt.Sprintf("https://%s/v1alpha/projects/%s/instances/%s/logTypes/%s/logs:import", cfg.Endpoint, cfg.Project, cfg.CustomerID, logType)
+	}
 	formatString := "https://%s-%s/v1alpha/projects/%s/locations/%s/instances/%s/logTypes/%s/logs:import"
 	return fmt.Sprintf(formatString, cfg.Location, cfg.Endpoint, cfg.Project, cfg.Location, cfg.CustomerID, logType)
 }
--- a/exporter/chronicleexporter/config_test.go
+++ b/exporter/chronicleexporter/config_test.go
@@ -109,7 +109,7 @@ func TestConfigValidate(t *testing.T) {
 			expectedErr: "invalid compression type",
 		},
 		{
-			desc: "Protocol is https and location is empty",
+			desc: "Protocol is https and endpoint is empty",
 			config: &Config{
 				CredsFilePath:             "/path/to/creds_file",
 				LogType:                   "log_type_example",
@@ -120,11 +120,12 @@ func TestConfigValidate(t *testing.T) {
 				BatchRequestSizeLimitHTTP: defaultBatchRequestSizeLimitHTTP,
 				BatchLogCountLimitHTTP:    defaultBatchLogCountLimitHTTP,
 			},
-			expectedErr: "location is required when protocol is https",
+			expectedErr: "endpoint is required when protocol is https",
 		},
 		{
 			desc: "Protocol is https and forwarder is empty",
 			config: &Config{
+				Endpoint:                  "myendpoint.com",
 				CredsFilePath:             "/path/to/creds_file",
 				LogType:                   "log_type_example",
 				Protocol:                  protocolHTTPS,
@@ -139,6 +140,7 @@ func TestConfigValidate(t *testing.T) {
 		{
 			desc: "Protocol is https and project is empty",
 			config: &Config{
+				Endpoint:                  "myendpoint.com",
 				CredsFilePath:             "/path/to/creds_file",
 				LogType:                   "log_type_example",
 				Protocol:                  protocolHTTPS,
@@ -153,6 +155,7 @@ func TestConfigValidate(t *testing.T) {
 		{
 			desc: "Protocol is https and http batch log count limit is 0",
 			config: &Config{
+				Endpoint:                  "myendpoint.com",
 				CredsFilePath:             "/path/to/creds_file",
 				LogType:                   "log_type_example",
 				Protocol:                  protocolHTTPS,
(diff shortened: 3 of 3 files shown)


exporter/chronicleexporter/http_exporter.go
```diff
@@ -156,6 +156,9 @@ func (exp *httpExporter) uploadToChronicleHTTP(ctx context.Context, logs *api.Im
 // URL for the request: https://{region}-chronicle.googleapis.com/{version}/projects/{project}/location/{region}/instances/{customerID}
 // Override for testing
 var httpEndpoint = func(cfg *Config, logType string) string {
+	if cfg.Location == "" {
+		return fmt.Sprintf("https://%s/v1alpha/projects/%s/instances/%s/logTypes/%s/logs:import", cfg.Endpoint, cfg.Project, cfg.CustomerID, logType)
```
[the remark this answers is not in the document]
colelaven: Can confirm that we definitely still need the location in the URL

exporter/chronicleexporter/config.go
```diff
@@ -137,8 +137,8 @@ func (cfg *Config) Validate() error {
 	}
 
 	if cfg.Protocol == protocolHTTPS {
-		if cfg.Location == "" {
-			return errors.New("location is required when protocol is https")
```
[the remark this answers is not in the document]
shazlehu: This is outdated

exporter/chronicleexporter/config.go
```diff
@@ -133,15 +143,20 @@ func (cfg *Config) Validate() error {
 	}
 
 	if cfg.Protocol == protocolHTTPS {
-		if cfg.Location == "" {
```
colelaven, line 136: I think we still want to require Location for HTTPS because it is used to construct the endpoint in `baseEndpoint` regardless of `ignore_location`

exporter/chronicleexporter/http_exporter.go
```diff
@@ -426,8 +429,17 @@ var httpStatsEndpoint = func(cfg *Config, collectorID string) string {
 }
 
 func baseEndpoint(cfg *Config) string {
-	formatString := "https://%s-%s/v1alpha/projects/%s/locations/%s/instances/%s"
-	return fmt.Sprintf(formatString, cfg.Location, cfg.Endpoint, cfg.Project, cfg.Location, cfg.CustomerID)
+	var endpoint string
+	if cfg.IgnoreLocation {
+		endpoint = cfg.Endpoint
+	} else {
+		endpoint = fmt.Sprintf("%s-%s", cfg.Location, cfg.Endpoint)
```
colelaven: Do you know of any use case where a user would want to use the location in the endpoint but still override just the 2nd part of the endpoint? I'm not aware of any myself, not sure if we need this option here. I might be missing some context.

-- Assuming there isn't a clear use case for that --

IMO I think we could do it like this:
- If endpoint is provided, use it for the whole base URL
  - `https://<endpoint>/<apiVersion>/...`
- If no endpoint is provided, construct it from the location & default endpoint (`chronicle.googleapis.com`)
  - `https://<location>-<defaultEndpoint>/<apiVersion>/...`

I'm thinking ahead to when we eventually switch to REP format endpoints, the way we combine region and endpoint will be different: `https://chronicle.<location>.rep.googleapis.com`

If we still allow overriding the endpoint AND not ignoring the location, the combination logic gets messier. Is the default endpoint now `chronicle.%s.rep.googleapis.com`? or just `.rep.googleapis.com`?

I think it'd be more future-proof to use an approach of users "overriding baseURL" instead of "ignoring location". That way users either override the entire baseURL (including location) or not configure it at all and we will construct it for them.

I know this is long and nitpicky, happy to chat about it if you want. I just want to avoid regretting a decision later once this exporter is upstreamed and tedious to modify.

exporter/chronicleexporter/config.go
shazlehu: Good catch

exporter/chronicleexporter/http_exporter.go
shazlehu: Agreed, we should make sure we've got the flexibility we want. Would this change break a collector upgrading from the previous version (until it received a new config based on version)? Adding the `IgnoreLocation` override ensures an upgrade would still be correctly configured. Maybe change the name to `UseEndpointOnly` or something similar?  Or we could add `OverrideEndpoint`, and ignore `Endpoint` when it's set.

colelaven: Ahh great callout... I think it would be an issue to implement it the way I described.

I like the idea of renaming the boolean to something unrelated to location (`useEndpointOnly`, `overrideEndpoint` or something along those lines). That seems to be the cleanest solution we can get.

Somewhat related, I received intel this morning that the REP format endpoints wont be fully ready until sometime in Q2 of this year, so that migration wont happen for a while.

colelaven: Will you have time to make this adjustment today?
If not I'm happy to help so we can make sure this gets in the next collector release on Mon/Tues.

exporter/chronicleexporter/README.md
```diff
@@ -37,6 +37,8 @@ The exporter can be configured using the following fields:
 | `collect_agent_metrics`         | bool              | `true`                                 | `false`  | Enables collecting metrics about the agent's process and log ingestion metrics              |
 | `batch_request_size_limit_grpc` | int               | `4000000`                              | `false`  | The maximum size, in bytes, allowed for a gRPC batch creation request.                      |
 | `batch_request_size_limit_http` | int               | `4000000`                              | `false`  | The maximum size, in bytes, allowed for a HTTP batch creation request.                      |
+| `api_version`                   | string            | `v1alpha`                              | `false`  | The API version to use. Valid values are `v1alpha` and `v1beta`.                             |
+| `ignore_location`               | bool              | `false`                                | `false`  | Whether or not to ignore the Location field when constructing the endpoint.                  |
```
colelaven: need to update this reference to `ignore_location`
can you make a note that these 2 params are only for HTTPS logic as well?

exporter/chronicleexporter/http_exporter.go
```diff
@@ -426,8 +426,17 @@ var httpStatsEndpoint = func(cfg *Config, collectorID string) string {
 }
 
 func baseEndpoint(cfg *Config) string {
-	formatString := "https://%s-%s/v1alpha/projects/%s/locations/%s/instances/%s"
-	return fmt.Sprintf(formatString, cfg.Location, cfg.Endpoint, cfg.Project, cfg.Location, cfg.CustomerID)
+	var endpoint string
```
colelaven: nit: can you call this variable `baseURL` to be a bit more accurate?

unrelated - the cfg.Endpoint field should really be called baseURL as well but its not worth the effort at this point. I'm hoping to get that changed for the upstreamed exporter now that we have a clean refresh.

exporter/chronicleexporter/config.go
```diff
@@ -107,6 +109,14 @@ type Config struct {
...
 	// LogErroredPayloads is a flag that determines whether or not to log errored payloads.
 	LogErroredPayloads bool `mapstructure:"log_errored_payloads"`
+
+	// OverrideEndpoint determines whether or not to ignore the Location field when constructing the endpoint.
+	// This is useful for when the endpoint is a custom endpoint and the Location field is not needed.
+	// We still need the Location field for the API call to Chronicle, but we don't want to use it in the endpoint.
+	OverrideEndpoint bool `mapstructure:"override_endpoint"`
+
+	// APIVersion is the version of the API to use. Default is "v1alpha".
+	APIVersion string `mapstructure:"api_version"`
```
colelaven, line 120: can you make a note in the comments for these 2 params that they are only affecting HTTPS logic

PR #2171, opened 2025-02-18
deps: Update OTEL to v0.120.0

mrsillydog:
### Proposed Change
Update OTEL collector and contrib dependencies to v0.120.0, pdata to v1.26.0

##### Checklist
- [X] Changes are tested
- [X] CI has passed


collector/mocks/mock_collector.go
dpaasman00, line 1: It looks like mockery was ran with a different version than what's specified in `internal/tools/go.mod`. For this PR, can you run `make install-tools` and then `make generate`. I know the dependabot PR will probably update mockery at that point to v2.52.2 so this is really just a nit

dpaasman00 (changes requested):

dpaasman00 (approved):

PR #2174, opened 2025-02-18
chore(docker): Remove no install recommends from Ubuntu build

jsirianni:
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


mrsillydog (approved):
Will handle Mergefreeze if the release process does not complete on re-run.

PR #2191, opened 2025-02-25
feat(support): Update support bundle script to include go profiling

Dylan-M:
### Proposed Change
Updated the support bundle to collect profiles in addition to the log files.

##### Checklist
- [X] Changes are tested
- [x] CI has passed

### DIFF
--- a/scripts/support/create-support-bundle.sh
+++ b/scripts/support/create-support-bundle.sh
@@ -304,13 +304,13 @@ collect_profiles() {
     info "where 13133 is the port you configured in your profile extension (13133 is the default)"
     read -p "Endpoint: " ENDPOINT
     printf "\n"
-    curl -ksS http://localhost:13133/debug/pprof/goroutine --output goroutines.pprof
-    curl -ksS http://localhost:13133/debug/pprof/heap --output heap.pprof
-    curl -ksS http://localhost:13133/debug/pprof/threadcreate --output threadcreate.pprof
-    curl -ksS http://localhost:13133/debug/pprof/block --output block.pprof
-    curl -ksS http://localhost:13133/debug/pprof/mutex --output mutex.pprof
-    curl -ksS http://localhost:13133/debug/pprof/profile --output profile.pprof
-    curl -ksS http://localhost:13133/debug/pprof/trace?seconds=5 > profile.pb.gz
+    curl -ksS "$ENDPOINT/debug/pprof/goroutine" --output goroutines.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/heap" --output heap.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/threadcreate" --output threadcreate.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/block" --output block.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/mutex" --output mutex.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/profile" --output profile.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/trace?seconds=5" > profile.pb.gz
     tar_filename="pprof_$(date +%Y%m%d_%H%M%S).tar.gz"
     tar -czvf $tar_filename goroutines.pprof heap.pprof threadcreate.pprof block.pprof mutex.pprof profile.pprof profile.pb.gz
 
--- a/scripts/support/create-support-bundle.sh
+++ b/scripts/support/create-support-bundle.sh
@@ -303,7 +303,7 @@ function bundle_files() {
 
 collect_profiles() {
   # shellcheck disable=SC2162
-  read -p "Collect go pprof? (y/n) " PPROF
+  read -p "Collect go pprof profiles? (y/n) " PPROF
   if [[ "$PPROF" == y*  ]]; then
     tar_filename="$1"
     increase_indent
--- a/scripts/support/create-support-bundle.sh
+++ b/scripts/support/create-support-bundle.sh
@@ -77,6 +77,7 @@ success() { printf "$fg_green$*$reset\\n" ; }
 
 bindplane_banner()
 {
+  # shellcheck disable=SC2154
   if [ "$non_interactive" = "false" ]; then
     fg_cyan " oooooooooo.   o8o                    .o8  ooooooooo.   oooo\\n"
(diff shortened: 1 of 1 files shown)


scripts/support/create-support-bundle.sh
```diff
@@ -273,27 +279,54 @@ function bundle_files() {
...
+    printf "\n"
+    curl -ksS "$ENDPOINT/debug/pprof/goroutine" --output goroutines.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/heap" --output heap.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/threadcreate" --output threadcreate.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/block" --output block.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/mutex" --output mutex.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/profile" --output profile.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/trace?seconds=5" > profile.pb.gz
+    tar_filename="pprof_$(date +%Y%m%d_%H%M%S).tar.gz"
+    tar -czvf "$tar_filename" goroutines.pprof heap.pprof threadcreate.pprof block.pprof mutex.pprof profile.pprof profile.pb.gz
```
dpaasman00, line 324: Can we add something to clean up the different pprof files that are created from this once we've packaged them into the tarball? This just creates a lot of clutter that the user will need to then clean up

scripts/support/create-support-bundle.sh
```diff
@@ -75,19 +75,20 @@ error() {
...
+  # shellcheck disable=SC2154
+  if [ "$non_interactive" = "false" ]; then
+    fg_cyan " oooooooooo.   o8o                    .o8  ooooooooo.   oooo\\n"
+    fg_cyan " '888'   '88b  '\"'                   \"888  '888   'Y88. '888\\n"
+    fg_cyan "  888     888 oooo  ooo. .oo.    .oooo888   888   .d88'  888   .oooo.   ooo. .oo.    .ooooo.\\n"
+    fg_cyan "  888oooo888' '888  '888P\"Y88b  d88' '888   888ooo88P'   888  'P  )88b  '888P\"Y88b  d88' '88b\\n"
+    fg_cyan "  888    '88b  888   888   888  888   888   888          888   .oP\"888   888   888  888ooo888\\n"
+    fg_cyan "  888    .88P  888   888   888  888   888   888          888  d8(  888   888   888  888    .o\\n"
+    fg_cyan " o888bood8P'  o888o o888o o888o 'Y8bod88P\" o888o        o888o 'Y888\"\"8o o888o o888o '88bod8P'\\n"
+
```
dpaasman00, line 89: Pretty nit, but this should be "Bindplane" to go along with the recent brand changes

scripts/support/create-support-bundle.sh
```diff
@@ -273,27 +279,54 @@ function bundle_files() {
...
+    read -p "Endpoint: " ENDPOINT
+    printf "\n"
+    curl -ksS "$ENDPOINT/debug/pprof/goroutine" --output goroutines.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/heap" --output heap.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/threadcreate" --output threadcreate.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/block" --output block.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/mutex" --output mutex.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/profile" --output profile.pprof
+    curl -ksS "$ENDPOINT/debug/pprof/trace?seconds=5" > profile.pb.gz
+    tar_filename="pprof_$(date +%Y%m%d_%H%M%S).tar.gz"
```
dpaasman00, line 323: Are we able to append these profiles to the support tarball that's already created? That way the user still has just one tarball to send not 2?

scripts/support/create-support-bundle.sh
```diff
@@ -75,19 +75,20 @@ error() {
...
-  fg_cyan "   .d88b.  88888b.  .d8888b   .d88b.  888d888 888  888  888  888     888\\n"
-  fg_cyan "  d88\"\"88b 888 \"88b 88K      d8P  Y8b 888P\"   888  888  888  888     888\\n"
-  fg_cyan "  888  888 888  888 \"Y8888b. 88888888 888     Y88  88P  888  888 Y8b 888\\n"
-  fg_cyan "  Y88..88P 888 d88P      X88 Y8b.     888      Y8bd8P   888  Y88b.Y8b88P\\n"
-  fg_cyan "   \"Y88P\"  88888P\"   88888P'  \"Y8888  888       Y88P  8888888 \"Y888888\"\\n"
-  fg_cyan "                                                                   Y8b  \\n"
-
-  reset
+  # shellcheck disable=SC2154
+  if [ "$non_interactive" = "false" ]; then
```
dpaasman00, line 81: This banner didn't get printed when I ran it. I don't think this `non_interactive` variable is used elsewhere, so I think this just needs to get set.

dpaasman00 (changes requested):
Few changes, otherwise it worked well collecting profiles.

dpaasman00:
Is it possible to collect profiles on Windows at all?

scripts/support/create-support-bundle.sh
Dylan-M: All of our other scripts are using this version. When we change it everywhere else, it can be changed here too. I'm not our ASCII artist ;)

scripts/support/create-support-bundle.sh
Dylan-M: I believe this was someone's bad copy/pasta from another script where we do use it. (Maybe even mine), but it is in a bunch of our scripts where it likely shouldn't be. I'll address in my next revision.

Dylan-M:
> Is it possible to collect profiles on Windows at all?

I don't know, but I was going to look at that as a separate capability in a PowerShell script. In a PR unique to that script.

scripts/support/create-support-bundle.sh
Dylan-M: Can and will do. I meant to, and just forgot. Thanks.

dpaasman00 (approved):

PR #2194, opened 2025-02-27
feat: Add Google Cloud Storage Exporter

kuiperda:
### Proposed Change

Add GCS exporter. It heavily copies from the Azure Blob Exporter which is very similar. The main differences are in the config and storage client. 

It will marshal telemetry into OTLP JSON and create a bucket > folders (year/month/day/hour/minute) > object containing the data, based on the timestamp and configuration. 

Commit history is a mess, sorry. 

You can test pretty easily in Bindplane:
1. Create a config with Bindplane Agent source and Custom destination
2. Install an Agent
3. Build collector from this branch and replace the binary in the installed agent
4. Test in project `iris-dev`, I have a service account with the right permissions set up already. 
5. See readme for configuration options

It handles when the bucket already exists (and you have access), when the bucket needs to be created, and when the bucket already exists but you don't have access. 

##### Checklist
- [x] Changes are tested
- [x] CI has passed


exporter/googlecloudstorageexporter/config.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Validate validates the config.
+func (c *Config) Validate() error {
+	if c.BucketName == "" {
+		return errors.New("bucket_name is required")
+	}
+	if c.ProjectID == "" {
+		return errors.New("project_id is required")
+	}
```
jsirianni, line 85: Generally the Google SDK will detect a project ID if one is not set (based on the configured credentials, or credentials from the environment). Is that possible here?

jsirianni (changes requested):
I have not tried the exporter, just a question around project id.

exporter/googlecloudstorageexporter/config.go
kuiperda: Yeah, I noticed that after finishing most of the work. I can make that enhancement.

exporter/googlecloudstorageexporter/config.go
```diff
@@ -0,0 +1,107 @@
...
+	// Valid values are "hour" or "minute". Default: minute
+	Partition partitionType `mapstructure:"partition"`
+
+	// Compression is the type of compression to use.
+	// Valid values are "none" or "gzip". Default: none
+	Compression compressionType `mapstructure:"compression"`
+
+	exporterhelper.TimeoutConfig `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
+	exporterhelper.QueueConfig   `mapstructure:"sending_queue"`
+	configretry.BackOffConfig    `mapstructure:"retry_on_failure"`
```
jsirianni, line 75: It would be nice to include these options in the readme, even if it means linking to upstream docs.

Does the exporter support persistent queue with storage extension?

jsirianni (changes requested):
I tried some basic functionality with metric and logs using compression. It is working great. I left one more question.

exporter/googlecloudstorageexporter/exporter.go
```diff
@@ -0,0 +1,136 @@
...
+	logger        *zap.Logger
+	marshaler     marshaler
+}
+
+// newExporter creates a new Google Cloud Storage exporter
+func newExporter(cfg *Config, params exporter.Settings) (*googleCloudStorageExporter, error) {
+	storageClient, err := newGoogleCloudStorageClient(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create storage client: %w", err)
+	}
```
dpaasman00, line 45: General comment to apply through out the pr, but for errors can we make sure they're all wrapped with a message like is done here. However, we shouldn't include "failed to..." or "error while..." just because it'll be propagated as an error and when you get several errors deep it becomes a bit excessive.
```suggestion
	if err != nil {
		return nil, fmt.Errorf("create storage client: %w", err)
	}
```

exporter/googlecloudstorageexporter/factory.go
```diff
@@ -0,0 +1,118 @@
...
+
+func createMetricsExporter(ctx context.Context, params exporter.Settings, config component.Config) (exporter.Metrics, error) {
+	cfg, ok := config.(*Config)
+	if !ok {
+		return nil, errors.New("not a Google Cloud Storage config")
+	}
+	exp, err := newExporter(cfg, params)
+	if err != nil {
+		return nil, err
+	}
```
dpaasman00, line 65: Can this error and others get wrapped
```suggestion
	if err != nil {
		return nil, fmt.Errorf("new exporter: %w", err)
	}
```

exporter/googlecloudstorageexporter/config.go
kuiperda: I am not sure, I copied these from the Azure Blob exporter. I can link to (and read the) upstream docs.

dpaasman00 (changes requested):
Couple minor comments, otherwise it's looking good

exporter/googlecloudstorageexporter/storage_client.go
```diff
@@ -0,0 +1,156 @@
...
+	default:
+		// Find application default credentials from the environment
+		creds, err := google.FindDefaultCredentials(ctx, storage.ScopeFullControl)
+		if err != nil {
+			return nil, fmt.Errorf("find default credentials: %w", err)
+		}
+		opts = append(opts, option.WithCredentials(creds))
+		if cfg.ProjectID == "" {
+			cfg.ProjectID = creds.ProjectID
+		}
```
kuiperda, line 79: Default creds will get read even without setting opts here, but this follows the pattern of the GCP exporter.

dpaasman00 (approved):

PR #2195, opened 2025-02-27
feat(secops): Exporter metrics

jsirianni:
### Proposed Change

Would like some feedback before I implement the gosec fixes.

Implemented three new metrics. Batch size, request latency, payload size. These will be useful when determining batch processor settings and exporter queue consumer count.

I could use some help on bucket sizing. What I have right now seems to be okay, not sure if it is overboard or not granular enough.

Screenshot From 2025-03-07 12-48-12

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


dpaasman00 (changes requested):
One quick nit

dpaasman00 (changes requested):

dpaasman00 (approved):
Metrics were working well

PR #2196, opened 2025-03-03
Fix: all outdated branding from observIQ to Bindplane

BominRahmani:
The following goes back through the code base and finds all non code based instances of observIQ and changes it to "Bindplane" this includes the licenses.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,3 @@
-
                                  Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
@@ -187,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright observIQ Inc.
+   Copyright Bindplane Inc.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,7 @@ check-license:
 # This target adds a license copyright header is on every source file that is missing one
 .PHONY: add-license
 add-license:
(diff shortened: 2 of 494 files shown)


LICENSE
```diff
@@ -187,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright observIQ Inc.
+   Copyright Bindplane Inc.
```
dpaasman00, line 189: When it comes to legal stuff like this license I'm not sure we can change it to Bindplane. I would ask Greg

Makefile
```diff
@@ -182,7 +182,7 @@ check-license:
 # This target adds a license copyright header is on every source file that is missing one
 .PHONY: add-license
 add-license:
-	@ADDLICENSEOUT=`addlicense -y "" -c "observIQ, Inc." $(ALL_SRC) 2>&1`; \
+	@ADDLICENSEOUT=`addlicense -y "" -c "Bindplane, Inc." $(ALL_SRC) 2>&1`; \
```
dpaasman00, line 185: Same concern here

buildscripts/download-dependencies.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Copyright  observIQ, Inc.
+# Copyright Bindplane, Inc.
```
dpaasman00, line 2: Same concerns with this statement too

exporter/azureblobexporter/blob_client.go
```diff
@@ -1,4 +1,4 @@
-// Copyright observIQ, Inc.
+// CopyrightBindplane, Inc.
```
dpaasman00, line 1: A bunch of these are missing a space

dpaasman00 (changes requested):

PR #2210, opened 2025-03-06
deps: Use correct `nop` receiver import

dpaasman00:
### Proposed Change

Turns out we're using the wrong nop receiver import, receivertest, instead of the legit one which is causing issues with available components in bp

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


colelaven (approved):
good catch haha

jsirianni (approved):

PR #2211, opened 2025-03-07
feat: bindplane audit logs receiver 

justinianvoss22:
### Proposed Change
- New receiver for Bindplane Audit Logs.
- Uses Bindplane API endpoint `/audit-events`
- Sort by newest audit log first

##### Checklist
- [x ] Changes are tested
- [ x] CI has passed


receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (r *bindplaneAuditLogsReceiver) getLogs(ctx context.Context) []AuditLogEvent {
+	var logs []AuditLogEvent
+
+	reqURL := r.cfg.BindplaneURL
+	reqURL = strings.TrimRight(reqURL, "/")
+	if !strings.Contains(reqURL, "://") {
+		reqURL = "http://" + reqURL
+	}
+	reqURL = reqURL + "/v1/audit-events"
```
jsirianni, line 119: I think it would be better if r.cfg.BindplaneURL was type `url.URL`. Then you can avoid trimming the path because the URL type's methods handle it for you.

```go
reqURL := r.cfg.BindPlaneURL
reqURL.Path = "/v1/audit-events"
...
...
req, err := http.NewRequestWithContext(ctx, "GET", reqURL.String(), nil)
```

This way you do not need to handle edge cases that should have been caught during start validation. Also, sometimes URL's can have funny characters (e.g. IPv6).

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,226 @@
...
+	return nil
+}
+
+func (r *bindplaneAuditLogsReceiver) getLogs(ctx context.Context) []AuditLogEvent {
+	var logs []AuditLogEvent
+
+	reqURL := r.cfg.BindplaneURL
+	reqURL = strings.TrimRight(reqURL, "/")
+	if !strings.Contains(reqURL, "://") {
+		reqURL = "http://" + reqURL
```
jsirianni, line 117: Similar to my other comment, we should require the user to include a scheme in their configuration, and handle that check during startup validation.

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	reqURL = reqURL + "/v1/audit-events"
+
+	req, err := http.NewRequestWithContext(ctx, "GET", reqURL, nil)
```
jsirianni, line 121: HTTP requests should include a reasonable timeout on the context to ensure requests are terminated if they run too long (bindplane server misbehaving, etc).

```go
// timeout should probably be a const defined somewhere
ctx, cancel := context.WithTimeout(ctx, time.Second * 10)
```

10 seconds may not be reasonable, please using something shorter or longer that makes sense according to your testing. I believe Bindplane server has a 2 minute timeout for all requests, we should use a shorter timeout to ensure timeout errors are correctly handled on the client.

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,226 @@
...
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		r.logger.Error("error reading response", zap.Error(err))
+		return logs
+	}
+
+	var response apiResponse
+	err = json.Unmarshal(body, &response)
+	if err != nil {
+		r.logger.Error("unable to unmarshal log events", zap.Error(err), zap.String("body", string(body)))
```
jsirianni, line 154: We should avoid logging the response body because it could leak audit data to the log file. It might be okay in this case, however, a client misconfiguration (incorrect apiResponse struct) could fail despite the data being full of valid events, getting logged here.

```go
		r.logger.Error("unable to unmarshal log events", zap.Error(err))
```

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,226 @@
...
+		if !latestTime.IsZero() {
+			r.lastTimestamp = latestTime
+		}
+	}
+
+	return logs
+}
+
+func (r *bindplaneAuditLogsReceiver) processLogEvents(observedTime pcommon.Timestamp, logEvents []AuditLogEvent) plog.Logs {
+	logs := plog.NewLogs()
```
jsirianni, line 182: Can we include resource attributes for describing the Bindplane server? Something like `bindplane_url` or anything else that might be useful for filtering audit events based on the originating server.

jsirianni (changes requested):
This looks great. I have not tried it yet, left some feedback.

@justinianvoss22 will this work well if pointed at a load balancer with multiple bindplane servers? I suspect it would but want to make sure we consider it.

receiver/bindplaneauditlogs/logs.go
justinianvoss22: Oops yeah that was for debugging purposes, forgot to delete

receiver/bindplaneauditlogs/config.go
```diff
@@ -0,0 +1,83 @@
...
+	c.BindplaneURL = URLConfig{URL: bindplaneURL}
+	if c.BindplaneURL.URL == nil {
+		return errors.New("bindplane_url must be initialized")
+	}
+
+	if c.PollInterval == 0 {
+		c.PollInterval = 10 * time.Second
+	} else if c.PollInterval < 10*time.Second || c.PollInterval > 24*time.Hour {
+		return errors.New("poll_interval must be between 10 seconds and 24 hours")
+	}
```
dpaasman00, line 80: If we set this parameter in the default config, then we don't need to worry about setting it here if it equals 0. This should just be:
```suggestion
        if c.PollInterval < 10*time.Second || c.PollInterval > 24*time.Hour {
		return errors.New("poll_interval must be between 10 seconds and 24 hours")
	}
```

receiver/bindplaneauditlogs/factory.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/observiq/bindplane-otel-collector/receiver/bindplaneauditlogs/internal/metadata"
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/receiver"
+)
+
+const (
+	// TypeStr is the unique identifier for the BindPlane audit logs receiver
+	TypeStr = "bindplaneauditlogs"
+)
```
dpaasman00, line 30: This is unused
```suggestion
```

receiver/bindplaneauditlogs/config.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// BindplaneURL is the URL of the Bindplane instance, taken from BindplaneURLString
+	BindplaneURL URLConfig `mapstructure:"bindplane_url"`
+
+	// PollInterval is the interval at which the receiver polls for new audit logs
+	PollInterval time.Duration `mapstructure:"poll_interval"`
+
+	// ClientConfig is the configuration for the HTTP client
+	confighttp.ClientConfig `mapstructure:",squash"`
+}
```
dpaasman00, line 48: This config structure needs to get reworked. Right now there are 3 different urls that can be specified in the configuration (`bindplane_url_string`, `bindplane_url.url`, and `endpoint` (this one is in the squashed ClientConfg)). 

My suggestion: Use the endpoint that's in the client config and delete the others. The yaml key is already just `endpoint`. If you still want to parse the endpoint into a `url.URL` object, add it to the config as an unexported parameter and don't add a mapstructure tag. Something like 
```
bindplaneURL *url.URL
```

dpaasman00: Make sure the README gets updated as well. I think it's currently out of date with what's in here at the moment.

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (r *bindplaneAuditLogsReceiver) getLogs(ctx context.Context) []AuditLogEvent {
+	var logs []AuditLogEvent
+	const timeout = 1 * time.Minute
+
+	// Defensive check for nil URL
+	if r.cfg.BindplaneURL.URL == nil {
+		r.logger.Error("BindplaneURL is not initialized")
+		return logs
+	}
```
dpaasman00, line 118: Not sure what the motivation for adding this check was, but if we're doing proper config validation this shouldn't be necessary.

dpaasman00 (changes requested):
First pass, some changes but generally looks good.

- Make sure you add an entry to the docs in this repo for this component. Look through Dan's recent Google Cloud Storage exporter pr for an example.

receiver/bindplaneauditlogs/README.md
```diff
@@ -0,0 +1,50 @@
...
+| bindplane_url_string | string |         | `true`   | The Bindplane URL the receiver should collect logs from. (e.g. `https://app.bindplane.com`)                                                                                                                                  |
+| api_key              | string |         | `true`   | The Bindplane API key with read access to audit logs.                                                                                                                                                                        |
+| scheme               | string |         | `false`  | The scheme of the Bindplane URL. If not provided, the scheme will be `https`.                                                                                                                                                |
+| poll_interval        | string | 1m      | `false`  | The rate at which this receiver will poll Bindplane for logs. This value must be in the range [10 seconds - 24 hours] and must be a string readable by Golang's [time.ParseDuration](https://pkg.go.dev/time#ParseDuration). |
```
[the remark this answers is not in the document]
justinianvoss22: Not sure what the best default value would be. Andy said 10 seconds was a good length for polling so I have stuck with that.

receiver/bindplaneauditlogs/README.md
```diff
@@ -0,0 +1,53 @@
...
+- A Bindplane API key with read access to audit logs.
+
+## Configuration
+
+| Field         | Type   | Default | Required | Description                                                                                                                                                                                                                  |
+| ------------- | ------ | ------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| api_key       | string |         | `true`   | The Bindplane API key with read access to audit logs. This API key has access to the audit logs of a single project.                                                                                                         |
+| scheme        | string |         | `false`  | The scheme of the Bindplane URL. If not provided, the scheme will be `https`.                                                                                                                                                |
```
dpaasman00, line 31: The scheme field isn't used in the receiver, it should get removed from here.

receiver/bindplaneauditlogs/config.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// Config defines the configuration for the Bindplane audit logs receiver
+type Config struct {
+
+	// APIKey is the authentication key for accessing BindPlane audit logs
+	APIKey string `mapstructure:"api_key"`
+
+	// Scheme is the URL scheme used to connect to the BindPlane audit logs API
+	Scheme string `mapstructure:"scheme"`
```
dpaasman00, line 34: This is unused.

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,232 @@
...
+	logs := r.processLogEvents(observedTime, logEvents)
+	if logs.LogRecordCount() > 0 {
+		if err := r.consumer.ConsumeLogs(ctx, logs); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *bindplaneAuditLogsReceiver) getLogs(ctx context.Context) []AuditLogEvent {
```
dpaasman00, line 111: This function should get reworked to return an error instead of logging an error. This way we can check if there was an error inside of `poll` before processing. If there was then we can return the error and `startPolling` will catch it and log the error message.

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,232 @@
...
+	}, nil
+}
+
+func (r *bindplaneAuditLogsReceiver) Start(_ context.Context, _ component.Host) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	r.cancel = cancel
+	r.wg.Add(1)
+	go r.startPolling(ctx)
+	return nil
+}
```
dpaasman00, line 76: This should be refactored to make use of the `confighttp.ToClient` method that's available on the squashed client config we're using. You'll have to wait until start though to create and assign it to the receiver struct bc it needs the component.Host that's passed in.

dpaasman00 (changes requested):

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,236 @@
...
+
+func (r *bindplaneAuditLogsReceiver) Start(ctx context.Context, host component.Host) error {
+	client, err := r.cfg.ToClient(ctx, host, r.settings)
+	if err != nil {
+		return fmt.Errorf("failed to create HTTP client: %w", err)
+	}
+	r.client = client
+
+	ctx, cancel := context.WithCancel(ctx)
```
dpaasman00, line 73: This should be made with `context.Background` since the passed in context will get canceled shortly after this func is called.
```suggestion
	ctx, cancel := context.WithCancel(context.Background())
```

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	observedTime := pcommon.NewTimestampFromTime(time.Now())
+	logs := r.processLogEvents(observedTime, logEvents)
+	if logs.LogRecordCount() > 0 {
+		if err := r.consumer.ConsumeLogs(ctx, logs); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
dpaasman00, line 114: I'd rework this function a bit so we can log some info if there's no logs to process
```suggestion
func (r *bindplaneAuditLogsReceiver) poll(ctx context.Context) error {
	logEvents, err := r.getLogs(ctx)
	if err != nil {
		return fmt.Errorf("getLogs: %w", err)
	}

	observedTime := pcommon.NewTimestampFromTime(time.Now())
	logs := r.processLogEvents(observedTime, logEvents)

	if !(logs.LogRecordCount() > 0) {
		r.logger.Debug("no logs to process")
		return nil
	}

	if err := r.consumer.ConsumeLogs(ctx, logs); err != nil {
		return fmt.Errorf("consumeLogs: %w", err)
	}
	return nil
```

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, "GET", reqURL.String(), nil)
+	if err != nil {
+		err = fmt.Errorf("error creating request: %w", err)
+		return nil, err
+	}
```
dpaasman00, line 132: The errors in this function should be wrapped in the return statement. Also they should not have the word "error" or "failed" in the error message itself. The errors will get logged as an error so it'll be redundant and take up a lot of extra space in the log. As a general rule of thumb, the error msg should tell you what function produced the error. This is an example of how to wrap one of these errors.
```suggestion
	req, err := http.NewRequestWithContext(ctx, "GET", reqURL.String(), nil)
	if err != nil {
		return nil, return nil, fmt.Errorf("newRequestWithContext: %w", err)
	}
```

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		err = fmt.Errorf("error making request: %w", err)
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		err = fmt.Errorf("non-200 response: %s", res.Status)
+		return nil, err
+	}
```
dpaasman00, line 150: This error message is okay because we're not just wrapping one that was already produced. The wrapping should still get moved to the return statement though.

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// newBindplaneAuditLogsReceiver returns a newly configured bindplaneAuditLogsReceiver
+func newBindplaneAuditLogsReceiver(cfg *Config, logger *zap.Logger, consumer consumer.Logs) (*bindplaneAuditLogsReceiver, error) {
+	return &bindplaneAuditLogsReceiver{
+		cfg:           *cfg,
+		consumer:      consumer,
+		logger:        logger,
+		wg:            &sync.WaitGroup{},
+		lastTimestamp: time.Now().UTC().Add(-cfg.PollInterval),
```
dpaasman00, line 62: For setting this initial lastTimestamp value, have we considered that bindplane has a configurable retention policy and the default is 30 days? So if we start this receiver with this lastTimestamp, we could miss a lot of audit log events, especially considering if the retention policy in bindplane is significantly greater than the default of 30 days.

dpaasman00 (changes requested):

receiver/bindplaneauditlogs/README.md
```diff
@@ -0,0 +1,52 @@
...
+| endpoint      | string |         | `true`   | The endpoint to collect logs from. (e.g. `https://app.bindplane.com`)                                                                                                                                                        |
+| poll_interval | string | 10s     | `false`  | The rate at which this receiver will poll Bindplane for logs. This value must be in the range [10 seconds - 24 hours] and must be a string readable by Golang's [time.ParseDuration](https://pkg.go.dev/time#ParseDuration). |
+
+### Example Configuration
+
+```yaml
+receivers:
+  bindplaneauditlogs:
+    api_key: 1234567890
+    scheme: https
```
dpaasman00, line 40: scheme is still present here

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,217 @@
...
+
+	return logs
+}
+
+func (r *bindplaneAuditLogsReceiver) Shutdown(_ context.Context) error {
+	r.logger.Debug("shutting down logs receiver")
+	if r.cancel != nil {
+		r.cancel()
+	}
+	r.client.CloseIdleConnections()
```
dpaasman00, line 214: This should now check if client is nil before calling this func

dpaasman00: It should also go after the call the `r.wg.Wait()`

receiver/bindplaneauditlogs/README.md
```diff
@@ -0,0 +1,50 @@
...
+- Introduced: [v1.70.0](https://github.com/observIQ/bindplane-otel-collector/releases/tag/v1.70.0)
+
+## Supported Pipelines
+
+- Logs
+
+## How It Works
+
+1. The user configures this receiver in a pipeline.
+2. The user configures a supported component to route telemetry from this receiver.
```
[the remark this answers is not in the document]
jsirianni: Good question. I think it could be common for an org admin to create an API key in each project so they can gather audit logs for one or many projects. This could get messy because the API keys do not have friendly names.

jsirianni (changes requested):
I did not think of this until now. We should have project and project_id resource attributes in order to filter between multiple projects. Similar to the bindplane_url resource I recommended previously.

I know you can list projects with an API key, but I am not sure if we can detect which project the API key is for. If this information is already present. Can you share an example audit log with me, in OTLP format (after its been processed by the receiver)?

receiver/bindplaneauditlogs/logs.go
dpaasman00: I think there was a misunderstanding here. We should not have gotten rid of the lastTimestamp functionality, just that it shouldn't be set initially. The first poll to bindplane shouldn't use one so we can get all existing audit events. After the first poll then we can start using this lastTimestamp parameter so we don't continually poll the same events.

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,219 @@
...
+			attrs.PutStr("configuration", logEvent.Configuration)
+		}
+		attrs.PutStr("action", string(logEvent.Action))
+		attrs.PutStr("user", logEvent.User)
+		if logEvent.Account != "" {
+			attrs.PutStr("account", logEvent.Account)
+		}
+
+		resourceAttributes := logRecord.Attributes()
+		resourceAttributes.PutStr("bindplane_url", r.cfg.Endpoint)
```
dpaasman00, line 203: This isn't a resource attribute, this is just a normal attribute on the log

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,219 @@
...
+
+	// Sort by timestamp (newest first)
+	sort.Slice(logs, func(i, j int) bool {
+		return logs[i].Timestamp.After(*logs[j].Timestamp)
+	})
+
+	return logs, nil
+}
+
+func (r *bindplaneAuditLogsReceiver) processLogEvents(observedTime pcommon.Timestamp, logEvents []AuditLogEvent) plog.Logs {
```
dpaasman00, line 169: Not sure how I feel about having an empty log body and only setting attributes. @jsirianni what do you think? There isn't really an obvious log body that's returned by the audit events. There is a description field which would be great, but all the events I've seen this value is empty and wouldn't provide any value.

dpaasman00 (changes requested):

dpaasman00 (changes requested):

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,231 @@
...
+	var logs []AuditLogEvent
+	const timeout = 1 * time.Minute
+
+	reqURL := *r.cfg.bindplaneURL
+	reqURL.Path = "/v1/audit-events"
+
+	// Only add the since parameter if we have a lastTimestamp
+	if r.lastTimestamp != nil {
+		q := reqURL.Query()
+		q.Add("since", r.lastTimestamp.Format(bindplaneTimeFormat))
```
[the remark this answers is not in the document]
justinianvoss22: @dpaasman00 Yes, this parameter works well with the query. For example, this query: `curl -X GET -v "http://localhost:3000/v1/audit-events?since=2024-03-01T12:00:00Z" -H "X-Bindplane-Api-Key: ec835dbe-60d9-4b77-820d-52e3d5674a23" ` returns audit logs since a certain point.

receiver/bindplaneauditlogs/logs.go
jsirianni: Spoke with Dakota offline. I think we could move description to the body as a string body: <description value>, but we would need to make sure all audit events have descriptions. Even something generic.

k8s cluster events are similar.

Screenshot From 2025-03-14 11-35-04

justinianvoss22: How is this for a generic description? 
image
Or should it be even more generic

receiver/bindplaneauditlogs/logs.go
```diff
@@ -0,0 +1,238 @@
...
+			logRecord.SetTimestamp(pcommon.NewTimestampFromTime(*logEvent.Timestamp))
+		}
+
+		// Set body as description
+		body := logRecord.Body()
+		if logEvent.Description != "" {
+			body.SetStr(logEvent.Description)
+		} else {
+			body.SetStr(fmt.Sprintf("Audit log event - Action: %s, Resource: %s, User: %s", logEvent.Action, logEvent.ResourceName, logEvent.User))
+		}
```
dpaasman00, line 205: Since there's still a chance old audit events have an empty description we should still check
```suggestion
		// Set body as description
		if logEvent.Description != "" {
			logRecord.Body().SetStr(logEvent.Description)
		}
```

dpaasman00 (changes requested):
One last change then we should be good

jsirianni (approved):
Im good if @dpaasman00 is

go.mod
```diff
@@ -208,6 +208,7 @@ require (
 
 require (
 	github.com/observiq/bindplane-otel-collector/processor/topologyprocessor v1.73.0
+	github.com/observiq/bindplane-otel-collector/receiver/bindplaneauditlogs v0.0.0-00010101000000-000000000000
```
dpaasman00, line 211:
```suggestion
	github.com/observiq/bindplane-otel-collector/receiver/bindplaneauditlogs v1.73.0
```

receiver/bindplaneauditlogs/go.mod
```diff
@@ -0,0 +1,74 @@
...
+require (
+	github.com/stretchr/testify v1.10.0
+	go.opentelemetry.io/collector/component/componenttest v0.121.0
+	go.opentelemetry.io/collector/config/confighttp v0.121.0
+	go.opentelemetry.io/collector/confmap v1.24.0
+	go.opentelemetry.io/collector/consumer/consumertest v0.118.0
+	go.opentelemetry.io/collector/pdata v1.27.0
+	go.opentelemetry.io/collector/receiver/receivertest v0.118.0
+	go.uber.org/goleak v1.3.0
+	go.uber.org/zap v1.27.0
```
dpaasman00, line 16: These versions for otel need to get updated... latest contrib version is `v0.122.0` and latest stable is `v1.28.0`

dpaasman00 (changes requested):
Couple more changes but this is working well and almost there

dpaasman00 (approved):

PR #2249, opened 2025-03-13
fix: Ticker bug in gRPC metrics collection

shazlehu:
### Proposed Change

Ticker was cancelled immediately after creation in hostMetricsReporter.start()

This was a regression introduced in `v1.67.0`

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/chronicleexporter/hostmetrics.go
+++ b/exporter/chronicleexporter/hostmetrics.go
@@ -85,13 +85,16 @@ func newHostMetricsReporter(cfg *Config, set component.TelemetrySettings, export
 
 func (hmr *hostMetricsReporter) start() {
 	ticker := time.NewTicker(5 * time.Minute)
-	defer ticker.Stop()
-
 	ctx, cancel := context.WithCancel(context.Background())
 	hmr.cancel = cancel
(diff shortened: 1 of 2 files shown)


exporter/chronicleexporter/hostmetrics.go
```diff
@@ -85,13 +85,16 @@ func newHostMetricsReporter(cfg *Config, set component.TelemetrySettings, export
 
 func (hmr *hostMetricsReporter) start() {
 	ticker := time.NewTicker(5 * time.Minute)
```
jsirianni, line 87: I think we can declare this in the go routine to make it more clear.

jsirianni (changes requested):
Looks good, one comment. @tbm48813 to build and test.

exporter/chronicleexporter/hostmetrics.go
shazlehu: Agreed

dpaasman00 (approved):
LGTM once Travis can test

tbm48813 (approved):

PR #2262, opened 2025-03-19
Revert "chore: Update modules to v1.74.0", bump modules to v1.73.1, replace `stanza`

dpaasman00:
Doing a patch release, so reverting the last module version bump and going to v1.73.1. 

Since only change for patch is replacing the stanza pkg, including that replace in this PR too to avoid running CI more than necessary.

### DIFF
--- a/receiver/azureblobrehydrationreceiver/go.mod
+++ b/receiver/azureblobrehydrationreceiver/go.mod
@@ -5,8 +5,8 @@ go 1.23.6
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.17.0
 	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.0
-	github.com/observiq/bindplane-otel-collector/internal/rehydration v1.74.0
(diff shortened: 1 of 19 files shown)


receiver/azureblobrehydrationreceiver/go.mod
```diff
@@ -5,8 +5,8 @@ go 1.23.6
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.17.0
 	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.0
-	github.com/observiq/bindplane-otel-collector/internal/rehydration v1.74.0
-	github.com/observiq/bindplane-otel-collector/internal/testutils v1.74.0
+	github.com/observiq/bindplane-otel-collector/internal/rehydration v1.62.0
```
mrsillydog, line 8: Let's run the script with v1.73.1 a second time. This should probably be added to the release workflow docs.

mrsillydog (approved):
Assuming CI passses

PR #2265, opened 2025-03-20
feat: Add S3 event receiver

Caleb-Hurshman:
### Proposed Change

Adds the AWS S3 event receiver. The receiver reads object creation notifications from AWS SQS, fetches the associated objects, and sends them as logs.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


receiver/awss3eventreceiver/metadata.go
colelaven, line 1: I think we should delete this file, if you follow the pattern for other receivers, it should be autogenerated from `metadata.yaml` with a go:generate command that lives in `doc.go`

receiver/awss3eventreceiver/doc.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package awss3eventreceiver implements a receiver that consumes S3 event notifications
+// from SQS and processes the objects containing OTLP data.
+//
+// The receiver polls an SQS queue for S3 event notifications. When an object creation
+// event is received, the receiver downloads the S3 object and processes it as OTLP data
+// (metrics, logs, or traces).
+package awss3eventreceiver // import "github.com/observiq/bindplane-otel-collector/receiver/awss3eventreceiver"
```
colelaven, line 21:
```suggestion
//go:generate mdatagen metadata.yaml

package awss3eventreceiver // import "github.com/observiq/bindplane-otel-collector/receiver/awss3eventreceiver"
```

colelaven: youll also need to run a make generate

receiver/awss3eventreceiver/config_test.go
```diff
@@ -0,0 +1,182 @@
...
+	cfg := f.CreateDefaultConfig().(*Config)
+	cfg.SQSQueueURL = "https://sqs.us-east-1.amazonaws.com/123456789012/test-queue"
+
+	assert.NoError(t, cfg.Validate())
+}
+
+func TestLoadConfig(t *testing.T) {
+	t.Parallel()
+
+	cm, err := confmaptest.LoadConf(filepath.Join("testdata", "test_config.yaml"))
```
colelaven, line 41: nit: I dont think testing a config being loaded from a file adds much to these tests, its not really used much else in the repo. We could probably just get rid of this test and the testdata folder imo.

colelaven (changes requested):
first pass
basic functionality is working great, havent messed with configuring any options besides the endpoint
mostly left comments on file structure stuff

I will test more complex functionality and look more into the receiver code tomorrow.

receiver/awss3eventreceiver/config_test.go
djaglowski: Disagree. It's closer to a real world test than most and should be done for all our components.

receiver/awss3eventreceiver/go.mod
```diff
@@ -0,0 +1,79 @@
...
+go 1.23.6
+
+toolchain go1.24.0
+
+require (
+	github.com/aws/aws-lambda-go v1.47.0
+	github.com/aws/aws-sdk-go-v2 v1.24.0
+	github.com/aws/aws-sdk-go-v2/config v1.25.12
+	github.com/aws/aws-sdk-go-v2/service/s3 v1.47.3
+	github.com/aws/aws-sdk-go-v2/service/sqs v1.29.3
```
colelaven, line 12: just curious, any reason to use these specific versions of the aws-sdk dependencies? Looks like other aws components in the agent use newer ones.

receiver/awss3eventreceiver/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+| Field                  | Type   | Default | Required | Description |
+|------------------------|--------|---------|----------|-------------|
+| sqs_queue_url          | string |         | `true`   | The URL of the SQS queue to poll for S3 event notifications (the AWS region is automatically extracted from this URL) |
+| standard_poll_interval | duration | 15s   | `false`  | The interval at which the SQS queue is polled for messages |
+| max_poll_interval      | duration | 2m   | `false`  | The maximum interval at which the SQS queue is polled for messages |
+| polling_backoff_factor | float    | 2     | `false`  | The factor by which the polling interval is multiplied after an unsuccessful poll |
+| workers                | int      | 5     | `false`  | The number of workers to process messages in parallel |
+| visibility_timeout     | duration | 300s  | `false`  | The visibility timeout for SQS messages |
+| max_log_size           | int      | 1048576  | `false`  | The maximum size of a log record in bytes |
```
colelaven, line 24: What happens if a log record is larger than this size? does it get skipped?

receiver/awss3eventreceiver/README.md
```diff
@@ -0,0 +1,54 @@
...
+| max_logs_emitted       | int      | 1000  | `false`  | The maximum number of log records to emit |
```
colelaven, line 25: is this max logs emitted per poll?

receiver/awss3eventreceiver/metadata.yaml
```diff
@@ -0,0 +1,20 @@
+type: s3event
+
+status:
+  class: receiver
+  stability:
+    development: [logs]
```
colelaven, line 6: any reason to keep this at development not alpha?

Caleb-Hurshman: Not anymore.

receiver/awss3eventreceiver/go.mod
Caleb-Hurshman: I'm not sure, Dan would know more. I'd assume it's fine to use the newer versions, assuming it's not a jump in the major version.

receiver/awss3eventreceiver/factory.go
```diff
@@ -0,0 +1,66 @@
...
+		receiver.WithLogs(createLogsReceiver, metadata.LogsStability),
+	)
+}
+
+// createDefaultConfig creates a default configuration
+func createDefaultConfig() component.Config {
+	return &Config{
+		StandardPollInterval: 5 * time.Second,
+		MaxPollInterval:      60 * time.Second,
+		PollingBackoffFactor: 1.5,
```
colelaven, line 46: These default values are different than the default values listed in the README and the ones used in the BP source.
We should standardize to one or another.

Caleb-Hurshman: Good catch. That's interesting, bc it's also different in the README for the receiver, which is what I used for the source.

colelaven (approved):
working great, code looks good, apologies for the delay in review

PR #2271, opened 2025-03-24
feat: Windows Event Trace Log Receiver

schmikei:
### Proposed Change

ETW is a kernel-level tracing facility provided by Windows that allows applications, services, and the operating system itself to emit structured events. It's designed as a high-performance, low-overhead tracing system that can handle large volumes of data without significantly impacting system performance.

- Decided to do a separate receiver within this repo because ETL logs are ingested using real-time sessions
- A windows event log requires channel names and XPath queries whereas ETL format logs do not.
- Each addresses different observability needs and use cases
- Windows Event Logs are more general-purpose and often higher-level

A lot of the use cases for these logs are around troubleshooting. One thing to note is that since we open a real time subscription to ingest these logs we cannot employ a bookmark approach like what we can do with the WEL receiver. Other company agents also note this as a caveat.

> The WEL Source in Edge supports reading events from ETL logs, but using them for continuous event collection may result in dropped events and higher resource usage.

> They cannot be read when set to Overwrite mode and they don’t support seeking to skip past previously collected events

Configuration

```yaml
receivers:
  windowseventtrace:
    session_name: OtelCollectorETW
    providers:
      - name: 'Microsoft-Windows-DNS-Client'
      - name: 'Microsoft-Windows-PowerShell'
      - name: 'Microsoft-Windows-Kernel-File'

exporters:
  debug:
  file:
    path: ./test/output.json

service:
  pipelines:
    logs:
      receivers: [windowseventtrace]
      exporters: [debug, file]
```

Sample OTLP
```json
{
  "resourceLogs": [
    {
      "resource": {},
      "scopeLogs": [
        {
          "scope": {},
          "logRecords": [
            {
              "timeUnixNano": "1743624227770907800",
              "severityNumber": 9,
              "body": {
                "kvlistValue": {
                  "values": [
                    {
                      "key": "channel",
                      "value": {
                        "stringValue": "Microsoft-Windows-DNS Client Events/Operational"
                      }
                    },
                    {
                      "key": "computer",
                      "value": { "stringValue": "dev-vm-20250321-190819" }
                    },
                    { "key": "thread_id", "value": { "stringValue": "7544" } },
                    {
                      "key": "level",
                      "value": {
                        "kvlistValue": {
                          "values": [
                            {
                              "key": "name",
                              "value": { "stringValue": "Information" }
                            },
                            { "key": "value", "value": { "stringValue": "4" } }
                          ]
                        }
                      }
                    },
                    {
                      "key": "opcode",
                      "value": {
                        "kvlistValue": {
                          "values": [
                            {
                              "key": "name",
                              "value": { "stringValue": "Info" }
                            },
                            { "key": "value", "value": { "stringValue": "0" } }
                          ]
                        }
                      }
                    },
                    {
                      "key": "task",
                      "value": { "stringValue": "DnsServerForInterface" }
                    },
                    {
                      "key": "provider",
                      "value": {
                        "kvlistValue": {
                          "values": [
                            {
                              "key": "name",
                              "value": {
                                "stringValue": "Microsoft-Windows-DNS-Client"
                              }
                            },
                            {
                              "key": "guid",
                              "value": {
                                "stringValue": "{1C95126E-7EEA-49A9-A3FE-A378B03DDB4D}"
                              }
                            }
                          ]
                        }
                      }
                    },
                    {
                      "key": "event_data",
                      "value": {
                        "kvlistValue": {
                          "values": [
                            { "key": "Index", "value": { "stringValue": "1" } },
                            {
                              "key": "DynamicAddress",
                              "value": { "stringValue": "dynamic" }
                            },
                            {
                              "key": "AddressLength",
                              "value": { "stringValue": "16" }
                            },
                            {
                              "key": "Address",
                              "value": { "stringValue": "10.128.0.1" }
                            },
                            {
                              "key": "Interface",
                              "value": { "stringValue": "Ethernet" }
                            },
                            {
                              "key": "TotalServerCount",
                              "value": { "stringValue": "1" }
                            }
                          ]
                        }
                      }
                    },
                    {
                      "key": "event_id",
                      "value": {
                        "kvlistValue": {
                          "values": [
                            { "key": "guid", "value": { "stringValue": "" } },
                            { "key": "id", "value": { "stringValue": "1001" } }
                          ]
                        }
                      }
                    },
                    {
                      "key": "execution",
                      "value": {
                        "kvlistValue": {
                          "values": [
                            {
                              "key": "process_id",
                              "value": { "stringValue": "1596" }
                            },
                            {
                              "key": "thread_id",
                              "value": { "stringValue": "7544" }
                            }
                          ]
                        }
                      }
                    }
                  ]
                }
              },
              "traceId": "",
              "spanId": ""
            }
          ]
        }
      ]
    }
  ]
}

```

## TODO

- further testing

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dpaasman00 (commented):
Initial doc thoughts

dpaasman00 (changes requested):
First pass of windows code with some questions and comments

receiver/windowseventtracereceiver/receiver.go
```diff
@@ -0,0 +1,301 @@
...
+		lr.logger.Info("Enabled provider", zap.String("provider", providerConfig.Name))
+		successfulProviders++
+	}
+
+	if successfulProviders == 0 && totalProviders > 0 {
+		lr.logger.Error("Failed to enable any providers",
+			zap.String("session", lr.cfg.SessionName),
+			zap.Int("totalProviders", totalProviders))
+		return
+	}
```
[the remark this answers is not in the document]
schmikei: Decent idea and not too hard to implement 👍

receiver/windowseventtracereceiver/internal/etw/session.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+
+		return nil
+	}
+
+	providerNames := make([]string, 0, len(s.providerMap))
+	for name := range s.providerMap {
+		providerNames = append(providerNames, name)
+	}
+	return fmt.Errorf("provider %s not found, available providers: %v", nameOrGuid, strings.Join(providerNames, ", "))
```
dpaasman00, line 100: Pretty nit, but this feels more idiomatic for go if instead we check if `ok` is false and then do the error handling inside that block instead of the other way around

receiver/windowseventtracereceiver/internal/etw/session_controller.go
```diff
@@ -0,0 +1,266 @@
...
+				advapi32.EVENT_CONTROL_CODE_ENABLE_PROVIDER,
+				traceLevel,
+				provider.MatchAnyKeyword,
+				provider.MatchAllKeyword,
+				timeout,
+				&params,
+			)
+		}()
+
+		switch r1 {
```
dpaasman00, line 213: In the event of a panic, which of these cases would r1 be?

dpaasman00 (commented):
Another code review pass, will test next

receiver/windowseventtracereceiver/internal/etw/consumer.go
```diff
@@ -0,0 +1,294 @@
...
+			r, err := advapi32.CloseTrace(handle)
+			switch r {
+			case 0:
+				jitter++
+			// we've deleted it so return
+			case windows.ErrorInvalidHandle:
+				if jitter > 1 {
+					return
+				}
+				jitter++
```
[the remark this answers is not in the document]
schmikei: We only call it once now but all the logic should be in the single shutdown wait now

receiver/windowseventtracereceiver/internal/etw/consumer.go
dpaasman00, line 1: The consumer and subsequent functions in this file are setup for processing multiple sessions, however the receiver can only be created with a single session. I think this should be refactored a bit to reflect that.

dpaasman00 (approved):

PR #2276, opened 2025-03-25
fix: prevent agent from emitting metrics that have already been emitted

briangardner:
### Proposed Change

This change adds timestamps when throughput metrics are both collected and sent, and ensures that we are only sending metrics that have been collected after last transmission.  This prevents a scenario where an agent re-emits metrics after a disconnect/reconnect, causing out of order metrics to be received on the server.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


dpaasman00:
Do we need to keep track of a last reported sequence number for each processor? Right now it looks like we only track one which is the highest we see in a given collection/call to OTLPMeasurements. However each processor is going to have it's own sequence number. Just concerned if this could get out of sync

dpaasman00 (approved):

PR #2291, opened 2025-04-04
Feat: Adds Azure Log Analytics exporter (sentinel)

BominRahmani:
### Proposed Change

The following PR adds the Azure log analytics exporter.
##### Checklist
- [x] Changes are tested
- [x] CI has passed


exporter/azureloganalyticsexporter/marshal.go
```diff
@@ -0,0 +1,182 @@
...
+			return "", fmt.Errorf("marshal log record expression result: %w", err)
+		}
+		return string(bytes), nil
+	default:
+		return "", fmt.Errorf("unsupported log record expression result type: %T", lrExprResult)
+	}
+}
+
+// transformLogsToSentinelFormat transforms logs to Microsoft Sentinel format
+func (m *azureLogAnalyticsMarshaler) transformLogsToSentinelFormat(ctx context.Context, ld plog.Logs) ([]byte, error) {
```
dpaasman00, line 65: I think this function needs to be reconsidered a bit. I'm not sure I like how much we're marshaling and unmarshaling here, and with it happening on every single log record it will very quickly become expensive and slow. Is there a better way to convert the plog structure to what azure expects and only need to marshal once?

exporter/azureloganalyticsexporter/marshal.go
```diff
@@ -0,0 +1,182 @@
...
+					continue
+				}
+
+				// Try to parse rawLogStr as JSON first
+				var rawLogData map[string]interface{}
+				if err := json.Unmarshal([]byte(rawLogStr), &rawLogData); err != nil {
+					// If not valid JSON, use as raw string
+					azureLogAnalyticsLogs = append(azureLogAnalyticsLogs, map[string]interface{}{
+						"RawData": rawLogStr,
+					})
```
dpaasman00, line 169: Does this field need to be "RawData" or is this something that's dependent on how the user configures Azure? If it's the latter, I think this should be a config parameter for the exporter.

exporter/azureloganalyticsexporter/marshal.go
```diff
@@ -0,0 +1,182 @@
...
+				}
+
+				if rawLogStr == "" {
+					m.logger.Error("Error processing log record: raw log is empty")
+					continue
+				}
+
+				// Try to parse rawLogStr as JSON first
+				var rawLogData map[string]interface{}
+				if err := json.Unmarshal([]byte(rawLogStr), &rawLogData); err != nil {
```
dpaasman00, line 165: This section also feels a bit odd. Trying to understand how we want to send this data to azure.

exporter/azureloganalyticsexporter/factory.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Define type to replace metadata.Type temporarily
+var typeStr = component.MustNewType("azureloganalytics")
+
+// NewFactory creates a factory for Azure Log Analytics Exporter
+func NewFactory() exporter.Factory {
+	return exporter.NewFactory(
+		typeStr,
+		createDefaultConfig,
+		exporter.WithLogs(createLogsExporter, component.StabilityLevelAlpha),
```
dpaasman00, line 34: This should import the metadata file and use the value from that. Here's an example: https://github.com/observIQ/bindplane-otel-collector/blob/main/exporter/chronicleexporter/factory.go#L33

exporter/azureloganalyticsexporter/config.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// StreamName is the name of the custom log table in Log Analytics workspace
+	StreamName string `mapstructure:"stream_name"`
+
+	// RawLogField is the field name that will be used to send raw logs to the Log Analytics workspace.
+	RawLogField string `mapstructure:"raw_log_field"`
+
+	TimeoutSettings exporterhelper.TimeoutConfig `mapstructure:",squash"`
+
+	QueueSettings exporterhelper.QueueConfig `mapstructure:",squash"`
```
dpaasman00, line 46: The options here aren't actually used it seems like. I'd find a way to use them with the azure client or drop them.

exporter/azureloganalyticsexporter/factory.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+	"errors"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/exporter"
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+)
+
+// Define type to replace metadata.Type temporarily
+var typeStr = component.MustNewType("azureloganalytics")
```
dpaasman00, line 27: Same thing here, should be able to use the metadata file.

exporter/azureloganalyticsexporter/README.md
```diff
@@ -0,0 +1,196 @@
...
+[
+  {
+    "RawData": "<log data from field specified in raw_log_field>"
+  }
+]
+```
+
+## Minimum Agent Versions
+
+- Introduced: v1.73.1
```
dpaasman00, line 15:
```suggestion
- Introduced: v1.75.0
```

dpaasman00 (changes requested):
A few questions and comments. Can you also rebase this with main and make sure the otel versions are up to date. Also add an entry for this exporter to `docs/exporters.md`.

exporter/azureloganalyticsexporter/marshal.go
```diff
@@ -0,0 +1,182 @@
...
+			scopeLog := resourceLog.ScopeLogs().At(j)
+
+			for k := 0; k < scopeLog.LogRecords().Len(); k++ {
+				logRecord := scopeLog.LogRecords().At(k)
+
+				// Escape any unescaped newlines (if body is a string)
+				logBody := logRecord.Body()
+				if logBody.Type() == pcommon.ValueTypeStr {
+					logBody.SetStr(strings.ReplaceAll(logBody.AsString(), "\n", "\\n"))
+				}
```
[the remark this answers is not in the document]
BominRahmani: This was taken from the way we handle escaping any unescaped newlines for the raw log field in the chronicle forward exporter.

exporter/azureloganalyticsexporter/marshal.go
BominRahmani: I'll look into trying to optimize this.

exporter/azureloganalyticsexporter/marshal.go
BominRahmani: Log Analytics can have any field here, however whatever the field is MUST be matched with the field that the user has in the schema for the table they are sending it. Coincidentally, RawData is a field that Log Analytics provides as a default right when you create a table. I am ok with having this be configurable, but that might make it just a tad bit more user error prone.

dpaasman00 (changes requested):
A couple of smaller items to go with figuring out the marshaling/unmarshaling stuff. Otherwise looking good.

exporter/azureloganalyticsexporter/marshal.go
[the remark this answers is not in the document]
BominRahmani: No, however the goal of this specific portion is to copy how we handle raw_log_field 1-1 from the google chronicle forwarder.

exporter/azureloganalyticsexporter/marshal.go
BominRahmani: It will always start up as a default, however users can delete it. Its not like TimeGenerated column where that is nessesary and will be automatically added.

dpaasman00 (approved):

PR #2306, opened 2025-04-10
chore: Add generic install scripts

dpaasman00:
### Proposed Change

Adding some generic install scripts to the repo. These are not used by the Bindplane Agent at all. Instead these are being hosted here so that Bindplane get reference them when a user is installing an Agent of a custom AgentType. Specifically they are compatible with and meant to be used with distributions built using the OTel Distro Builder.

Note: There's a typo in the plist file generated by the builder that uses `supervisor-config.yaml` instead of `supervisor_config.yaml`. PR to fix it [here](https://github.com/observIQ/otel-distro-builder/pull/360.

### Testing
I've got a repo with some collector release artifacts built [here](https://github.com/dpaasman00/Test-Action). The latest release, `v0.0.46`, has artifacts built for `darwin-arm64`, `linux-arm64`, and `linux-amd64`.

To do a very basic install on a macos arm64 laptop, run: 
```
sudo ./scripts/generic-install/install_macos.sh -d dp-test -u https://github.com/dpaasman00/Test-Action
```

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (changes requested):
Please run shellcheck against the new scripts. There are a handful of decent findings.

One that sticks out is:

> In scripts/generic-install/install_unix.sh line 198:
            local tmp_file="/tmp/${DISTRIBUTION}_${version}.deb"
            ^------------^ SC3043 (warning): In POSIX sh, 'local' is undefined.

If we are using /bin/sh, we need to ensure the scripts are POSIX compliant.

jsirianni (approved):
Working great. Once comment around some wording.

@dpaasman00 I think it would be great to add inspec tests for these scripts, so we can easily test against a set of supported distributions and ensure things are deployed correctly (file permissions, service status, etc). This can come laterf and is not critical now, IMO.

PR #2318, opened 2025-04-15
fix(windows): Enable overwrite of binary files during MSI package upgrade

jsirianni:
### Proposed Change

This should allow the MSI package upgrade to overwrite binaries even if they were replaced out of band using Bindplane's upgrade mechanism.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni:
This change did not work as expected. Closing for now.

PR #2320, opened 2025-04-16
chore(release): Publish release artifacts to GCS

jsirianni:
### Proposed Change

Updates Goreleaser to publish artifacts to Google Cloud Storage in addition to Github. In the future, we will update install scripts to download from GCS instead of Github.

### Testing

This PR includes a manual github workflow for ensuring the GCS auth is working. I had it run in the PR with success.

image

Locally, I disabled the Github release / changelog steps and ran a full release against a local tag. I can see artifacts in the bucket, in the correct directory. I have since cleaned up the folder.

Screenshot From 2025-04-16 15-25-21

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/.github/workflows/manual_gcs_release_test.yml
+++ b/.github/workflows/manual_gcs_release_test.yml
@@ -1,7 +1,6 @@
 name: Test GCS Release Bucket
 on:
   workflow_dispatch:
-  pull_request:
 
 jobs:
   test_gcs_release_bucket:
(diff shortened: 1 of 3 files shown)


.github/workflows/manual_gcs_release_test.yml
```diff
@@ -0,0 +1,17 @@
+name: Test GCS Release Bucket
```
jsirianni, line 1: This workflow utilizes the same auth steps as the release workflow, giving us confidence that the release workflow will be successful when uploading to GCS.

dpaasman00 (approved):

PR #2352, opened 2025-04-21
feat: New awss3eventextension

djaglowski:
This extension is much like the `awss3eventreceiver` but is responsible only for downloading files. It it intended to be used alongside a `filelog` reciever.

Some code was extracted from the receiver and is now shared. I'll open some prerequisite PRs to make these changes more reviewable.


djaglowski:
The code is just about where I think it needs to be, but the PR is huge. Now that I know approximately what the end state looks like, I'll break out some incremental PRs to simplify this one. Plan is roughly:
1. Clean up a few issues noticed in the receiver.
2. Create `internal/aws` module and populate with corresponding packages from the receiver's internals.
3. Refactor `internal/aws` to support multiple Marshal/Unmarshal strategies (in preparation for supporting FDR format)
4. Add support to `internal/aws` for FDR events
5. Rebase this PR (just adds the extension)

extension/awss3eventextension/config_test.go
```diff
@@ -0,0 +1,195 @@
...
+				SQSQueueURL:          "https://sqs.us-east-1.amazonaws.com/123456789012/test-queue",
+				StandardPollInterval: 30 * time.Second,
+				MaxPollInterval:      60 * time.Second,
+				PollingBackoffFactor: 2,
+				VisibilityTimeout:    600 * time.Second,
+				Workers:              5,
+				EventFormat:          "fdr",
+				Directory:            "/tmp/s3event",
+			},
+			expectError: false,
```
mrsillydog, line 79: Hmm, it's a bit surprising to me that a config with an event format of `fdr` passes validation, considering the enumerated value described in the README and `eventFormatCrowdstrikeFDR` is `crowdstrike_fdr`.

extension/awss3eventextension/testdata/test_config.yaml
```diff
@@ -0,0 +1,23 @@
...
+    directory: /tmp/s3event
+
+  s3event/fdr:
+    sqs_queue_url: "https://sqs.us-east-1.amazonaws.com/123456789012/test-queue"
+    standard_poll_interval: 30s
+    max_poll_interval: 60s
+    polling_backoff_factor: 2
+    visibility_timeout: 600s
+    workers: 5
+    event_format: fdr
```
mrsillydog, line 22: Ditto, basically

extension/awss3eventextension/config.go
```diff
@@ -0,0 +1,94 @@
...
+package awss3eventextension // import "github.com/observiq/bindplane-otel-collector/extension/awss3eventextension"
+
+import (
+	"errors"
+	"time"
+)
+
+const (
+	eventFormatAWSS3          = "aws_s3"
+	eventFormatCrowdstrikeFDR = "crowdstrike_fdr"
```
mrsillydog, line 24: Also, I would like to see these constants used more ideally, but maybe we're not using them in tests for a reason?

mrsillydog (commented):
Will be interested in a quick walkthrough of how to test this manually tomorrow from you, but here are my initial, fairly minor thoughts

extension/awss3eventextension/config_test.go
djaglowski: Good catch. I forgot to add validation of this parameter.

extension/awss3eventextension/config.go
djaglowski: Two things going on here. First, test code is in a separate `_test` package in order to ensure that tests are interacting with via the same API that everyone outside the package must use. Second, the constants aren't exported because even if they were, the most realistic scenario is that users have to specify the correct magic values. In other words, I'd rather our tests validate that the constants are correct than test that using the constant works. The only reason to export would be for the sake of programatic usage, but I'm not aware of any use cases which require that currently.

mrsillydog (approved):
Going to approve under the assumption that you've done manual testing in order to get the release out

PR #2376, opened 2025-04-29
chore(release): Publish latest install scripts to GCS

jsirianni:
### Proposed Change

During the release, the latest install scripts will be copied to the GCS release bucket.

NOTE: This means any changes to the scripts will become live despite the release being in a "pre release" state. I spoke with @dpaasman00 about this. It is fine to start this way. The scripts are not changed often, and if they are changed, any relevant testing will be performed before the release.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.github/workflows/manual_gcs_release_test.yml
+++ b/.github/workflows/manual_gcs_release_test.yml
@@ -15,3 +15,5 @@ jobs:
         uses: google-github-actions/setup-gcloud@v2
       - name: Test Gsutil
         run: gsutil ls gs://bdot-release
+      - run: gsutil cp LICENSE gs://bdot-release/LICENSE
+      - run: gsutil rm gs://bdot-release/LICENSE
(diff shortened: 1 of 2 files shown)


.github/workflows/manual_gcs_release_test.yml
```diff
@@ -15,3 +15,5 @@ jobs:
         uses: google-github-actions/setup-gcloud@v2
       - name: Test Gsutil
         run: gsutil ls gs://bdot-release
+      - run: gsutil cp LICENSE gs://bdot-release/LICENSE
+      - run: gsutil rm gs://bdot-release/LICENSE
```
jsirianni, line 19: Because this is a public bucket, we need to perform a copy and rm to ensure we have the correct access. These are safe operations in the root of the bucket.

dpaasman00 (approved):

PR #2393, opened 2025-05-06
fix: Prevent panic in Chronicle HTTP exporter

kuiperda:
### Proposed Change

The exporter will panic (HTTP only) if the json creds are malformed (like putting a 'z' in for creds)

We were missing a nil check in `Shutdown`. Added a test to validate that we error instead of panicking on shutdown in this situation.

Side note, we should really `Validate()` the JSON creds configuration field immediately in our Google destinations instead of letting it error later. 

# Testing

Previous (with most recent agent):

https://github.com/user-attachments/assets/2ce8240e-a1c3-4abc-ba90-cd7c4b2494a9

This PR (replaced agent binary): 

https://github.com/user-attachments/assets/179f6740-670e-4f8d-ad42-9045d2ea73d9

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/exporter/chronicleexporter/http_exporter.go
+++ b/exporter/chronicleexporter/http_exporter.go
@@ -78,9 +78,11 @@ func (exp *httpExporter) Start(ctx context.Context, _ component.Host) error {
 
 func (exp *httpExporter) Shutdown(context.Context) error {
 	defer http.DefaultTransport.(*http.Transport).CloseIdleConnections()
-	t := exp.client.Transport.(*oauth2.Transport)
-	if t.Base != nil {
-		t.Base.(*http.Transport).CloseIdleConnections()
+	if exp.client != nil {
+		t := exp.client.Transport.(*oauth2.Transport)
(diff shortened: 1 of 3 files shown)


exporter/chronicleexporter/http_exporter.go
```diff
@@ -78,9 +78,11 @@ func (exp *httpExporter) Start(ctx context.Context, _ component.Host) error {
 
 func (exp *httpExporter) Shutdown(context.Context) error {
 	defer http.DefaultTransport.(*http.Transport).CloseIdleConnections()
-	t := exp.client.Transport.(*oauth2.Transport)
-	if t.Base != nil {
-		t.Base.(*http.Transport).CloseIdleConnections()
+	if exp.client != nil {
```
kuiperda, line 81:
```suggestion
	if exp.client == nil {
```
Try changing this to `== nil` and the test will panic instead.

mrsillydog (approved):

PR #2424, opened 2025-05-29
chore(linuxpackage): Move service files to package postinstall heredocs

jsirianni:
### Proposed Change

Requires https://github.com/observIQ/bindplane-otel-collector/pull/2418

This PR is in prep for configurable install directory. The service files will need to be dynamically generated during install / upgrade. Therefor, they cannot be part of the package manifest with static files.

A follow up PR will add logic for replacing `/opt/observiq-otel-collector` with some user configured value, using an environment variable. For now, the service files were copy pasted without changes**

**Any instances of `$`, particularly in the sysconfig / inid.d service files, have been escaped with `\$`.

### Archives

The Goreleaser archives were including service files. I removed all of them, not just the Linux service files. Lets chat and make sure this is okay.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


scripts/package/postinstall.sh
```diff
@@ -30,6 +30,358 @@ install() {
     rm -rf /usr/share/observiq-otel-collector
 }
 
+install_service() {
+  if command -v systemctl > /dev/null 2>&1; then
+    install_systemd_service
+  else
+    install_sysconfig_service
+    install_initd_service
```
jsirianni, line 38: TBD if this is correct. Currently, we deploy all three service files to all nix systems, and use the `service` command to maange sysconfig or init.d, depending on what the platform expects. I suspect this is okay, and follows that same pattern.

dpaasman00 (approved):
This was working well for fresh install and updating.

Should we delete the `/service/observiq-otel-collector` service file like you did with `/service/observiq-otel-collector`?

PR #2425, opened 2025-05-30
feat: add `save_log_record_original` option to plugins

justinianvoss22:
### Proposed Change
- This PR updates plugins used in Bindplane to have a boolean option, `save_log_record_original`. When set to true, it emits an additional attribute: `log.record.original`, that contains the raw log body. This will be used for measurement purposes. 
- Adds the following stanza operation that is used when the boolean option is true: 

`- id: save_log_record_original
          type: copy
          from: body
          to: attributes["log.record.original"]`

### Testing
- Tested several plugins in Bindplane. 
    - Steps: 
        - Apply updated plugin to collector (save the plugin in `/opt/observiq-otel-collector/plugins`)
        - Make a change to rollout the source again
        - Should see the raw body in the attribute `log.record.original`

##### Checklist
- [x] Changes are tested
- [x] CI has passed


plugins/active_directory_logs.yaml
```diff
@@ -38,13 +38,23 @@ template: |
       start_at: '{{ .start_at }}'
       attributes:
         log_type: "active_directory.general"
+      operators:
+        - id: save_log_record_original
+          type: copy
+          from: body
+          to: attributes["log.record.original"]
```
jsirianni, line 45: This change should be opt in, please wrap in a boolean option. (For all modified plugins)

When we update Bindplane sources, we can enable this option behind the scenes.

jsirianni (changes requested):
Please bump the plugin version at the top, for each modified plugin.

jsirianni (approved):

jsirianni (approved):

PR #2426, opened 2025-05-30
chore(linuxpackages): Merge legacy service files

jsirianni:
### Proposed Change

We are working on supporting configurable Linux packages. e.g. Configurable install directory. We intend to support config options in a user defined file at `/etc/default/observiq-otel-collector` (debian) and `/etc/sysconfig/observiq-otel-collector` (rhel).

This means we need to remove the sysconfig file from the Linux package. I moved all environment variables over to the init.d script. This is functionally the same.

In hindsite, we should have never had a sysconfig/observiq-otel-collector file because user modifications to it (such as changing the collector home variable) would not have worked as expected. Of course, we will support this exact functionality soon.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Dylan-M (changes requested):
Currently, local build fails.

```
    error=
    │ hook failed: hook failed: shell: 'make release-prep CURR_VERSION=1.77.2-SNAPSHOT-677992b1-SNAPSHOT-677992b1': exit status 2: make[1]: Entering directory '/home/dylanmyers/git/bindplane-otel-collector'
    │ ./buildscripts/download-dependencies.sh release_deps
    │ Retrieving java contrib at v1.28.0
    │ cp: cannot stat './service/sysconfig': No such file or directory
    │ make[1]: Leaving directory '/home/dylanmyers/git/bindplane-otel-collector'
    │ make[1]: *** [Makefile:229: release-prep] Error 1
make: *** [Makefile:236: release-test] Error 1
```

This is using: `make release-test` to generate all release artifacts.

Dylan-M (approved):
After several hours of testing on EL6 and SLES11SP4, everything seems to work here.

service/observiq-otel-collector
```diff
@@ -62,9 +62,15 @@ fi
 # with force-reload (in case signalling is not supported) are
 # considered a success.
 
-# Pull in sysconfig settings
-# shellcheck disable=SC1091
```
dpaasman00, line 66: Just curious why we no longer need the shellcheck or the ones from the `sysconfig/observiq-otel-collector`

.goreleaser.yml
```diff
@@ -115,9 +115,6 @@ archives:
       - src: release_deps/observiq-otel-collector
         dst: "install"
         strip_parent: true
-      - src: release_deps/sysconfig/observiq-otel-collector
-        dst: "install/sysconfig"
-        strip_parent: true
```
dpaasman00, line 120: Should we be doing a similar ghost file in the tar ball archives? I'm pretty sure the updater supports using tar balls for upgrades so wouldn't we run into the same issue you commented about for the nfpms packaging?

dpaasman00 (commented):
Couple questions otherwise looks good, going to test before approving.

.goreleaser.yml
jsirianni: Ghost allows us to remove a file from the package's manifest while also instructing the package manager to leave the file in place on the system if it exists. Otherwise, anything we copy over from /usr/share to /opt/observiq-otel-collector will get deleted on upgrade. Its pretty annoying, and this is how you get past that.

Going forward, we can add and remove new files without dealing with that, because the package manager is not tracking them. Similar to how it does not track the log file or manager.yaml, and will refuse to delete them on uninstall.

service/observiq-otel-collector
jsirianni: I am not sure if it was ever required. Shellcheck passes for me just fine.

dpaasman00 (approved):

PR #2428, opened 2025-06-02
deps: OTel v0.127.0

mrsillydog:
### Proposed Change

Updating the OTel dependency to v0.127.0/v1.33.0.

##### Checklist
- [X] Changes are tested
- [X] CI has passed


plugins/ingress_nginx_logs.yaml
```diff
@@ -27,10 +27,10 @@ template: |
           - '{{ $fp }}'
           {{ end }}
         start_at: {{ .start_at }}
+        {{ if .cluster_name }}
         resource:
-          {{ if .cluster_name }}
           k8s.cluster.name: {{ .cluster_name }}
-          {{ end }}
+        {{ end }}
```
mrsillydog, line 33: Not sure exactly what upstream change caused this, but there were issues unmarshalling what was basically an empty map - when cluster_name was blank, the unmarshal was failing, since the config basically looked like this:
```yaml
resource:
operators:
...
```

This led to a very unhelpful panic that I basically had to trial and error my way through. We'll definitely have to look through our BP resources to ensure that none of them have a similar issue.

dpaasman00: Is this going to cause an issue if someone using this plugin upgrades their collector?

mrsillydog: I believe they would get the fix for free - since we've updated the plugin YAML file, if they're using an old version of the plugin YAML file with the old collector, it'll work with the old OTel version, and the new file will work with the new OTel version. Upgrading the collector should also result in the fix

jsirianni: The change looks safe and I suspect this plugin has very low usage, or no users at all.

dpaasman00 (approved):

PR #2436, opened 2025-06-04
chore(linuxpackage): Rename observiq-otel-collector user to bdot

jsirianni:
### Proposed Change

This PR renames the Linux user `observiq-otel-collector` to `bdot`.
- Updated Goreleaser package manifest to use username `bdot`
- Updated install script to use username `bdot`
- Updated postinstall package script to use username `bdot`
  - This includes updating the systemd heredoc to use group name `bdot`
- Added a migration function to the preinstall script for migrating from `observiq-otel-collector` to `bdot`, using the `usermod` command.

##### Checklist
- [ ] Changes are tested
- [x] CI has passed

### DIFF
--- a/scripts/package/postinstall.sh
+++ b/scripts/package/postinstall.sh
@@ -24,10 +24,12 @@ set -e
 
 : "${BDOT_CONFIG_HOME:=/opt/observiq-otel-collector}"
 
+username="bdot"
+
 install() {
     mkdir -p "${BDOT_CONFIG_HOME}"
     chmod 0755 "${BDOT_CONFIG_HOME}"
-    chown observiq-otel-collector:observiq-otel-collector "${BDOT_CONFIG_HOME}"
+    chown "$username:$username" "${BDOT_CONFIG_HOME}"
     rm -f "${BDOT_CONFIG_HOME}/observiq-otel-collector" || true
     cp -r --preserve \
       /usr/share/observiq-otel-collector/stage/observiq-otel-collector/* \
@@ -64,7 +66,7 @@ StartLimitBurst=5
 [Service]
 Type=simple
 User=root
-Group=observiq-otel-collector
+Group=${username}
 Environment=PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
 Environment=OIQ_OTEL_COLLECTOR_HOME=${BDOT_CONFIG_HOME}
 Environment=OIQ_OTEL_COLLECTOR_STORAGE=${BDOT_CONFIG_HOME}/storage
@@ -432,7 +434,7 @@ manage_service() {
 finish_permissions() {
   # Goreleaser does not set plugin file permissions, so do them here
   # We also change the owner of the binary to observiq-otel-collector
-  chown -R observiq-otel-collector:observiq-otel-collector ${BDOT_CONFIG_HOME}/observiq-otel-collector ${BDOT_CONFIG_HOME}/plugins/*
+  chown -R "$username:$username" ${BDOT_CONFIG_HOME}/observiq-otel-collector ${BDOT_CONFIG_HOME}/plugins/*
   chmod 0640 ${BDOT_CONFIG_HOME}/plugins/*
 
   # Initialize the log file to ensure it is owned by observiq-otel-collector.
@@ -440,7 +442,7 @@ finish_permissions() {
   # the root user. By doing so, we allow the user to switch to observiq-otel-collector
   # user for 'non root' installs.
   touch ${BDOT_CONFIG_HOME}/log/collector.log
-  chown observiq-otel-collector:observiq-otel-collector ${BDOT_CONFIG_HOME}/log/collector.log
+  chown "$username:$username" ${BDOT_CONFIG_HOME}/log/collector.log
 }
 
 install
--- a/scripts/package/preinstall.sh
+++ b/scripts/package/preinstall.sh
@@ -16,18 +16,73 @@
 
 set -e
 
-username="observiq-otel-collector"
+username="bdot"
+legacy_username="observiq-otel-collector"
 
-if getent group "$username" > /dev/null 2>&1; then
-    echo "Group ${username} already exists."
-else
-    groupadd "$username"
-fi
+# Install creates the user and group for the collector
+# service. This function is idempotent and safe to call
+# multiple times.
+install() {
+    # Return early without output if the user and group already exist.
+    # This will help avoid confusion with the output in migrate_user().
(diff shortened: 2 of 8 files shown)


scripts/package/postinstall.sh
```diff
@@ -58,7 +60,7 @@ StartLimitBurst=5
 [Service]
 Type=simple
 User=root
-Group=observiq-otel-collector
+Group=bdot
```
Dylan-M: Should this be `${username}` so that it is consistent? (I'm not sure if that will work, but it **should**, since it will evaluate first, then write)

scripts/package/preinstall.sh
```diff
@@ -16,18 +16,70 @@
...
+        return
+    fi
+
+    if getent group "$username" > /dev/null 2>&1; then
+        echo "Skipping group migration: Group ${username} already exists."
+        return
+    fi
+
+    echo "Renaming group ${legacy_username} to ${username}"
+    groupmod -n "$username" "$legacy_username"
```
Dylan-M, line 84: This won't work for AIX (it doesn't have `groupmod`, despite having `usermod`) - I believe I have an alternative I can put in once I get to pulling these changes into the AIX branch. Just to remind me, could we put a TODO comment here?

Dylan-M (approved):
I left a few comments, but none are blockers.

scripts/package/postinstall.sh
jsirianni: Nice catch, it will work.

updater/internal/updater/templates.go
```diff
@@ -10,7 +10,7 @@ StartLimitBurst=5
 [Service]
 Type=simple
 User=root
-Group=observiq-otel-collector
+Group={{.Group}}
```
jsirianni, line 27: Offline agent updates does not deploy the new "bdot" user or handle any migration logic. Instead, we can read the current user from the service and keep running with it.

updater/internal/updater/updater.go
```diff
@@ -16,10 +16,12 @@
 package updater
```
kuiperda, line 16: I did some testing. 

1. Upgrading a system using the install script - This worked. 
- On a VM, I installed an agent in Stage and verified basic functionality. 
- I copied the `.deb` package from [here](https://github.com/observIQ/bindplane-otel-collector/actions/runs/15903539822) onto the VM
- I ran the install script with the --file flag pointed at the new `.deb` package
- The agent updated fine and its service file showed the group updated to `bdot`
- The user was unchanged since it installed as root.

updater/internal/updater/updater.go
```diff
@@ -16,10 +16,12 @@
...
```
kuiperda, line 17: 2. Upgrading a system using the updater - Some difficulties.
- Joe had an env with offline updates set up. There was a 1.79.2 agent already set up. I think everything was ready for me to just hit the upgrade button. This worked and did not change the group, which was still `observiq-otel-collector`
- I tried uninstalling and reinstalling with the v1.80 snapshot version and got this: 
image. 
- the `group` in the service file was `bdot` and I think the service was not running.
- I deleted the snapshot agent version and the UI started failing to get the install script. I tried syncing agent versions and got this:
image. 
- At first I ran the 1.79.2 install by just changing the version in the install script, but later I was able to get the install script again. In both cases, when I hit 'upgrade' in the UI, it hung saying 'upgrading' with no messages in `updater.log`
- I saw another agent update package, v1.80.0-alpha.1-SNAPSHOT-562fabf5 was added in the /root/agent directory
- So, the first attempt to update an agent worked but subsequent attempts to install a 1.79.2 agent and update it did not work, all were hanging with the `upgrading` chip and no logs. 
- I would sometimes install a 1.79.2 agent and see an upgrade button even after deleting other agent versions.

updater/internal/updater/updater.go
```diff
@@ -16,10 +16,12 @@
...
 
 import (
```
kuiperda, line 18: 3. Upgrading a system using the updater, after upgrading the system from step 1
- I updated a 1.79.2 agent to the 1.80-alpha.0-SNAPSHOT version using the package install with --file flag
- This worked but trying to use the update button to 1.80-alpha.1-SNAPSHOT hit the same behavior as above, where it hangs with no logs.

updater/internal/updater/updater.go
```diff
@@ -16,10 +16,12 @@
...
+	"bytes"
```
kuiperda, line 19: 4. Normal installs of the 1.80 versions seem fine, with Group=bdot in the service file.

kuiperda (changes requested):
I had some issues, possibly user error. I am happy to discuss further. 

Updating using the button in the UI was inconsistent for me.

updater/internal/updater/updater.go
Dylan-M: Isn't that sync URL broken there? `vlatest` instead of `latest`?

jsirianni: Ill look into the issue you identified where uninstalling a UI upgraded agent and re-installing with the new package fails. This is surprising to me.

The other issues seem like they could be unrelated, an issue with offline agent updates.

updater/internal/updater/updater.go
jsirianni: Okay, this invalidates my previous statement. ill look into it.

updater/internal/updater/updater.go
jsirianni: @kuiperda I was able to replicate this issue by doing the following
1. Install v1.79 using install script generated by Bindplane
2. Delete manager.yaml
3. Upgrade to snapshot build using the command from step one, appending `--file` flag

This cause the script to fail, because it runs the change group command while creating the manager config file. This fails because the script has `observiq-otel-collector` hardcoded. The new version of the script has `bdot` hardcoded. 

Everything works fine if I don't `rm` the manager.yaml before the upgrade. I would expect the user's flow to look like this
1. Install collector using script (v1.79 or whatever)
2. Upgrade collector using script, with manager.yaml preserved.

This flow works fine, as the manager creation logic is not hit.

In either case, I think the recommendation would be for the user to use the script bundled with the release. 

Side note: Your issues with the broken install command after deleting an agent version appear to be unrelated, I made a triage ticket for that bug.

updater/internal/updater/updater.go
jsirianni: I was unable to replicate this issue.

jsirianni:
Marking as ready for review again. Dan and I are meeting tomorrow to go over some final details.

NOTE: We will be merging to an epic branch instead of main. I would like to be able to merge these package changes into a branch that can be used to test the complete picture before we commit to main, and potentially get the changes included in a release prematurely.

PR #2437, opened 2025-06-04
feat: throughput measurements processor should support measurements from log record original

justinianvoss22:
### Proposed Change
- Add field to throughput measurements processor: `rawBytes`
    - This field is calculated using `log.record.original`, which contains the raw log body value. 
    - If `log.record.original` is NOT present, `rawBytes` will be calculated using the log body. 
    - `rawBytes` is an aggregation of the all of the log records' raw body values. 
    - `logSize` is still implemented. But `rawBytes` is ALSO used. 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/internal/measurements/throughput.go
+++ b/internal/measurements/throughput.go
@@ -43,7 +43,6 @@ type ThroughputMeasurements struct {
 	logSize, metricSize, traceSize      *int64Counter
 	logCount, datapointCount, spanCount *int64Counter
 	rawBytes                            *int64Counter
-	fullBytes                           *int64Counter
 	attributes                          attribute.Set
 	collectionSequenceNumber            atomic.Int64
 }
@@ -107,15 +106,6 @@ func NewThroughputMeasurements(mp metric.MeterProvider, processorID string, extr
 	}
 
 	rawBytes, err := meter.Int64Counter(
-		"bp.ingest.raw_bytes",
-		metric.WithDescription("Raw bytes ingested per log record"),
-		metric.WithUnit("By"),
-	)
-	if err != nil {
-		return nil, fmt.Errorf("create raw_bytes gauge: %w", err)
-	}
-
-	fullBytes, err := meter.Int64Counter(
 		"bp.ingest.full_bytes",
 		metric.WithDescription("Full bytes ingested per log record"),
 		metric.WithUnit("By"),
@@ -134,7 +124,6 @@ func NewThroughputMeasurements(mp metric.MeterProvider, processorID string, extr
 		traceSize:                newInt64Counter(traceSize, attrs),
 		spanCount:                newInt64Counter(spanCount, attrs),
 		rawBytes:                 newInt64Counter(rawBytes, attrs),
-		fullBytes:                newInt64Counter(fullBytes, attrs),
 		attributes:               attrs,
 		collectionSequenceNumber: atomic.Int64{},
 	}, nil
@@ -147,7 +136,7 @@ func (tm *ThroughputMeasurements) AddLogs(ctx context.Context, l plog.Logs) {
 	// Calculate total size using full log size
 	sizer := plog.ProtoMarshaler{}
 	totalSize := int64(sizer.LogsSize(l))
-	fullBytes := int64(0)
+	rawBytes := int64(0)
 
 	resourceLogs := l.ResourceLogs()
 	for i := 0; i < resourceLogs.Len(); i++ {
@@ -161,19 +150,15 @@ func (tm *ThroughputMeasurements) AddLogs(ctx context.Context, l plog.Logs) {
 
 				// Record raw bytes if log.record.original is present
 				if original, ok := logRecord.Attributes().Get("log.record.original"); ok {
-					rawBytes := int64(len(original.Str()))
-					tm.rawBytes.Add(ctx, rawBytes)
+					logRecordRawBytes := int64(len(original.Str()))
 
-					fullBytes += rawBytes
-
-					// Remove the attribute after measurement
-					logRecord.Attributes().Remove("log.record.original")
+					rawBytes += logRecordRawBytes
 				}
 			}
 		}
 	}
-	// Full bytes is the sum of all raw bytes
-	tm.fullBytes.Add(ctx, fullBytes)
+	// Raw bytes is the sum of all raw bytes
+	tm.rawBytes.Add(ctx, rawBytes)
 
 	tm.logSize.Add(ctx, totalSize)
 	tm.logCount.Add(ctx, int64(l.LogRecordCount()))
--- a/internal/measurements/throughput.go
+++ b/internal/measurements/throughput.go
@@ -42,6 +42,8 @@ type ThroughputMeasurementsRegistry interface {
 type ThroughputMeasurements struct {
 	logSize, metricSize, traceSize      *int64Counter
 	logCount, datapointCount, spanCount *int64Counter
+	rawBytes                            *int64Counter
+	fullBytes                           *int64Counter
 	attributes                          attribute.Set
 	collectionSequenceNumber            atomic.Int64
 }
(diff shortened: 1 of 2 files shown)


internal/measurements/throughput.go
```diff
@@ -113,17 +123,51 @@ func NewThroughputMeasurements(mp metric.MeterProvider, processorID string, extr
...
+		scopeLogs := resourceLog.ScopeLogs()
+		for j := 0; j < scopeLogs.Len(); j++ {
+			scopeLog := scopeLogs.At(j)
+			logRecords := scopeLog.LogRecords()
+			for k := 0; k < logRecords.Len(); k++ {
+				logRecord := logRecords.At(k)
+
+				// Record raw bytes if log.record.original is present
+				if original, ok := logRecord.Attributes().Get("log.record.original"); ok {
+					// fmt.Println("ORIGINAL", original.Str())
```
colelaven: left some print statements in here

internal/measurements/throughput.go
```diff
@@ -113,17 +123,49 @@ func NewThroughputMeasurements(mp metric.MeterProvider, processorID string, extr
...
 		traceSize:                newInt64Counter(traceSize, attrs),
 		spanCount:                newInt64Counter(spanCount, attrs),
+		rawBytes:                 newInt64Counter(rawBytes, attrs),
 		attributes:               attrs,
 		collectionSequenceNumber: atomic.Int64{},
 	}, nil
 }
 
 // AddLogs records throughput metrics for the provided logs.
 func (tm *ThroughputMeasurements) AddLogs(ctx context.Context, l plog.Logs) {
```
jsirianni: Can you write benchmarks for this function with and without the new logic? This would be a separate PR so we can compare main to this branch.

I am concerned about performance when handling a high volume of logs.

justinianvoss22: @jsirianni Regarding your comment above this one, [this PR](https://github.com/observIQ/bindplane-op-enterprise/pull/5333) introduces logic to handle the new and old agents in new and old Bindplane. The description explains what should happen for each case.

internal/measurements/throughput.go
```diff
@@ -113,17 +123,51 @@ func NewThroughputMeasurements(mp metric.MeterProvider, processorID string, extr
...
-func (tm *ThroughputMeasurements) AddLogs(ctx context.Context, l plog.Logs) {
-	sizer := plog.ProtoMarshaler{}
+func (tm *ThroughputMeasurements) AddLogs(ctx context.Context, l plog.Logs, measureRawBytes bool) {
 	tm.collectionSequenceNumber.Add(1)
 
-	tm.logSize.Add(ctx, int64(sizer.LogsSize(l)))
+	// Calculate total size using full log size
+	sizer := plog.ProtoMarshaler{}
+	totalSize := int64(sizer.LogsSize(l))
+	rawBytes := int64(0)
```
jsirianni: nit: This variable should be declared inside the `if measureRawBytes` condition, closer to where it is used.

processor/throughputmeasurementprocessor/README.md
```diff
@@ -1,7 +1,9 @@
 # Throughput Measurement Processor
+
 This processor samples OTLP payloads and measures the protobuf size as well as number of OTLP objects in that payload. These measurements are added to the following counter metrics that can be accessed via the collectors internal telemetry service. Units for each `data_size` counter are in Bytes.
 
 Counters:
+
```
colelaven, line 6: can you list the raw log counter here, as well as a quick explanation of what it is (specifically that it is only for logs)

internal/measurements/throughput.go
```diff
@@ -42,6 +42,7 @@ type ThroughputMeasurementsRegistry interface {
 type ThroughputMeasurements struct {
 	logSize, metricSize, traceSize      *int64Counter
 	logCount, datapointCount, spanCount *int64Counter
+	rawBytes                            *int64Counter
```
dpaasman00: I'd rename this to `logRawBytes` since as I understand it this is specific to logs

internal/measurements/throughput.go
```diff
@@ -113,17 +123,51 @@ func NewThroughputMeasurements(mp metric.MeterProvider, processorID string, extr
...
+			resourceLog := resourceLogs.At(i)
+			scopeLogs := resourceLog.ScopeLogs()
+			for j := 0; j < scopeLogs.Len(); j++ {
+				scopeLog := scopeLogs.At(j)
+				logRecords := scopeLog.LogRecords()
+				for k := 0; k < logRecords.Len(); k++ {
+					logRecord := logRecords.At(k)
+
+					// Record raw bytes if log.record.original is present
+					if original, ok := logRecord.Attributes().Get("log.record.original"); ok {
```
dpaasman00, line 153: What's going to be adding this attribute to the log records? Will it be the responsibility of the receiver or another processor earlier in the pipeline?

internal/measurements/throughput.go
dpaasman00: Correct me if I'm wrong though

internal/measurements/throughput.go
jsirianni: Yes, it is the responsibility of the receiver / Bindplane source. It is a semantic convention https://opentelemetry.io/docs/specs/semconv/registry/attributes/log/#log-record-attributes

We have Bindplane work in flight that implements this attribute where appropriate.

internal/measurements/throughput.go
colelaven: Agreed, I know youve already made some other PRs that would need changes Justin, but if its not too difficult itd be good to rename this measurement everywhere it is reference so that it is clearly logs only.

internal/measurements/throughput.go
justinianvoss22: This attribute will be added automatically by our sources and will be dropped in the destination of `Drop Raw Copy` is set to true. 
https://github.com/observIQ/bindplane-op-enterprise/commit/5917be147bc38e42e615b0f91af43d5356ca297e

dpaasman00: Perfect, that was going to be my next question

internal/measurements/throughput.go
dpaasman00: Yea I would prefer this, because whether or not we do so soon, I could see us needing to solve this problem for metrics and traces in the future (I'm assuming this is an issue for them as well).

justinianvoss22: Addressing your comment at the top:

As per the document, (https://docs.google.com/document/d/1AO_L7bpvObFCQeliw8_mK3dJSUu2TTE_Ap4ljVBpq-c/edit?tab=t.0)
Our approach was just for logs, because logs we were parsing with lots of additional plugins, and logs are the only one with the "raw" form that we could use, metrics and traces are already in a structured format. Metrics and traces could definitely be addressed in a later issue though.

justinianvoss22: True, I'll change it to `logRawBytes`

internal/measurements/throughput.go
```diff
@@ -42,6 +42,7 @@ type ThroughputMeasurementsRegistry interface {
 type ThroughputMeasurements struct {
 	logSize, metricSize, traceSize      *int64Counter
 	logCount, datapointCount, spanCount *int64Counter
+	logLogRawBytes                      *int64Counter
```
colelaven: did you mean to put logLog?

processor/throughputmeasurementprocessor/README.md
colelaven: thank you 🙌

internal/measurements/throughput.go
justinianvoss22: Nope... find and replace did it again for capitalized ones. whoops.

jsirianni (approved):

PR #2441, opened 2025-06-05
feat: create basic webhook exporter for logs

briangardner:
### Proposed Change

Adds a webhook exporter that supports logs signal

##### Checklist
- [x] Changes are tested
- [x] CI has passed


exporter/webhookexporter/README.md
```diff
@@ -0,0 +1,46 @@
...
+   - The configured Content-Type header is applied
+   - Any additional headers specified in the configuration are included
+   - The request body contains the data in JSON format
+
+4. **Error Handling**:
+   - Failed requests are retried according to the sending queue configuration
+   - Non-2xx HTTP responses are treated as errors
+   - Connection timeouts are handled according to the configured timeout settings
+
+## Configuration
```
jsirianni, line 35: We should include the timeout configuration option(s) here. What is the default timeout is?

exporter/webhookexporter/config.go
```diff
@@ -0,0 +1,139 @@
...
+	"go.opentelemetry.io/collector/config/configretry"
+	"go.opentelemetry.io/collector/config/configtls"
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+)
+
+// Endpoint represents a validated HTTP/HTTPS endpoint URL
+type Endpoint string
+
+// UnmarshalText implements the encoding.TextUnmarshaler interface
+func (e *Endpoint) unmarshalText(text []byte) error {
```
jsirianni: There are several failure cases that could be caught by url.Parse, does it make sense to run the string through it for validation?

We should still check the scheme for http or https :+1:

exporter/webhookexporter/config.go
```diff
@@ -0,0 +1,139 @@
...
+	"fmt"
+	"strings"
+
+	"go.opentelemetry.io/collector/config/configretry"
+	"go.opentelemetry.io/collector/config/configtls"
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+)
+
+// Endpoint represents a validated HTTP/HTTPS endpoint URL
+type Endpoint string
```
jsirianni: Could this be url.URL type?

exporter/webhookexporter/config.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Headers contains additional HTTP headers to include in the webhook requests
+	Headers map[string]string `mapstructure:"headers"`
+
+	// ContentType specifies the Content-Type header for the webhook requests
+	// This field is required
+	ContentType string `mapstructure:"content_type"`
+
+	// TLSSetting struct exposes TLS client configuration.
+	TLSSetting *configtls.ClientConfig `mapstructure:"tls"`
```
jsirianni: nit: I saw a PR upstream renaming `TLSSetting` to `TLS`. We should probably do the same here. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40455#event-17959362855

jsirianni (changes requested):
It would be nice if we had an integration test that spun up a simple HTTP server w/ TLS. This would allow us to test against an API similar to what we expect to see in a user environment. This could be a follow up PR.

exporter/webhookexporter/config.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+// Config defines the configuration for the webhookexporter
+type Config struct {
+	LogsConfig *SignalConfig `mapstructure:"logs,omitempty"`
+}
+
+// SignalConfig defines the configuration for a single signal type (logs, metrics, traces)
+type SignalConfig struct {
```
dpaasman00, line 57: I think we'll want to add the `ConfigAuth` struct in this config as well. At a high level it would allow the user to configure an authenticator extension component they can reference in this config and then the exporter could use that to authenticate with. More info [here](https://github.com/open-telemetry/opentelemetry-collector/tree/main/config/configauth).

exporter/webhookexporter/factory.go
```diff
@@ -0,0 +1,94 @@
...
+
+func createDefaultConfig() component.Config {
+	return &Config{
+		LogsConfig: &SignalConfig{
+			Endpoint:    Endpoint("https://localhost"),
+			Verb:        POST,
+			ContentType: "application/json",
+		},
+		//MetricsConfig: nil,
+		//TracesConfig:  nil,
```
dpaasman00: Errant comments

exporter/webhookexporter/exporter_logs.go
```diff
@@ -0,0 +1,183 @@
...
+	le.logger.Info("shutting down webhook logs exporter")
+	le.client.CloseIdleConnections()
+	le.logger.Info("webhook logs exporter shutdown complete")
+	return nil
+}
+
+func (le *logsExporter) logsDataPusher(ctx context.Context, ld plog.Logs) error {
+	if le.cfg.LogsConfig == nil {
+		return fmt.Errorf("logs config is required")
+	}
```
dpaasman00: Not required because of the check in config.Validate

exporter/webhookexporter/config.go
```diff
@@ -0,0 +1,139 @@
...
+		if err := c.LogsConfig.Validate(); err != nil {
+			return fmt.Errorf("logs config validation failed: %w", err)
+		}
+	}
+
+	if c.LogsConfig == nil {
+		return fmt.Errorf("logs config is required")
+	}
+	return nil
+}
```
dpaasman00:
```suggestion
func (c *Config) Validate() error {
	if c.LogsConfig != nil {
		if err := c.LogsConfig.Validate(); err != nil {
			return fmt.Errorf("logs config validation failed: %w", err)
		}
	} else {
	        return fmt.Errorf("logs config is required")
	}

	return nil
}
```

exporter/webhookexporter/exporter_logs.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func newLogsExporter(
+	_ context.Context,
+	cfg *Config,
+	params exporter.Settings,
+) (*logsExporter, error) {
+	if cfg.LogsConfig == nil {
+		return nil, fmt.Errorf("logs config is required")
+	}
```
dpaasman00: Not required because of the check in config.Validate

exporter/webhookexporter/exporter_logs.go
```diff
@@ -0,0 +1,183 @@
...
+		var parsedLog any
+		if err := json.Unmarshal([]byte(logStr), &parsedLog); err != nil {
+			// If the log isn't valid JSON, keep it as a string
+			logs = append(logs, logStr)
+			continue
+		}
+		logs = append(logs, parsedLog)
+	}
+	return logs
+}
```
dpaasman00, line 154: The `plog` package has some helpful tools that I think we should use instead here for extracting and marshaling our logs. Specifically, if you look at our AzureLogAnalytics(Sentinel) exporter [here](https://github.com/observIQ/bindplane-otel-collector/blob/main/exporter/azureloganalyticsexporter/marshal.go#L87-L95) there's a good example of how to do JSON marshaling using the functions on the OTel data type.

exporter/webhookexporter/config.go
```diff
@@ -0,0 +1,139 @@
...
+	// Verb specifies the HTTP method to use for the webhook requests
+	// Must be one of: POST, PATCH, PUT
+	Verb HTTPVerb `mapstructure:"verb"`
+
+	// Headers contains additional HTTP headers to include in the webhook requests
+	Headers map[string]string `mapstructure:"headers"`
+
+	// ContentType specifies the Content-Type header for the webhook requests
+	// This field is required
+	ContentType string `mapstructure:"content_type"`
```
dpaasman00, line 72: The exporter only exports logs as JSON, right? Should we allow specifying this right now? Would it be worthwhile building this out to support more content types?

exporter/webhookexporter/config.go
```diff
@@ -0,0 +1,139 @@
...
+// SignalConfig defines the configuration for a single signal type (logs, metrics, traces)
+type SignalConfig struct {
+	// TimeoutConfig contains settings for request timeouts
+	TimeoutConfig exporterhelper.TimeoutConfig `mapstructure:",squash"`
+
+	// QueueBatchConfig contains settings for the sending queue and batching
+	QueueBatchConfig exporterhelper.QueueBatchConfig `mapstructure:"sending_queue"`
+
+	// BackOffConfig contains settings for retry behavior on failures
+	BackOffConfig configretry.BackOffConfig `mapstructure:"retry_on_failure"`
```
dpaasman00, line 64: This config option isn't used right now

exporter/webhookexporter/config.go
dpaasman00: This config option isn't used right now either

dpaasman00 (commented):
First pass has solid bones. Some thoughts

exporter/webhookexporter/config.go
briangardner: Renamed

exporter/webhookexporter/config.go
briangardner: Updated to eliminate the Endpoint type, and use url.URL in the config

exporter/webhookexporter/config.go
briangardner: This Endpoint type is eliminated.  Moved http/https check to config validation.

exporter/webhookexporter/config.go
briangardner: Yes, we do need this option now.  Salesforce uses a custom json header value (`Content-Type: application/vnd.kafka.json.v1+json`), and we also support passing along raw strings if its not JSON.

exporter/webhookexporter/config.go
briangardner: I was told these were somehow automatically used by the underlying collector

exporter/webhookexporter/config.go
briangardner: I was told these were somehow automatically used by the underlying collector

exporter/webhookexporter/exporter_logs.go
briangardner: Removed duplicate logic

exporter/webhookexporter/factory.go
briangardner: Removed

exporter/webhookexporter/config.go
briangardner: I've added configauth.

exporter/webhookexporter/config.go
dpaasman00: Hmm, I don't think so, normally you have to pass it to a http client for it to get used. OTel also has a custom http client pkg that can be used by the collector. That might be useful here.

exporter/webhookexporter/README.md
briangardner: We are now using the built in Timeout from `confighttp.ClientConfig`, which defaults to 0

exporter/webhookexporter/exporter_logs.go
```diff
@@ -0,0 +1,184 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/exporter"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.uber.org/zap"
+)
+
+type logsExporter struct {
+	cfg      *Config
```
dpaasman00: I'd consider changing this type from `Config` to `SignalConfig`. Since this is the logs exporter it should only need access to the logs config

exporter/webhookexporter/exporter_logs.go
```diff
@@ -0,0 +1,184 @@
...
+func (le *logsExporter) Capabilities() consumer.Capabilities {
+	return consumer.Capabilities{MutatesData: false}
+}
+
+func (le *logsExporter) start(_ context.Context, host component.Host) error {
+	le.logger.Info("starting webhook logs exporter")
+	client, err := le.cfg.LogsConfig.ClientConfig.ToClient(context.Background(), host, le.settings)
+	if err != nil {
+		return fmt.Errorf("failed to create http client: %w", err)
+	}
```
dpaasman00: I'd consider move this into the `newLogsExporter` so you don't need to worry about saving the telemetry settings object on the logs struct.

exporter/webhookexporter/exporter_logs.go
dpaasman00: Any thoughts on this?

exporter/webhookexporter/exporter_logs.go
briangardner: Putting it in the `start` method is how the `alertmanagerexporter` in `opentelmetry-collector-contrib` does it.  I think that's the easiest place to get the `host`

exporter/webhookexporter/exporter_logs.go
briangardner: Good call. Done

exporter/webhookexporter/exporter_logs.go
briangardner: I looked at it, and I feel like what we're doing is more closely related to what happens inside `transformRawLogsToAzureLogAnalyticsFormat` from that same file you referenced.  And in there, they process the entries very similar to what we do

exporter/webhookexporter/exporter_logs.go
dpaasman00: Yep, great point, totally missed the host being used here.

dpaasman00 (approved):

exporter/webhookexporter/README.md
jsirianni: Can we set a default, such as 30 seconds? Go's HTTP client without timeouts can cause some nasty behavior when the user environment is not working well.

jsirianni (changes requested):
@briangardner @Caleb-Hurshman one comment about timeouts and ill approve.

exporter/webhookexporter/README.md
briangardner: I've set a default timeout of 30s in `factory.go` in the `createDefaultConfig()` func.

PR #2453, opened 2025-06-11
feat: Webhook exporter default user agent (PIPE-176)

Caleb-Hurshman:
### Proposed Change

Adds a default "User-Agent" header to the webhook exporter requests. The default header will be "bindplane-otel-collector/<version>".
Example:
```
Headers: map[Accept-Encoding:[gzip] Content-Length:[92782] Content-Type:[application/json] My-Header:[hello] User-Agent:[bindplane-otel-collector/v1.78.0-SNAPSHOT-fb0c0c5f]]
```

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


jsirianni (approved):

exporter/webhookexporter/factory.go
```diff
@@ -19,37 +19,49 @@ import (
...
+func createDefaultConfig(collectorVersion string) func() component.Config {
+	userAgent := fmt.Sprintf("%s/%s", defaultUserAgent, collectorVersion)
+	return func() component.Config {
+		return &Config{
+			LogsConfig: &SignalConfig{
+				ClientConfig: confighttp.ClientConfig{
+					Endpoint: "https://localhost",
+					Timeout:  30 * time.Second,
+					Headers: map[string]configopaque.String{
+						"User-Agent": configopaque.String(userAgent),
```
briangardner, line 56: What happens in these scenarios:
1. User saves a configuration with a different User Agent header set
2. User saves a configuration, and then a updates collector versions.  Does the old value persist, or do we get the new version?

Separate question: Does this default value render in Bindplane UI when creating a webhook destination?

Caleb-Hurshman: 1. The new header replaces the default
2. Not sure, it depends on whether the factory gets called again anytime during/after an upgrade. I'd guess we'd get a new value.
3. No, we would have to make a PR to BPOP to change the default behavior in the UI. The defaults in the collector and in the UI are completely separate.

briangardner (approved):

PR #2454, opened 2025-06-12
feat: Webhook exporter parse strings to JSON (PIPE-177)

Caleb-Hurshman:
### Proposed Change

Parse incoming strings to JSON when `content_type` is set to `application/json`.
Update testing and documentation for this behavior.

Example filelog receiver input:
```
"my message 1"
"my message 2"
"my message 3"
```

Example Webhook exporter test server output:
```
Headers: map[Accept-Encoding:[gzip] Content-Length:[82] Content-Type:[application/json] User-Agent:[bindplane-otel-collector/v2.2.3]]
Body:
[
  {
    "message": "my message 1"
  },
  {
    "message": "my message 2"
  },
  {
    "message": "my message 3"
  }
]
```

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


briangardner:
I'm not sure I agree with this "feature" generally speaking.  We've already seen that Salesforce will use a custom header for JSON based on their examples.  If the user configures the header to say its JSON, but doesn't pass JSON, that is a configuration issue, not logic we should be building into the collector.

image

Caleb-Hurshman:
> I'm not sure I agree with this "feature" generally speaking. We've already seen that Salesforce will use a custom header for JSON based on their examples. If the user configures the header to say its JSON, but doesn't pass JSON, that is a configuration issue, not logic we should be building into the collector.
> 
> image

@jsirianni thoughts?

PR #2463, opened 2025-06-16
feat: Add log_type validation to https (DataPlane) SecOps exporter

shazlehu:
### Proposed Change

This change adds a new configuration parameter, `validate_log_types`. When set to true, the exporter will make an API call to SecOps on startup to create a list of valid log types. Additionally, if the default log type isn't valid or is empty, it will fall back to the `CATCH_ALL` log type.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


briangardner:
PR needs to be rebased

briangardner (approved):

PR #2467, opened 2025-06-17
feat: Webhook Exporter event per request limit (PIPE-293)

Caleb-Hurshman:
### Proposed Change

We've been mistakenly using the queue size as a limit for the number of logs per request. Create a new config param to configure this value instead. 
Update testing and documentation for this change

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/webhookexporter/README.md
+++ b/exporter/webhookexporter/README.md
@@ -22,7 +22,7 @@ The webhook exporter sends data to a configured HTTP endpoint. Here's how it pro
 
    - Data is extracted from the OpenTelemetry data model
    - Each entry's body is parsed as JSON if possible, otherwise kept as a string
-   - Entries are organized into batches based on the configured queue size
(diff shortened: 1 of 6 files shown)


Caleb-Hurshman:
> I am wondering if batch processor would be more appropriate than a configuration parameter. I am concerned with failed and retried requests. What happens if a large batch is split into multiple requests, and one of those requests fails? Is the entire batch retried?
> 
> The batch processor would ensure batches up to some configured size reach the exporter. We would need to do some testing, but I am fairly confident a batch max send size of "100" would split a large batch into smaller batches of 100.
> 
> Also, we should consider that there is work upstream for handling this kind of thing in the sending queue. See the OTLP exporter here https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md.
> 
> ```yaml
> sending_queue:
>   batch:
>     `min_size`: 1
>    `max_size`: 100
> ```
> 
> The new batching logic (in the queue) would handle this kind of thing better. The downside is it does not support **persistent** queue yet.

@jsirianni I guess relying on batching in any form (manual or processor) does complicate backpressure behavior, right? Not sure what's more important for the use case we're trying to solve.

jsirianni (approved):

PR #2509, opened 2025-07-02
feat: Offset storage

dpaasman00:
### Proposed Change

- Refactor existing storage code for rehydration checkpoints into `internal/storageclient` to accept generic data
- Refactor existing references to rehydration checkpoints to work with changes
- Add offset storage to the awss3event receiver

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/awss3eventreceiver/internal/worker/worker.go
```diff
@@ -218,6 +233,11 @@ func (w *Worker) consumeLogsFromS3Object(ctx context.Context, record events.S3Ev
 			batchesConsumedCount++
 			logger.Debug("Reached max logs for single batch, starting new batch", zap.Int("batches_consumed_count", batchesConsumedCount))
 
+			// Save the offset to storage
+			if err := w.offsetStorage.SaveStorageData(ctx, offsetStorageKey, NewOffset(parser.Offset())); err != nil {
+				logger.Error("Failed to save offset", zap.Error(err), zap.String("offset_storage_key", offsetStorageKey), zap.Int64("offset", parser.Offset()))
+			}
```
dpaasman00, line 241: Not sure if there's a different way we should handle this if there's an error saving the offset?

receiver/awss3eventreceiver/receiver.go
```diff
@@ -77,12 +81,25 @@ func newLogsReceiver(id component.ID, tel component.TelemetrySettings, cfg *Conf
...
 	}, nil
 }
 
-func (r *logsReceiver) Start(_ context.Context, _ component.Host) error {
+func (r *logsReceiver) Start(_ context.Context, host component.Host) error {
 	// Context passed to Start is not long running, so we can use a background context
 	ctx := context.Background()
+
+	// Create offset storage
+	// TODO: may need a const for the componentID to get proper syncing between collector instances talking to the same redis instance
```
dpaasman00, line 93: Highlighting this

receiver/awss3eventreceiver/internal/worker/worker.go
andykellr: I think this is fine. If we can't save the offset, it's not fatal, but it will impact retries and potentially result in duplicate logs.

PR #2510, opened 2025-07-03
chore: point `sqlqueryreceiver` to upstream instead of specific commit

justinianvoss22:
### Proposed Change
- As a part of this ticket: https://linear.app/bindplane/issue/PIPE-320/newest-upstream-sql-query-receiver-changes-need-to-be-implemented-in
- We need to point to the upstream changes instead of the specific commit. We were using the specific commit temporarily until we got the upstream changes in. 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1039,8 +1039,6 @@ replace github.com/observiq/bindplane-otel-collector/receiver/telemetrygenerator
 
 replace github.com/observiq/bindplane-otel-collector/receiver/windowseventtracereceiver => ./receiver/windowseventtracereceiver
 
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -1039,6 +1039,8 @@ replace github.com/observiq/bindplane-otel-collector/receiver/telemetrygenerator
 
 replace github.com/observiq/bindplane-otel-collector/receiver/windowseventtracereceiver => ./receiver/windowseventtracereceiver
 
+replace github.com/observiq/bindplane-otel-collector/receiver/sqlqueryreceiver => ./receiver/sqlqueryreceiver
+
```
dpaasman00: This will need to be deleted and then re-run make tidy

PR #2516, opened 2025-07-08
chore: Improve s3 receiver metrics

shazlehu:
### Proposed Change

Adds the following metrics to the awsS3eventreceiver:

### otelcol_s3event.batch_size

The number of logs in a batch, using these bucket boundaries: [1, 100, 250, 500, 750, 1000, 2500, 5000, 10000, 20000, 30000, 40000, 50000]

| Unit | Metric Type | Value Type |
| ---- | ----------- | ---------- |
| {logs} | Histogram | Int |

### otelcol_s3event.failures

The number of failures encountered while processing S3 objects
    

| Unit | Metric Type | Value Type | Monotonic |
| ---- | ----------- | ---------- | --------- |
| {failures} | Sum | Int | true |

### otelcol_s3event.objects_handled

The number of S3 objects processed by the receiver

| Unit | Metric Type | Value Type | Monotonic |
| ---- | ----------- | ---------- | --------- |
| {objects} | Sum | Int | true |

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/awss3eventreceiver/documentation.md
```diff
@@ -8,3 +8,31 @@
 | ---- | ----------- | ------ | ------- |
 | aws.s3.bucket | The name of the S3 bucket containing the processed object | Any Str | true |
 | aws.s3.key | The key of the S3 object that was processed | Any Str | true |
+
+## Internal Telemetry
```
briangardner, line 12: Should these metrics have a `bindplane` prefix, as this is our custom S3 receiver, and not the upstream?

receiver/awss3eventreceiver/internal/worker/worker_test.go
briangardner, line 1: We should have tests that ensure the metrics increment as expected

receiver/awss3eventreceiver/receiver.go
```diff
@@ -202,33 +206,29 @@ func (r *logsReceiver) receiveMessages(ctx context.Context) int {
...
 
-	resp, err := r.sqsClient.ReceiveMessage(ctx, params)
-	if err != nil {
-		r.telemetry.Logger.Error("receive messages", zap.Error(err))
-		return numMessages
-	}
-
 	// loop until we get no messages
-	for len(resp.Messages) > 0 {
+	for {
```
briangardner, line 208: Why did we change this behavior with the for loop?  Does this impact how the base collector loops?

briangardner (changes requested):

receiver/awss3eventreceiver/receiver.go
shazlehu: I was trying to make it easier to reason about, it should behave the same. Previously we had this block:
```go
        resp, err := r.sqsClient.ReceiveMessage(ctx, params)
 	if err != nil {
 		r.telemetry.Logger.Error("receive messages", zap.Error(err))
 		return numMessages
 	}
```
before the loop and at the end of the loop. The change means we just have this block once at the top of the loop, and exit when we have no messages.

PR #2517, opened 2025-07-09
feat: Implement s3 regex filtering on bucket name and object key

shazlehu:
### Proposed Change

Adds regex filtering on bucket name and object key to the `awsS3eventreceiver`. These are filters, i.e. when the bucket name or object key matches, we exclude them from processing.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/awss3eventreceiver/config.go
```diff
@@ -134,5 +142,19 @@ func (c *Config) Validate() error {
 		return err
 	}
 
+	if c.BucketNameFilter != "" {
```
briangardner: Do we need to trim whitespace?

receiver/awss3eventreceiver/internal/worker/worker.go
```diff
@@ -55,10 +56,12 @@ type Worker struct {
 	visibilityExtensionInterval time.Duration
 	maxVisibilityWindow         time.Duration
 	metrics                     *metadata.TelemetryBuilder
+	bucketNameFilter            *regexp.Regexp
```
briangardner, line 59: I'm wondering if we should move this to an options pattern.  The New() method is getting a large number of parameters

receiver/awss3eventreceiver/receiver.go
```diff
@@ -72,6 +73,24 @@ func newLogsReceiver(id component.ID, tel component.TelemetrySettings, cfg *Conf
...
+			return nil, fmt.Errorf("failed to compile regex filter: %w", err)
+		}
+	}
+
+	var objectKeyFilter *regexp.Regexp
+	if cfg.ObjectKeyFilter != "" {
+		objectKeyFilter, err = regexp.Compile(cfg.ObjectKeyFilter)
+		if err != nil {
+			// config validation should have already caught this
+			return nil, fmt.Errorf("failed to compile regex filter: %w", err)
```
briangardner: Should specify that its the object key filter in the error message for clarity

briangardner (changes requested):

briangardner (approved):

PR #2530, opened 2025-07-16
feat(updater): Support user configured systemd working directory / collector home

jsirianni:
### Proposed Change

We support a user configured install dir. See `scripts/package/postinstall.sh`. This is a follow up change to support the user configured directory when using the updater.

Updated the updater's systemd service file template to include a `.InstallDir` variable. The value used is extracted from the installed systemd service file. This follows the pattern established when detecting the runtime `group`. 

### Testing

Working great when the install dir is not configured.

I tested an install dir override by setting `BDOT_CONFIG_HOME=/opt/collector` and then building this branch with two tags:
```
observiq-otel-collector-v1.81.0-alpha.3
observiq-otel-collector-v1.81.0-alpha.4
```

The builds were identical, and allowed me to test upgrading with the new logic. We cannot upgrade from the latest release because it will fail to execute the updater binary, due to the `/opt/observiq-otel-collector` path being hardcoded in `updater/internal/path/`. This is okay because we will not support linux package customization for existing installs, only **new** installs.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/updater/cmd/updater/main.go
+++ b/updater/cmd/updater/main.go
@@ -40,7 +40,7 @@ func main() {
 
 	// We can't create the zap logger yet, because we don't know the install dir, which is needed
 	// to create the logger. So we pass a Nop logger here.
-	installDir, err := path.InstallDir(zap.NewNop())
+	installDir, err := path.InstallDir(zap.NewNop(), path.DefaultConfigOverrides)
 	if err != nil {
 		// Can't use "fail" here since we don't know the install directory
 		log.Fatalf("Failed to determine install directory: %s", err)
--- a/updater/internal/path/path_linux.go
+++ b/updater/internal/path/path_linux.go
@@ -43,7 +43,7 @@ var DefaultConfigOverrides = []string{
 }
 
 const (
-	configOverrideEnvVar = "BDOT_CONFIG_HOME"
+	configHomeOverrideKey = "BDOT_CONFIG_HOME"
 )
 
 // InstallDir returns the filepath to the install directory
@@ -104,7 +104,7 @@ func customInstallDir(logger *zap.Logger, config string) (string, error) {
 			continue
 		}
 
-		if strings.HasPrefix(line, configOverrideEnvVar+"=") {
+		if strings.HasPrefix(line, configHomeOverrideKey+"=") {
 			parts := strings.SplitN(line, "=", 2)
 			if len(parts) == 2 {
 				value := strings.TrimSpace(parts[1])
--- a/updater/internal/path/path_linux.go
+++ b/updater/internal/path/path_linux.go
@@ -15,14 +15,18 @@
 package path
 
 import (
+	"bufio"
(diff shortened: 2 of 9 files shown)


updater/cmd/updater/main.go
```diff
@@ -40,7 +40,7 @@ func main() {
 
 	// We can't create the zap logger yet, because we don't know the install dir, which is needed
 	// to create the logger. So we pass a Nop logger here.
-	installDir, err := path.InstallDir(zap.NewNop())
+	installDir, err := path.InstallDir(zap.NewNop(), path.DefaultConfigOverrides)
```
jsirianni, line 43: path.InstallDir takes a list of package overrides. On Linux this will be `/etc/default/observiq-otel-collector` and `/etc/sysconfig/observiq-otel-collector`. On Windows and Darwin, this option is ignored.

updater/internal/path/path_linux.go
```diff
@@ -15,24 +15,81 @@
...
+			if line == "" || strings.HasPrefix(line, "#") {
+				continue
+			}
+
+			if strings.HasPrefix(line, configOverrideEnvVar+"=") {
+				parts := strings.SplitN(line, "=", 2)
+				if len(parts) == 2 {
+					value := strings.TrimSpace(parts[1])
+					if value != "" {
+						return value, nil
```
jsirianni: InstallDir returns the value for `BDOT_CONFIG_HOME` right away. We do not expect users to configure both `/etc/default/observiq-otel-collector` (debian) and `/etc/sysconfig/observiq-otel-collector` (rhel). This function will read the one that exists and return the value if it is found.

updater/internal/path/path_linux.go
```diff
@@ -15,24 +15,81 @@
...
+				}
+			}
+		}
+
+		if err := scanner.Err(); err != nil {
+			return "", fmt.Errorf("error scanning config file %s: %w", config, err)
+		}
+	}
+
+	return DeafultLinuxInstallDir, nil
```
jsirianni: If the files do not exist, continue using `/opt/observiq-otel-collector`.

updater/internal/path/path_linux.go
```diff
@@ -15,24 +15,112 @@
...
+}
+
+const (
+	configOverrideEnvVar = "BDOT_CONFIG_HOME"
+)
+
 // InstallDir returns the filepath to the install directory
-func InstallDir(_ *zap.Logger) (string, error) {
-	return LinuxInstallDir, nil
+func InstallDir(logger *zap.Logger, configOverrides []string) (string, error) {
```
jsirianni, line 50: This function is used by the collector service to determine where it is installed. Returns /opt/observiq-otel-collector unless an override is detected.

This function is also used by the updater to determine which install dir to slip into the generated systemd service file.

dpaasman00 (approved):

PR #2531, opened 2025-07-16
feat: Run time configurable FeatureGates

dpaasman00:
### Proposed Change

Enables run time configurable feature gates so that users can experiment with them without requiring us to provide a new release.

Can test pretty easily with the feature gate `receiver.kafkareceiver.UseFranzGo`. You can enable it no problem using the cli argument or env var, and if you try providing a bad feature gate you get an error from the collector.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/collector/featuregates.go
+++ b/collector/featuregates.go
@@ -16,12 +16,14 @@ package collector
 
 import (
 	"fmt"
+	"strings"
 
 	"go.opentelemetry.io/collector/featuregate"
+	"go.uber.org/zap"
 )
 
 // SetFeatureFlags sets hardcoded collector feature flags
-func SetFeatureFlags(featureGates []string) error {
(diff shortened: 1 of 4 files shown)


collector/featuregates.go
```diff
@@ -34,3 +49,23 @@ func SetFeatureFlags() error {
...
+		case '-':
+			fg = fg[1:]
+			val = false
+		case '+':
+			fg = fg[1:]
+		}
+
+		if err := featuregate.GlobalRegistry().Set(fg, val); err != nil {
+			return fmt.Errorf("failed to set feature gate %s: %w", fg, err)
+		}
```
mrsillydog, line 74: One thing I think might be worth doing at the end of this loop is logging that the feature was successfully set

mrsillydog (approved):

mrsillydog (approved):

PR #2532, opened 2025-07-17
feat: bpop 3330 s3 sns sqs notification support

briangardner:
### Proposed Change
Adds SNS support to the AWS S3 Event receiver

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/receiver/awss3eventreceiver/internal/constants/constants.go
+++ b/receiver/awss3eventreceiver/internal/constants/constants.go
@@ -1,0 +1,23 @@
+// Copyright observIQ, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package constants provides shared constants for the AWS S3 Event receiver.
+package constants
+
+const (
+	// NotificationTypeS3 represents direct S3 event notifications
+	NotificationTypeS3 = "s3"
+	// NotificationTypeSNS represents S3 events wrapped in SNS notifications
+	NotificationTypeSNS = "sns"
+)
--- a/receiver/awss3eventreceiver/internal/worker/sns_parser.go
+++ b/receiver/awss3eventreceiver/internal/worker/sns_parser.go
@@ -39,8 +39,6 @@ const (
 
 // ParseSNSToS3Event parses an SNS notification containing an S3 event
 func (w *Worker) ParseSNSToS3Event(messageBody string) (*events.S3Event, error) {
-
-	w.logger.Debug("parsing standard SNS message")
 	// Standard SNS format - extract the Message field
 	return parseStandardSNSMessage(messageBody)
 }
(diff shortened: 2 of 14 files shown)


receiver/awss3eventreceiver/internal/worker/worker.go
```diff
@@ -124,9 +133,22 @@ func (w *Worker) ProcessMessage(ctx context.Context, msg types.Message, queueURL
...
-
-	notification := new(events.S3Event)
-	err := json.Unmarshal([]byte(*msg.Body), notification)
+	// Parse the message based on notification type
+	var notification *events.S3Event
+	var err error
+
+	switch w.notificationType {
+	case constants.NotificationTypeSNS:
+		logger.Info("parsing SNS message", zap.String("message_id", *msg.MessageId))
```
jsirianni: Could these be debug? Info logs could be too verbose at scale.

receiver/awss3eventreceiver/internal/worker/worker.go
```diff
@@ -124,9 +133,22 @@ func (w *Worker) ProcessMessage(ctx context.Context, msg types.Message, queueURL
...
+		notification, err = w.ParseSNSToS3Event(*msg.Body)
+	case constants.NotificationTypeS3:
+		fallthrough
+	default:
+		logger.Info("parsing S3 message", zap.String("message_id", *msg.MessageId))
+		// Direct S3 event (original behavior)
+		notification = new(events.S3Event)
+		err = json.Unmarshal([]byte(*msg.Body), notification)
+	}
```
jsirianni: Could we add a field to the Worker type `parseFunc` (or whatever makes sense). We can assign this value during setup instead of switching on the notification type for each message.

receiver/awss3eventreceiver/internal/constants/constants.go
```diff
@@ -0,0 +1,23 @@
...
+
+// Package constants provides shared constants for the AWS S3 Event receiver.
+package constants
+
+const (
+	// NotificationTypeS3 represents direct S3 event notifications
+	NotificationTypeS3 = "s3"
+	// NotificationTypeSNS represents S3 events wrapped in SNS notifications
+	NotificationTypeSNS = "sns"
+)
```
dpaasman00, line 23: Pretty nit but normally we just define these consts in `config.go` instead of defining in a specific constant file

briangardner: Yes, I had them there, but then was using them in the `worker` as well, which meant i needed to move them inward.

receiver/awss3eventreceiver/internal/worker/sns_parser.go
```diff
@@ -0,0 +1,77 @@
...
+	// Additional fields like Signature, SigningCertURL, UnsubscribeURL can be added if needed
+}
+
+const (
+	// StandardSNSMessageField is the name of the field that contains the S3 event in a standard SNS notification
+	StandardSNSMessageField = "Message"
+)
+
+// ParseSNSToS3Event parses an SNS notification containing an S3 event
+func (w *Worker) ParseSNSToS3Event(messageBody string) (*events.S3Event, error) {
```
jsirianni: I think we can remove this function and the standard parser function from the worker type.
```suggestion
func ParseSNSToS3Event(messageBody string) (*events.S3Event, error) {
```

This would require removing the logger, but we already log a similar message in the main loop right before calling the parser function.

receiver/awss3eventreceiver/internal/worker/sns_parser.go
```diff
@@ -0,0 +1,77 @@
...
+
+	w.logger.Debug("parsing standard SNS message")
```
jsirianni: I think we can remove this log, we already debug log in the main loop before this function is called.

jsirianni (changes requested):

receiver/awss3eventreceiver/internal/worker/sns_parser.go
briangardner: Done

jsirianni (approved):

PR #2533, opened 2025-07-17
feat: Add support for parsing Avro OCFs

andykellr:
### Proposed Change
If the file ends with `.avro` or has the content type `application/avro`, this will inspect the stream looking for the AVRO "magic bytes" `Obj\x01` and if found, parse as an avro set of records using the linkedin goavro library. This is the same library used in the encoding extension.

The difference between this and the encoding extension is that the extension requires a schema to be specified in the configuration and expects each payload to be a single avro record. It was implemented for use with Kafka. This parser expects Avro OCF files with embedded schemas and many records to be read from S3.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dpaasman00 (commented):
Code looks good, going to look into testing with the loader

receiver/awss3eventreceiver/internal/worker/parser.go
```diff
@@ -39,11 +39,22 @@ type LogParser interface {
...
 func newParser(ctx context.Context, stream LogStream, reader BufferedReader) (parser LogParser, err error) {
+	// check for avro first
+	isAvro, err := isAvroOcf(ctx, stream, reader)
+	if err != nil {
+		stream.Logger.Warn("failed to check if is avro", zap.Error(err))
+		isAvro = false
+	}
+	if isAvro {
+		return NewAvroOcfParser(reader, stream.Logger), nil
+	}
```
mrsillydog, line 55: I think that 1) this should be after the if clause on line 53 and 2) we should rename `stream.TryJSON` to `stream.FirstTry` or `stream.TrySpecificFormats` or something like that. Doesn't make much sense to me to check for Avro on the second attempt when we already didn't match it on the first.

andykellr: Good call. I actually thought about this last night when I was away from my computer. I'll change it.

dpaasman00 (approved):
Tested with changes made to the loader tool [here](https://github.com/observIQ/s3-loader/pull/13). Was successful.

PR #2535, opened 2025-07-19
feat: handle unprocessable errors for DLQ

briangardner:
### Proposed Change

This change allows the collector to fail fast and clear the Visibility Window in known failure scenarios like unprocessable files, missing files, or files where IAM permissions prevent getting to the data.  The built in redrive policy and DLQ capabilities in SQS handle the actual dead lettering of a message.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


receiver/awss3eventreceiver/internal/metadata/generated_telemetry.go
```diff
@@ -48,10 +50,7 @@ func (builder *TelemetryBuilder) Shutdown() {
 	builder.mu.Lock()
 	defer builder.mu.Unlock()
 	for _, reg := range builder.registrations {
-		err := reg.Unregister()
-		if err != nil {
-			fmt.Println("Error unregistering metric:", err)
-		}
```
jsirianni, line 54: Was this a generated change? Seems weird to let any potential errors go unnoticed. Equally weird that it was printing before.

If it was generated lets just let it be.

receiver/awss3eventreceiver/internal/worker/worker.go
```diff
@@ -63,6 +64,34 @@ func parseS3Event(messageBody string) (*events.S3Event, error) {
...
+	return isAccessDeniedError(err) || isNoSuchKeyError(err) || isUnsupportedFileTypeError(err)
+}
+
+// isAccessDeniedError checks if the error is an IAM permission (AccessDenied) error
+func isAccessDeniedError(err error) bool {
+	var apiErr smithy.APIError
+	if errors.As(err, &apiErr) {
+		return apiErr.ErrorCode() == "AccessDenied" || apiErr.ErrorCode() == "Forbidden"
+	}
+	return strings.Contains(err.Error(), "AccessDenied") || strings.Contains(err.Error(), "Forbidden")
```
jsirianni: I suspect the the AWS API is stable but if we are falling back here maybe we should be safe and use a case insensitive check. Here and in the other functions.

```go
        errStr := strings.ToLower(err.Error())
	return strings.Contains(errStr, "accessdenied") || strings.Contains(errStr, "forbidden")
```

receiver/awss3eventreceiver/internal/worker/worker.go
```diff
@@ -63,6 +64,34 @@ func parseS3Event(messageBody string) (*events.S3Event, error) {
 	return notification, err
 }
 
+// isDLQConditionError checks if an error should trigger DLQ behavior
+func isDLQConditionError(err error) bool {
```
jsirianni: If this function returned the error type, the switch in ProcessMessage would not need to call each "is error" function a second time.

```go
type dlqErrType string

const (
  dlqErrAccessDenied = "isAccessDeniedError"
  dlqErrNoSuchKey = "isNoSuchKeyError"
  ...
)

func isDLQConditionError(err error) (bool, dlqErrType) {
```

jsirianni (changes requested):
Nice. Left some totally optional comments. Looking forward to testing this change on Monday.

receiver/awss3eventreceiver/internal/metadata/generated_telemetry.go
briangardner: It was generated.  So any change will be overwritten on the next `make generate`.

receiver/awss3eventreceiver/internal/worker/worker.go
briangardner: Updated the IAM permissions and no such key checks.

receiver/awss3eventreceiver/internal/worker/worker.go
```diff
@@ -499,3 +548,70 @@ func (w *Worker) extendVisibility(ctx context.Context, msg types.Message, queueU
...
 }
+
+// recordDLQMetrics records metrics for DLQ conditions based on the error type.
+func (w *Worker) recordDLQMetrics(ctx context.Context, errorType string) {
+	if w.metrics == nil {
+		return
+	}
+
+	switch errorType {
+	case "iam_permission_denied":
```
jsirianni, line 573: We should have consts for these strings as they are used in several places.

PR #2554, opened 2025-07-30
fix(linuxpackage): Stop service before migrating username if running unprivileged

jsirianni:
### Proposed Change

The branch `epic/linux-package-overhaul` will migrating the Linux user `observiq-otel-collector` to `bdot`. This works great when running as the default `root` user. When running as the `observiq-otel-collector` user, the upgrade fails with this error:

```
usermod: user observiq-otel-collector is currently used by process 5287
```

This PR updates the Linux package `preinstall` script to detect this scenario and stop the service before running usermod.

Release notes will go into detail here, to ensure users are aware that the package upgrade will stop and restart the service if they have configured their collector to run as the `observiq-otel-collector` user. Additionally, notes will be included to instruct users to update their systemd override to use `bdot` instead of `observiq-otel-collector`.

### Testing

On Debian 12 and RHEL 9, I received the following log output when upgrading a Bindplane managed v1.81.0 install to this build.

```
Service observiq-otel-collector is running
Service is running as user: observiq-otel-collector
Service is running as user observiq-otel-collector, stopping service before user migration
Renaming user observiq-otel-collector to bdot
Restarting service observiq-otel-collector
Renaming group observiq-otel-collector to bdot
User migration complete.
```

I used the following steps:
  1. Install v1.81.0 using Bindplane install command
  2. Configure override to use `observiq-otel-collector` user, restart and verify functionality
  3. Modify systemd override to use `bdot` user
  4. Reload systemd, do not restart service
  5. Install new package, this will restart AFTER migrating user but before upgrading the package for minimal downtime
  6. Ensure v1.81.0 is running as `bdot` user
  7. Restart service to pickup new collector version / binary, verify now running as snapshot build with user `bdot`.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/scripts/package/preinstall.sh
+++ b/scripts/package/preinstall.sh
@@ -79,6 +79,9 @@ _migrate_user() {
         return
     fi
 
+    # Initialize service_requires_restart to false
+    service_requires_restart=false
+
     # Check if the service is running
     if systemctl is-active --quiet "$service_name"; then
         echo "Service $service_name is running"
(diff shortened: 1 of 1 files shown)


scripts/package/preinstall.sh
```diff
@@ -78,8 +79,34 @@ _migrate_user() {
...
+    if systemctl is-active --quiet "$service_name"; then
+        echo "Service $service_name is running"
+
+        # Check if the service is running as the legacy user
+        service_user=$(ps -o user= -p "$(systemctl show -p MainPID --value "$service_name")" 2>/dev/null || echo "")
+        echo "Service is running as user: $service_user"
+        if [ "$service_user" = "$legacy_username" ]; then
+            echo "Service is running as user ${legacy_username}, stopping service before user migration"
+            systemctl stop "$service_name"
+            service_was_running=true
```
joe-armstrong: We should initialize `service_was_running=false` to be safe incase `systemctl stop...` caused the script to exit with an error

PR #2555, opened 2025-07-31
fix(linuxpackage): Fail if BDOT_SKIP_RUNTIME_USER_CREATION is set and user missing

jsirianni:
### Proposed Change

When testing `BDOT_SKIP_RUNTIME_USER_CREATION`, I found that the install would proceed even if the username was not created ahead of time. This causes failures later in the install process. Its better to fail fast with a clear error.

### Testing

Now the failure is right away, and clear to the user.

```
BDOT_SKIP_RUNTIME_USER_CREATION is set to true, checking if bdot user exists
ERROR: BDOT_SKIP_RUNTIME_USER_CREATION is true but user bdot does not exist
dpkg: error processing archive /home/jsirianni/observiq-otel-collector_v1.81.0-alpha.1_linux_amd64.deb (--unpack):
```

The failure is similar on RHEL.

If I run the install script instead of installing the package directly (script with `--file` flag), the failure occurs in the script:

Screenshot From 2025-07-31 12-20-46

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/scripts/install/install_unix.sh
+++ b/scripts/install/install_unix.sh
@@ -37,6 +37,7 @@ fi
 
 # Script Constants
 COLLECTOR_USER="bdot"
+COLLECTOR_USER_LEGACY="observiq-otel-collector"
 TMP_DIR=${TMPDIR:-"/tmp"} # Allow this to be overriden by cannonical TMPDIR env var
 MANAGEMENT_YML_PATH="${BDOT_CONFIG_HOME}/manager.yaml"
 PREREQS="curl printf $SVC_PRE sed uname cut"
@@ -533,6 +534,7 @@ check_prereqs()
   os_arch_check
   package_type_check
   dependencies_check
+  user_check
   success "Prerequisite check complete!"
   decrease_indent
 }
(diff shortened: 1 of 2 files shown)


scripts/install/install_unix.sh
```diff
@@ -623,6 +625,36 @@ dependencies_check()
   succeeded
 }
 
+# This will check if the required collector user exists when BDOT_SKIP_RUNTIME_USER_CREATION is set to true.
+user_check()
```
jsirianni, line 629: Check for either user. When the package is installed, it will migrate the legacy user to `bdot` if necessary.

scripts/package/preinstall.sh
```diff
@@ -32,7 +32,12 @@ service_name="observiq-otel-collector"
 # multiple times.
 install() {
```
jsirianni, line 33: Install is called after migrate, the user `bdot` should exist here if it was migrated from the legacy user or provisioned by the systemd administrator.

PR #2558, opened 2025-08-04
chore(v2): Remove linux-arm

dpaasman00:
### Proposed Change

Removes linux-arm builds from the v2 release. The last release partially failed at the end because we upload a tarball containing all of the release artifacts to Github and it was over 2GB, Github's limit. Removing this binary should be enough to get us under the limit. I just ran a test release [here](https://github.com/observIQ/bindplane-otel-collector/actions/runs/16725721789) and the size of the resulting tarball is 1.7GB. 

I don't believe anyone will be using v2 on linux-arm and if they do we can figure it out when the time comes.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ reset: kill
 build-all: build-linux build-darwin build-windows
 
 .PHONY: build-linux
-build-linux: build-linux-amd64 build-linux-arm64 build-linux-arm build-linux-ppc64 build-linux-ppc64le
+build-linux: build-linux-amd64 build-linux-arm64 build-linux-ppc64 build-linux-ppc64le
 
 .PHONY: build-darwin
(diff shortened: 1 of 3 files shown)


mrsillydog (approved):

Makefile
```diff
@@ -67,7 +67,7 @@ reset: kill
 build-all: build-linux build-darwin build-windows
 
 .PHONY: build-linux
-build-linux: build-linux-amd64 build-linux-arm64 build-linux-arm build-linux-ppc64 build-linux-ppc64le
+build-linux: build-linux-amd64 build-linux-arm64 build-linux-ppc64 build-linux-ppc64le
```
dpaasman00, line 70: This gets called in the release flow so removing it from here. The make command for `build-linux-arm` still exists however.

PR #2564, opened 2025-08-05
feat: SecOps exporter should fall back to `time.Now()` if no timestamps are set (BPOP-3570)

dyl10s:
### Proposed Change

A user ran into a case where the Google Cloud Pub/Sub receiver was not correctly setting the timestamp or observed timestamp. The build-in encoding type from the receiver that has this bug is not going to be fixed because it is getting deprecated and we are recommended use the `textencoding` extension instead https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41800. As a good temp solution that was recommended on the ticket we can add a fallback to `time.Now()` if the timestamps are not set.

There are a couple other exporters upstream that are already doing this

- Logic Monitor
- Azure Monitor
- Dataset

so this seems like a safe change to make

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/chronicleexporter/marshal.go
+++ b/exporter/chronicleexporter/marshal.go
@@ -120,17 +120,12 @@ func (m *protoMarshaler) extractRawLogs(ctx context.Context, ld plog.Logs) (*log
 					continue
 				}
 
-				var timestamp time.Time
-
-				if logRecord.Timestamp() != 0 {
-					timestamp = logRecord.Timestamp().AsTime()
-				} else {
-					timestamp = logRecord.ObservedTimestamp().AsTime()
-				}
+				timestamp := getTimestamp(logRecord)
+				collectionTime := getObservedTimestamp(logRecord)
 
 				entry := &api.LogEntry{
 					Timestamp:      timestamppb.New(timestamp),
-					CollectionTime: timestamppb.New(logRecord.ObservedTimestamp().AsTime()),
+					CollectionTime: timestamppb.New(collectionTime),
 					Data:           []byte(rawLog),
 				}
 				logGrouper.Add(entry, namespace, logType, ingestionLabels)
@@ -519,16 +514,12 @@ func (m *protoMarshaler) extractRawHTTPLogs(ctx context.Context, ld plog.Logs) (
 					continue
 				}
 
-				var timestamp time.Time
-				if logRecord.Timestamp() != 0 {
-					timestamp = logRecord.Timestamp().AsTime()
-				} else {
-					timestamp = logRecord.ObservedTimestamp().AsTime()
-				}
+				timestamp := getTimestamp(logRecord)
+				collectionTime := getObservedTimestamp(logRecord)
 
 				entry := &api.Log{
 					LogEntryTime:         timestamppb.New(timestamp),
-					CollectionTime:       timestamppb.New(logRecord.ObservedTimestamp().AsTime()),
+					CollectionTime:       timestamppb.New(collectionTime),
 					Data:                 []byte(rawLog),
 					EnvironmentNamespace: namespace,
 					Labels:               ingestionLabels,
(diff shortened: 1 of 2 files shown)


exporter/chronicleexporter/marshal.go
```diff
@@ -541,6 +532,23 @@ func (m *protoMarshaler) extractRawHTTPLogs(ctx context.Context, ld plog.Logs) (
 	return entries, nil
 }
 
+func getTimestamp(logRecord plog.LogRecord) time.Time {
```
justinianvoss22, line 535: Could we get a quick test for these cases in marshal_test.go?

exporter/chronicleexporter/marshal.go
```diff
@@ -541,6 +532,23 @@ func (m *protoMarshaler) extractRawHTTPLogs(ctx context.Context, ld plog.Logs) (
...
+	if logRecord.Timestamp() != 0 {
+		return logRecord.Timestamp().AsTime()
+	}
+	if logRecord.ObservedTimestamp() != 0 {
+		return logRecord.ObservedTimestamp().AsTime()
+	}
+	return time.Now()
```
justinianvoss22: We can just use this instead of repeating behavior
```suggestion
	return getObservedTimestamp(logRecord)
```

exporter/chronicleexporter/marshal_test.go
```diff
@@ -135,6 +139,55 @@ func TestProtoMarshaler_MarshalRawLogs(t *testing.T) {
...
+				require.Equal(t, batch.Entries[1].Timestamp.Seconds, ts1)
+				// This gets set to Time.Now() instead of being default 0
+				require.NotEqual(t, batch.Entries[1].CollectionTime.Seconds, 0)
+
+				require.Equal(t, batch.Entries[2].CollectionTime.Seconds, ts1)
+				require.Equal(t, batch.Entries[2].Timestamp.Seconds, ts2)
+
+				// This gets set to Time.Now() instead of being default 0
+				require.NotEqual(t, batch.Entries[3].CollectionTime.Seconds, 0)
+				require.NotEqual(t, batch.Entries[3].CollectionTime.Seconds, 0)
```
justinianvoss22, line 188:
```suggestion
				require.NotEqual(t, batch.Entries[3].Timestamp.Seconds, 0)
```

exporter/chronicleexporter/marshal_test.go
```diff
@@ -135,6 +139,55 @@ func TestProtoMarshaler_MarshalRawLogs(t *testing.T) {
...
+				require.Equal(t, batch.Entries[1].Timestamp.Seconds, ts1)
+				// This gets set to Time.Now() instead of being default 0
+				require.NotEqual(t, batch.Entries[1].CollectionTime.Seconds, 0)
+
+				require.Equal(t, batch.Entries[2].CollectionTime.Seconds, ts1)
+				require.Equal(t, batch.Entries[2].Timestamp.Seconds, ts2)
+
+				// This gets set to Time.Now() instead of being default 0
+				require.NotEqual(t, batch.Entries[3].CollectionTime.Seconds, 0)
+				require.NotEqual(t, batch.Entries[3].Timestamp.Seconds, 0)
```
justinianvoss22: did some clean up of these comments here to make it clearer what the results should be:
```suggestion
				// Timestamp gets set with observed timestamp if it is not set
				require.Equal(t, batch.Entries[0].Timestamp.Seconds, ts1)
				require.Equal(t, batch.Entries[0].CollectionTime.Seconds, ts1)

				// Collection time gets set to Time.Now() instead of being default 0
				require.Equal(t, batch.Entries[1].Timestamp.Seconds, ts1)
				require.NotEqual(t, batch.Entries[1].CollectionTime.Seconds, 0)

				// Timestamp and collection time get set to their set values
				require.Equal(t, batch.Entries[2].CollectionTime.Seconds, ts1)
				require.Equal(t, batch.Entries[2].Timestamp.Seconds, ts2)

				// Collection time and timestamp get set to Time.Now() instead of being default 0
				require.NotEqual(t, batch.Entries[3].CollectionTime.Seconds, 0)
				require.NotEqual(t, batch.Entries[3].Timestamp.Seconds, 0)
```

exporter/chronicleexporter/marshal_test.go
```diff
@@ -135,6 +139,55 @@ func TestProtoMarshaler_MarshalRawLogs(t *testing.T) {
...
+				record1.SetObservedTimestamp(timestamp1)
+				record2 := logs.ResourceLogs().AppendEmpty().ScopeLogs().AppendEmpty().LogRecords().AppendEmpty()
+				record2.SetTimestamp(timestamp1)
+				record2.Body().SetStr("Log 2 with timestamp set")
+				record3 := logs.ResourceLogs().AppendEmpty().ScopeLogs().AppendEmpty().LogRecords().AppendEmpty()
+				record3.SetObservedTimestamp(timestamp1)
+				record3.SetTimestamp(timestamp2)
+				record3.Body().SetStr("Log 3 with timestamp and collection time set")
+				record4 := logs.ResourceLogs().AppendEmpty().ScopeLogs().AppendEmpty().LogRecords().AppendEmpty()
+				record4.Body().SetStr("Log 4 with no timestamp set")
```
justinianvoss22:
```suggestion
				record4.Body().SetStr("Log 4 with no timestamp or collection time set")
```

justinianvoss22 (approved):

PR #2565, opened 2025-08-11
feat(install): Install macOS artifacts from bdot.bindplane.com

jsirianni:
### Proposed Change

Based on the Linux change here https://github.com/observIQ/bindplane-otel-collector/pull/2536

### Testing

Working great. Tested install locally and tested the generated uninstall command shown at the end of the install flow.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/scripts/install/install_macos.sh
+++ b/scripts/install/install_macos.sh
@@ -17,7 +17,7 @@ set -e
 
 # Collector Constants
 SERVICE_NAME="com.observiq.collector"
-DOWNLOAD_BASE="https://github.com/observIQ/bindplane-otel-collector/releases/download"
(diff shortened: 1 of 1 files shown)


scripts/install/install_macos.sh
```diff
@@ -382,6 +382,32 @@ dependencies_check()
   succeeded
 }
 
+# This will validate that the version is at least v1.82.0
```
dpaasman00, line 385: Maybe worthwhile clarifying that this script doesn't require v1.82.0 like the unix script does, but that we'll enforce it for consistency between scripts

jsirianni: Sorry missed this on merge. Ill rebase the nix PR and implement this.

PR #2625, opened 2025-09-03
feat: Add secops raw bytes egress metric

dyl10s:
### Proposed Change

This PR exposes a metric tracking the raw bytes sent to secops. This is the first step needed to track how much data SecOps ingests from Bindplane agents.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/exporter/chronicleexporter/grpc_exporter.go
+++ b/exporter/chronicleexporter/grpc_exporter.go
@@ -65,10 +65,16 @@ func newGRPCExporter(cfg *Config, params exporter.Settings, telemetry *metadata.
 		exporterID: params.ID.String(),
 		marshaler:  marshaler,
 		telemetry:  telemetry,
-		metricAttributes: attribute.NewSet(attribute.KeyValue{
-			Key:   "exporter",
-			Value: attribute.StringValue(params.ID.String()),
-		}),
+		metricAttributes: attribute.NewSet(
+			attribute.KeyValue{
+				Key:   "exporter",
+				Value: attribute.StringValue(params.ID.String()),
+			},
+			attribute.KeyValue{
+				Key:   "exporter_type",
+				Value: attribute.StringValue(params.ID.Type().String()),
+			},
+		),
 	}, nil
 }
 
--- a/exporter/chronicleexporter/http_exporter.go
+++ b/exporter/chronicleexporter/http_exporter.go
@@ -65,10 +65,16 @@ func newHTTPExporter(cfg *Config, params exporter.Settings, telemetry *metadata.
 		set:       params.TelemetrySettings,
 		marshaler: marshaler,
 		telemetry: telemetry,
-		metricAttributes: attribute.NewSet(attribute.KeyValue{
-			Key:   "exporter",
-			Value: attribute.StringValue(params.ID.String()),
-		}),
+		metricAttributes: attribute.NewSet(
+			attribute.KeyValue{
+				Key:   "exporter",
+				Value: attribute.StringValue(params.ID.String()),
+			},
+			attribute.KeyValue{
+				Key:   "exporter_type",
+				Value: attribute.StringValue(params.ID.Type().String()),
+			},
+		),
 	}, nil
 }
 


exporter/chronicleexporter/grpc_exporter.go
```diff
@@ -123,6 +128,7 @@ func (exp *grpcExporter) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
 			return err
 		}
 	}
+	exp.telemetry.ExporterRawBytes.Add(ctx, int64(totalSize), metric.WithAttributeSet(exp.metricAttributes))
```
briangardner: If we upload 4 of 5 payloads in the loop above, then error on the 5th payload, then it seems like we ignore that data that has been egressed.  It feels like we get the raw bytes of each payload, and add it to the metric after we've successfully uploaded each payload.

exporter/chronicleexporter/grpc_exporter_test.go
```diff
@@ -188,6 +218,20 @@ func TestGRPCExporter(t *testing.T) {
 			}
 
 			require.Equal(t, tc.expectedRequests, mockServer.requests)
+
+			if tc.expectedErr == "" {
```
briangardner: I feel like we should have a separate test that tests the metric, instead of co-mingling it with testing the exporter itself.  That's typically how tests are done upstream as well.

exporter/chronicleexporter/http_exporter.go
```diff
@@ -195,6 +200,7 @@ func (exp *httpExporter) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
 			}
 		}
 	}
+	exp.telemetry.ExporterRawBytes.Add(ctx, int64(totalSize), metric.WithAttributeSet(exp.metricAttributes))
```
briangardner: Same issue here.  If we successfully send 4 payloads of 5, we don't count those metrics.

exporter/chronicleexporter/http_exporter_test.go
```diff
@@ -194,6 +221,19 @@ func TestHTTPExporter(t *testing.T) {
 			}
 
 			require.Equal(t, tc.expectedRequests, mockServer.requestCount)
+
```
briangardner: Same comment here.  Split testing the counter out from testing the exporter

exporter/chronicleexporter/marshal.go
```diff
@@ -488,15 +490,16 @@ func (m *protoMarshaler) buildGRPCRequest(entries []*api.LogEntry, logType, name
...
 	if err != nil {
-		return nil, fmt.Errorf("extract raw logs: %w", err)
+		return nil, 0, fmt.Errorf("extract raw logs: %w", err)
 	}
-	return m.constructHTTPPayloads(rawLogs), nil
+	return m.constructHTTPPayloads(rawLogs), totalSize, nil
 }
 
-func (m *protoMarshaler) extractRawHTTPLogs(ctx context.Context, ld plog.Logs) (map[string][]*api.Log, error) {
+func (m *protoMarshaler) extractRawHTTPLogs(ctx context.Context, ld plog.Logs) (map[string][]*api.Log, int, error) {
```
briangardner: as we cannot have negative bytes, it probably makes sense to use a uint for all of the size parameters

exporter/chronicleexporter/marshal.go
```diff
@@ -92,15 +92,16 @@ func newProtoMarshaler(cfg Config, teleSettings component.TelemetrySettings, tel
...
 	if err != nil {
-		return nil, fmt.Errorf("extract raw logs: %w", err)
+		return nil, 0, fmt.Errorf("extract raw logs: %w", err)
 	}
-	return m.constructPayloads(logGrouper), nil
+	return m.constructPayloads(logGrouper), totalSize, nil
 }
 
-func (m *protoMarshaler) extractRawLogs(ctx context.Context, ld plog.Logs) (*logGrouper, error) {
+func (m *protoMarshaler) extractRawLogs(ctx context.Context, ld plog.Logs) (*logGrouper, int, error) {
```
briangardner: use a `uint` instead, since we cannot have negative bytes

exporter/chronicleexporter/grpc_exporter.go
dyl10s: I agree this is an issue. I want to try to keep the happy path as fast as possible and avoid any extra counting loops. What I have done keeps the existing logic, but, when there is an error, I count up the raw bytes from each payload that successfully sent. This way we are only doing that extra work if we have a partial success.

briangardner (approved):

exporter/chronicleexporter/grpc_exporter.go
```diff
@@ -64,6 +65,10 @@ func newGRPCExporter(cfg *Config, params exporter.Settings, telemetry *metadata.
 		exporterID: params.ID.String(),
 		marshaler:  marshaler,
 		telemetry:  telemetry,
+		metricAttributes: attribute.NewSet(attribute.KeyValue{
```
briangardner: @dyl10s - I think we need to adjust this attribute.  I think `params.ID.String()` will give us something like `chronicle/some_unique_name`, which could result in high cardinality when sent back to bindplane.  We should either replace it with `params.ID.Type()`, or add a second attribute with the exporter type, so we can easily filter on just `chronicle` explicitly

exporter/chronicleexporter/http_exporter.go
```diff
@@ -64,6 +65,10 @@ func newHTTPExporter(cfg *Config, params exporter.Settings, telemetry *metadata.
 		set:       params.TelemetrySettings,
 		marshaler: marshaler,
 		telemetry: telemetry,
+		metricAttributes: attribute.NewSet(attribute.KeyValue{
```
briangardner: @dyl10s - same thing here as in GRPC

exporter/chronicleexporter/grpc_exporter.go
dyl10s: So for this you are correct it gets back for example: `chronicle/some_unique_name`. I do this specifically because it matches all the existing throughput metrics with the field `exporter` in place of `processor`.

Example throughput measurement with `processor`:
```
bindplane_agent_measurements{account="01K48BMZAMSD9P5PHDRE5233DC", agent="0197e67d-b234-7a48-af8d-11b4d36cc971", configuration="Mac", fleet="maxfleet", gateway="false", metric="otelcol_processor_throughputmeasurement_log_data_size", name="test2-0", pipeline_type="logs", position="d1", processor="throughputmeasurement/_d1_logs_test2-0", usage="false", version="v2"}
```

Exporter measurement with `exporter`:
```
bindplane_agent_measurements{account="01K48BMZAMSD9P5PHDRE5233DC", agent="0197e67d-b234-7a48-af8d-11b4d36cc971", configuration="Mac", exporter="chronicle/Secops", fleet="maxfleet", gateway="false", metric="otelcol_exporter_raw_bytes_B", usage="false", version="v2"}
```

So the existing measurements also just report the component ID. Then I am guessing enterprise actually looks this value up based on the name and account information. Similar to how you were able to add the destination type in to the existing metrics without having it reported from the agent. Let me know if you think this case is different though or if I am misunderstanding.

dyl10s: Screenshot 2025-09-05 at 10 12 02 AM

Added as an additional attribute

briangardner (approved):

PR #2627, opened 2025-09-08
fix(v2): Linux package staging directory

jsirianni:
### Proposed Change

Various fixes for the Linux package's staging dir. See inline comments.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


.goreleaser.yml
```diff
@@ -89,9 +89,6 @@ archives:
       - src: release_deps/windows_service.json
         dst: install
         strip_parent: true
-      - src: release_deps/bindplane-otel-collector.service
-        dst: "install"
-        strip_parent: true
```
jsirianni, line 94: This file is created with a heredoc in the package postinstall script, and was deleted from the repo. We don't need to ship the service file in the archive. If this is a problem in the future, we can document the service file for the user.

.goreleaser.yml
```diff
@@ -126,26 +123,26 @@ nfpms:
         dst: /usr/share/bindplane-otel-collector/stage/bindplane-otel-collector/LICENSE
         file_info:
           mode: 0644
-          owner: bindplane-otel-collector
-          group: bindplane-otel-collector
+          owner: root
+          group: root
```
jsirianni, line 127: Should be root. At some point `bindplane-otel-collector` will be migrated to `bdot`, and we will also support user defined username. Because of this, we cannot assume the username.

The postinstall script will manage the username instead of Goreleaser.

Makefile
```diff
@@ -236,7 +236,6 @@ release-prep:
 	@cp -r ./plugins release_deps/
 	@cp service/com.bindplane.otel.collector.plist release_deps/com.bindplane.otel.collector.plist
 	@jq ".files[] | select(.service != null)" windows/wix.json >> release_deps/windows_service.json
-	@cp service/bindplane-otel-collector.service release_deps/bindplane-otel-collector.service
```
jsirianni, line 239: This file was removed. See first comment above.

scripts/package/postinstall.sh
```diff
@@ -28,6 +28,7 @@ install() {
   chmod 0755 "${BDOT_CONFIG_HOME}"
   chown bindplane-otel-collector:bindplane-otel-collector "${BDOT_CONFIG_HOME}"
   rm -f "${BDOT_CONFIG_HOME}/bindplane-otel-collector" || true
+  chown -R bindplane-otel-collector:bindplane-otel-collector /usr/share/bindplane-otel-collector/stage/bindplane-otel-collector
```
jsirianni, line 31: This resolves an issue where `/opt/bindplane-otel-collector` had a bunch of files and directories owned by root.

PR #2629, opened 2025-09-08
feat(v2): Support non root install

jsirianni:
### Proposed Change

Support installing as non root. This is identical to the v1 implementation.

Because the postinstall script uses a systemd override for configuring the user, the main service file remains unmodified. 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/package/postinstall.sh
+++ b/scripts/package/postinstall.sh
@@ -21,7 +21,14 @@ set -e
 [ -f /etc/default/bindplane-otel-collector ] && . /etc/default/bindplane-otel-collector
 [ -f /etc/sysconfig/bindplane-otel-collector ] && . /etc/sysconfig/bindplane-otel-collector
 
+# The collectors installation directory
 : "${BDOT_CONFIG_HOME:=/opt/bindplane-otel-collector}"
+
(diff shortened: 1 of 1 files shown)


scripts/package/postinstall.sh
```diff
@@ -83,6 +90,24 @@ EOF
...
+  # Ensure the override dir exists.
+  override_dir="/etc/systemd/system/bindplane-otel-collector.service.d"
+  if [ ! -d "$override_dir" ]; then
+    mkdir -p "$override_dir"
+    echo "Created systemd override directory at $override_dir"
+  fi
+
+  # If BDOT_UNPRIVILEGED is true, add an override to run the service as the
+  # bdot (unprivileged) user.
+  override_user_path="${override_dir}/10-package-customizations-username.conf"
```
dpaasman00, line 103: Why is it called "10-package-customizations-username.conf"?

jsirianni: You can have more than one override. Naming is somewhat arbitrary. Because it pertains to the username, I named it after "username".

PR #2639, opened 2025-09-09
port statedump

schmikei:
### Proposed Change
Still need a way to validate this but was a port of the statedump of https://github.com/mandiant/macos-UnifiedLogs... Looking for a way to generate and do some testing around this since my macbook had zero statedumps according to `hexdump` 🤔 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/extension/encoding/macosunifiedloggingencodingextension/statedump.go
+++ b/extension/encoding/macosunifiedloggingencodingextension/statedump.go
@@ -4,16 +4,207 @@
 package macosunifiedloggingencodingextension
 
 import (
+	"bytes"
+	"encoding/binary"
+	"encoding/hex"
 	"fmt"
+	"strings"
 )
+
+type StateDump struct {
+	ChunkTag        uint32
+	ChunkSubtag     uint32
+	ChunkDataSize   uint64
+	FirstProcID     uint64
+	SecondProcID    uint32
+	TTL             uint8
+	UnknownReserved []uint8
+	ContinuousTime  uint64
+	ActivityID      uint64
+	UUID            string
+	UnknownDataType uint32 // 1 = plist, 3 = custom object?, 2 = (protocol buffer?)
+	UnknownDataSize uint32 // Size of statedump data
+	DecoderLibrary  string
+	DecoderType     string
+	TitleName       string
+	Data            []uint8
+}
+
+// ParseStateDump parses Statedump log entry. Statedumps are special log entries that may contain a plist file, custom object, or protocol buffer
+func ParseStateDump(data []byte) (*StateDump, error) {
+	if len(data) < 72 { // Minimum size for basic structure
+		return nil, fmt.Errorf("statedump data too small: %d bytes", len(data))
+	}
+
+	reader := bytes.NewReader(data)
+	result := &StateDump{}
+
+	// Parse the header fields
+	if err := binary.Read(reader, binary.LittleEndian, &result.ChunkTag); err != nil {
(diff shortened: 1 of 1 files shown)


extension/encoding/macosunifiedloggingencodingextension/statedump.go
```diff
@@ -4,16 +4,449 @@
...
+	Data            []uint8
+}
+
+// ParseStateDump parses Statedump log entry. Statedumps are special log entries that may contain a plist file, custom object, or protocol buffer
+func ParseStateDump(data []byte) (*StateDump, error) {
+	if len(data) < 72 { // Minimum size for basic structure
+		return nil, fmt.Errorf("statedump data too small: %d bytes", len(data))
+	}
+
+	reader := bytes.NewReader(data)
```
Caleb-Hurshman, line 40: Curious about this binary reader approach. Is this pretty standard for go? Probably better than the Take func I wrote for the firehose PR? Lmk what you think.

extension/encoding/macosunifiedloggingencodingextension/statedump.go
```diff
@@ -4,16 +4,449 @@
...
+		return getDNSConfig(objectData)
+	case "Network information":
+		return getNetworkInterface(objectData)
+	default:
+		return fmt.Sprintf("Unsupported Statedump object: %s-%s", name, encodeStandard(objectData))
+	}
+}
+
+// formatUUID formats a UUID byte array into standard UUID string format
+func formatUUID(uuidBytes []byte) string {
```
Caleb-Hurshman, line 161: This would be nice to add to helpers.go (once that is merged)

extension/encoding/macosunifiedloggingencodingextension/statedump.go
schmikei: I think both approaches are probably fine. Just one stores state inside of the reader and the other makes the user know the size of the bytes they are ingesting, whereas this approach you need to know the general order of how the byte sequence is structured...

Both are not ideal in my head compared to building some kind of decoder, but that'd be a slightly higher lift in my opinion.

extension/encoding/macosunifiedloggingencodingextension/statedump.go
```diff
@@ -4,16 +4,449 @@
...
+		if len(collection.Statedumps) >= 1000 {
+			break
+		}
+	}
+
+	return collection, nil
+}
+
+// GetStatedumpsByType filters statedumps by their data type
+func (sc *StatedumpCollection) GetStatedumpsByType(dataType uint32) []StateDump {
```
Caleb-Hurshman: How do you plan on using the StatedumpCollection methods?

extension/encoding/macosunifiedloggingencodingextension/statedump.go
```diff
@@ -4,16 +4,449 @@
...
+	return fmt.Sprintf("%08X%04X%04X%04X%012X",
+		binary.BigEndian.Uint32(uuidBytes[0:4]),
+		binary.BigEndian.Uint16(uuidBytes[4:6]),
+		binary.BigEndian.Uint16(uuidBytes[6:8]),
+		binary.BigEndian.Uint16(uuidBytes[8:10]),
+		uuidBytes[10:16])
+}
+
+// encodeStandard provides hex encoding for unsupported objects
+func encodeStandard(data []byte) string {
```
Caleb-Hurshman, line 176: Another nice candidate for helpers.go

PR #2646, opened 2025-09-10
Firehose loss entry parsing

Caleb-Hurshman:
### Proposed Change

Adds firehose loss entry parsing, test passes

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/extension/encoding/macosunifiedloggingencodingextension/firehose_loss.go
+++ b/extension/encoding/macosunifiedloggingencodingextension/firehose_loss.go
@@ -1,7 +1,23 @@
 package macosunifiedloggingencodingextension
 
-type FirehoseLoss struct{}
+import "encoding/binary"
+
+type FirehoseLoss struct {
+	StartTime uint64
+	EndTime   uint64
+	Count     uint64
+}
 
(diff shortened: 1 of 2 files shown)


extension/encoding/macosunifiedloggingencodingextension/firehose_loss.go
```diff
@@ -1,7 +1,23 @@
 package macosunifiedloggingencodingextension
 
-type FirehoseLoss struct{}
+import "encoding/binary"
+
+type FirehoseLoss struct {
+	StartTime uint64
+	EndTime   uint64
```
schmikei, line 8: You think its valuable to put these into a time.TIme? I know the rust implementation does not do that but its hard to tell if this UnixEpoch or not. The test case seems to target a time in 1992?

Tuesday, June 2, 1992 8:58:48 AM?

Maybe a comment on the uint64 value could be useful

Caleb-Hurshman: I'm not sure, maybe it makes more sense to handle all of the times when the timesync data is handy. I'll add a comment for now

schmikei (approved):
Aside from the time comment seems pretty good to me

PR #2649, opened 2025-09-11
Firehose signpost entry parsing

Caleb-Hurshman:
### Proposed Change

Adds signpost entry parsing. Creates two more internal packages: `sharedcache` and `types`. I did not rewrite the shared caching logic yet because the non-activity entry type will also use that.
The plan is to get the basic parsing for each of those merged into the firehose branch, then work on the caching logic.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/extension/encoding/macosunifiedloggingencodingextension/internal/sharedcache/dsc.go
+++ b/extension/encoding/macosunifiedloggingencodingextension/internal/sharedcache/dsc.go
@@ -84,7 +84,7 @@ func ParseDSC(data []byte, uuid string) (*SharedCacheStrings, error) {
 	offset += 4
 
 	// Parse UUIDs
-	dsc.UUIDs = make([]SharedCacheUUID, numUUIDs)
+	dsc.UUIDs = make([]UUIDDescriptor, numUUIDs)
 	for i := uint32(0); i < numUUIDs; i++ {
 		if len(data) < offset+16 {
 			return nil, fmt.Errorf("DSC file too small for UUID %d", i)
@@ -99,7 +99,7 @@ func ParseDSC(data []byte, uuid string) (*SharedCacheStrings, error) {
 			uuidBytes[12], uuidBytes[13], uuidBytes[14], uuidBytes[15])
 		offset += 16
 
-		dsc.UUIDs[i] = SharedCacheUUID{UUID: uuid}
+		dsc.UUIDs[i] = UUIDDescriptor{UUID: uuid}
 	}
 
 	// Parse path strings for UUIDs
@@ -124,7 +124,7 @@ func ParseDSC(data []byte, uuid string) (*SharedCacheStrings, error) {
 	// For now, create a single range covering remaining data
 	if offset < len(data) {
 		remainingData := data[offset:]
-		dscRange := SharedCacheRange{
+		dscRange := RangeDescriptor{
 			RangeOffset:      0,
 			RangeSize:        uint32(len(remainingData)),
 			UnknownUUIDIndex: 0,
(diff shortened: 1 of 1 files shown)


extension/encoding/macosunifiedloggingencodingextension/internal/firehose/firehose_signpost.go
```diff
@@ -14,10 +14,129 @@
...
+		// Then the signpost name has the same value after as the large_shared_cache
+		if signpost.FirehoseFormatters.LargeSharedCache != 0 {
+			data, _, _ = utils.Take(data, 2)
+		}
+	}
+
+	return signpost, data
+}
+
+func getFirehoseSignpost(signpost Signpost, provider types.FileProvider, stringOffset uint64, firstProcID uint64, secondProcID uint32, catalogs types.CatalogChunk) (types.MessageData, error) {
```
Caleb-Hurshman, line 110: Note: This method is not fully implemented yet; it relies on dsc and uuidtext functions that will be rewritten soon.

extension/encoding/macosunifiedloggingencodingextension/internal/types/types.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// FileProvider represents a file provider interface
+type FileProvider interface {
+	// Add methods as needed
+}
+
+// extractStringAtOffset extracts a null-terminated string from a byte array at a specific offset
+// This matches the rust implementation's approach for extracting subsystem and category strings
+func extractStringAtOffset(data []byte, offset int) string {
```
schmikei, line 180: Should this go in the util. package?

extension/encoding/macosunifiedloggingencodingextension/internal/firehose/firehose_signpost.go
```diff
@@ -14,10 +14,129 @@
...
+				extraOffsetValue = fmt.Sprintf("%x%x", largeOffset, stringOffset)
+			} else {
+				extraOffsetValue = fmt.Sprintf("%x%x", largeOffset, stringOffset)
+			}
+			extraOffsetValueResult, err := strconv.ParseUint(extraOffsetValue, 16, 64)
+			if err != nil {
+				return types.MessageData{}, fmt.Errorf("error parsing extra offset value: %w", err)
+			}
+			messageData, err := sharedcache.GetSharedFormatString(uint32(extraOffsetValueResult), firstProcID, secondProcID)
+			if err != nil {
```
schmikei, line 135: https://github.com/mandiant/macos-UnifiedLogs/blob/11220fb684f5fb65056f60b19fedea74661796fb/src/chunks/firehose/signpost.rs#L167-L246

It seems like we're missing some logic or maybe I'm missing something but could you point me where we're implementing the large offset combination?

Caleb-Hurshman: See my comment at the top of this func, this is more of a placeholder. Fully implementing this requires some dsc/uuidtext methods, so it's impossible to finish this function in this PR.

schmikei: Ope my browser did not get that comment when I made this comment, apologies about missing that

extension/encoding/macosunifiedloggingencodingextension/internal/types/types.go
Caleb-Hurshman: So far this is only used in this file, so I don't think it needs to be moved right now. I'll put a pin in that, though, definitely open to moving it.

PR #2651, opened 2025-09-11
Firehose nonactivity entry parsing

Caleb-Hurshman:
##### Checklist
- [x] Changes are tested
- [ ] CI has passed


extension/encoding/macosunifiedloggingencodingextension/internal/firehose/firehose_nonactivity.go
```diff
@@ -4,23 +4,138 @@
...
+	FirehoseFormatters      FirehoseFormatters
 }
 
-// ParseFirehoseNonActivity parses a firehose non activity entry
 func ParseFirehoseNonActivity(data []byte, flags uint16) (FirehoseNonActivity, []byte) {
-	return FirehoseNonActivity{}, data
+	var nonActivity FirehoseNonActivity
+	activityIDCurrent := uint16(0x1) // has_current_aid flag
+	if (flags & activityIDCurrent) != 0 {
+		firehoseData, unknownActivityID, _ := utils.Take(data, 4)
```
schmikei, line 39: Sort of throughout, but should we be passing the error up here and handling throughout instead of ignoring the error from Take?

Caleb-Hurshman: Yes absolutely. I anticipate handling the errors all at once. It seemed better to focus on the parsing for now, especially if we end up liking the binary reader approach better and ended up replacing the Take() code.

PR #2660, opened 2025-09-16
Firehose DSC/UUIDText caching

Caleb-Hurshman:
### Proposed Change

Update the DSC and UUID text logic for the encoding, and update the firehose event types the depend on it (nonactivity/signpost)

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/macosunifiedloggingreceiver/receiver.go
+++ b/receiver/macosunifiedloggingreceiver/receiver.go
@@ -433,11 +433,6 @@ func (r *macosUnifiedLogReceiver) loadUUIDTextData() error {
 		totalBytes += len(data)
 	}
 
-	r.set.Logger.Info("=== RECEIVER DEBUG: UUID TEXT RAW DATA LOADED ===",
-		zap.Int("totalUUIDTextFiles", len(uuidTextRawData)),
-		zap.Strings("loadedFiles", loadedFiles),
-		zap.Int("totalBytes", totalBytes))
-
 	// Configure the encoding extension with the raw UUID text data
 	if macosExt, ok := r.encodingExt.(*macosunifiedloggingencodingextension.MacosUnifiedLoggingExtension); ok {
 		macosExt.SetUUIDTextRawData(uuidTextRawData)


receiver/macosunifiedloggingreceiver/receiver.go
```diff
@@ -107,6 +107,30 @@ func (r *macosUnifiedLogReceiver) Start(ctx context.Context, host component.Host
...
+	// Load UUID text data if paths are configured
+	if len(r.config.UUIDTextPaths) > 0 {
+		err := r.loadUUIDTextData()
+		if err != nil {
+			r.set.Logger.Warn("Failed to load UUID text data, message parsing may be inaccurate", zap.Error(err))
+		} else {
+			r.set.Logger.Info("Successfully loaded UUID text data for accurate message parsing")
+		}
+	} else {
+		r.set.Logger.Warn("No UUID text paths configured, message parsing may be inaccurate")
```
schmikei, line 119: I'm sort of unfamiliar, how is message parsing accuracy affected by UUIDTextData?

receiver/macosunifiedloggingreceiver/receiver.go
```diff
@@ -365,45 +378,201 @@ func (r *macosUnifiedLogReceiver) loadTimesyncFromPath(path string, timesyncData
...
+
+	// Log all loaded UUID text files for debugging
+	var loadedFiles []string
+	totalBytes := 0
+	for filePath, data := range uuidTextRawData {
+		loadedFiles = append(loadedFiles, filePath)
+		totalBytes += len(data)
+	}
+
+	r.set.Logger.Info("=== RECEIVER DEBUG: UUID TEXT RAW DATA LOADED ===",
```
schmikei: Not sure if we're still looking to keep debug log lines but figured I'd highlight that this seems too verbose for a merge into main quite yet.

receiver/macosunifiedloggingreceiver/receiver.go
Caleb-Hurshman: My understanding is that UUIDs correspond to shared strings that appear across a lot of log messages (like the subsystem attribute). Instead of having that subsystem string on every log, the log will reference a UUID, which gets translated into the proper subsystem.

Caleb-Hurshman:
> I'd like to see some more tests/ports around the dsc and uuidtext parsing before approving. Let me know if I'm missing anything though as its a rather large PR 😅

Not missing anything! You're right that there aren't a lot of tests, but I have a ticket for unit testing before actually merging to main. Let's chat more about generating some test data for these features.

PR #2670, opened 2025-09-26
fix(awss3eventreceiver): Decode object key before using it

kuiperda:
### Proposed Change

S3 events encode object keys with special characters like `=`. For example, 
```
logs/hd=test/file.txt
```  
becomes 
```
logs/hd%3Dtest/file.txt
```

The fix is to just decode the key right away when we get it from the event, otherwise we get 404s when trying to get the object.

##### Checklist
- [ ] Changes are tested
- [x] CI has passed

Added a unit test which fails without these changes. 
I will need assistance setting up an environment to fully test this change.

### DIFF
--- a/receiver/awss3eventreceiver/internal/worker/worker.go
+++ b/receiver/awss3eventreceiver/internal/worker/worker.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


kuiperda:
@claude review

receiver/awss3eventreceiver/internal/worker/worker.go
```diff
@@ -237,29 +238,52 @@ func (w *Worker) ProcessMessage(ctx context.Context, msg types.Message, queueURL
...
 			continue
 		}
 
 		if w.bucketNameFilter != nil && !w.bucketNameFilter.MatchString(record.S3.Bucket.Name) {
 			recordLogger.Debug("skipping record due to bucket name filter", zap.String("bucket", record.S3.Bucket.Name))
 			continue
 		}
-		if w.objectKeyFilter != nil && !w.objectKeyFilter.MatchString(record.S3.Object.Key) {
-			recordLogger.Debug("skipping record due to object key filter", zap.String("key", record.S3.Object.Key))
+		if w.objectKeyFilter != nil && !w.objectKeyFilter.MatchString(decodedKey) {
```
briangardner, line 279: I'm concerned about this change.  This is using a customer defined regex filter on the object name.  If we start comparing to the url decoded version, that could break existing filters.  And we do not use a url decoded value for the bucket name on line 275.

kuiperda: My rationale here was that customers would expect to type the actual object name instead of accounting for parts of it being encoded. But I get that this would be breaking. 

I didn't think about the bucket name. 

For both of these, which value would you expect to match against? Encoded or Decoded?

briangardner: I think we need to test how it works.  I agree that customers would expect the actual object name.  But then it seems like we should use the actual value, no the decoded value in that case, right?

kuiperda: The decoded value should be the actual value with `=` instead of `%3D`

briangardner:
@claude run

kuiperda:
I think Claude isn't available in this repo

receiver/awss3eventreceiver/internal/worker/worker.go
briangardner: Can you add a test case covering that?

briangardner:
I updated the `s3-loader` tool to generate special characters in the object keys, and ran a build of the collector.

image

```
===========================================================================
                           VERIFICATION RESULTS
===========================================================================
Metric                    | Input           | Output          | % Change       
---------------------------------------------------------------------------
Object Count              | 50              | 50              | 0.0%           
Logs Per Object           | 1,000           | 1,000           | 0.0%           
Total Logs                | 50,000          | 50,000          | 0.0%           
Unique Hashes             | 50,000          | 50,000          | 0.0%           
===========================================================================
```

PR #2679, opened 2025-10-06
Adds agent ID change support

briangardner:
Adds the ability for the agent to change its ID.

- Adds `AgentIDFromInstanceUid` to create an AgentID from an OpAMP InstanceUid, trying UUID and then ULID parsing.
- Persists the agent ID to the manager config file when it changes.
- Updates agent description with the new ID.


briangardner:
@claude review

opamp/observiq/observiq_client.go
```diff
@@ -384,6 +395,48 @@ func (c *Client) onMessageFuncHandler(ctx context.Context, msg *types.MessageDat
...
+		c.logger.Error("Failed to persist new agent ID to manager config", zap.Error(err))
+		return fmt.Errorf("failed to persist new agent ID: %w", err)
+	}
+
+	// Update agent description with new ID
+	if err := c.opampClient.SetAgentDescription(c.ident.ToAgentDescription()); err != nil {
+		c.logger.Error("Failed to update agent description with new instance UID", zap.Error(err))
+		// Not rolling back here since the ID is already persisted
+		return fmt.Errorf("failed to update agent description: %w", err)
+	}
```
andykellr, line 434: I think OpAMP should do this automatically and is a bug in `opamp-go`. It would be nice to confirm this is the case and fix it.

dpaasman00: Looking at OpAMP Go, when we receive an AgentIdentification message, it calls [this](https://github.com/open-telemetry/opamp-go/blob/main/client/internal/receivedprocessor.go#L381-L398) function which only updates the Instance ID on the sender. If OpAMP Go should be doing this it might not be a quick change upstream.

briangardner: @andykellr - did you create the issue in `opamp-go` to link to before I merge this?

PR #2687, opened 2025-10-07
Macos unified encoding file based unit tests

postnati:
### Proposed Change
Adding all unit tests that read log data from a file for the tests. We skipped over these initially.

### DIFF
--- a/extension/encoding/macosunifiedloggingencodingextension/README.md
+++ b/extension/encoding/macosunifiedloggingencodingextension/README.md
@@ -85,11 +85,24 @@ The encoding shares the log archive test data with the macOS Unified Logging Rec
 In the ```./receiver/macosunifiedloggingreceiver/testdata``` folder off the root of the repository, run:
 
 ```bash
-curl -L -o logarchivetestdata.zip \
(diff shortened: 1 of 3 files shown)


Caleb-Hurshman:
Check on the gosec failures, I don't care too much about the lint.
It's just now occurring to me that pushing this without the test data will mean CI always fails. I'm sure there's a solution for this, but I'm not sure what it is.

Edit: nvm I see the skip `testutils` func, you just forgot to uncomment it in one place

extension/encoding/macosunifiedloggingencodingextension/timesync_test.go
```diff
@@ -15,11 +15,76 @@
...
+	"os"
+	"path/filepath"
 	"testing"
 
+	testutil "github.com/observiq/bindplane-otel-collector/extension/encoding/macosunifiedloggingencodingextension/internal/testutil"
 	"github.com/stretchr/testify/require"
 )
 
+func TestParseTimesyncData_DataFromArchive(t *testing.T) {
+	// testutil.SkipIfNoReceiverTestdata(t)
```
Caleb-Hurshman: Uncomment this to fix the unit test CI jobs

postnati: Fixed

PR #2793, opened 2025-11-06
[exporter/chronicleexporter] Cache log type field lookups

jsirianni:
### Proposed Change

This is a follow up for https://github.com/observIQ/bindplane-otel-collector/pull/1967. Previously, we short circuit and avoid the OTTL expression execution if possible.

This PR implements caching for calls to getRawField that do not satisfy the `switch field` implemented in PR 1967.

Implements the [Hashicorp LRU](https://github.com/hashicorp/golang-lru) to cache OTTL expression results. In general, a system will have one or a handful of log types. Performing a full ottl expression for each log entry is wasteful. Caching them allows the exporter to skip the ottl expression for repeat log types.

In testing, cache lookups were 160x faster than performing the ottl expression.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


exporter/chronicleexporter/marshal.go
```diff
@@ -355,13 +365,42 @@ func (m *protoMarshaler) getRawField(ctx context.Context, field string, logRecor
 		return "", nil
 	}
 
-	lrExpr, err := expr.NewOTTLLogRecordExpression(field, m.teleSettings)
-	if err != nil {
-		return "", fmt.Errorf("raw_log_field is invalid: %s", err)
+	// Check cache first
```
jsirianni, line 368: In some cases, the cache will be bypassed by the switch above ^ ^

jsirianni:
Not stale

PR #3255, opened 2026-03-06
feat: PIPE-857 switch systemd unit default user from root to bdot

Dylan-M:
### Proposed Change
Switches the systemd unit template from running as root to running as the `bdot` user/group. Adds `readUserFromSystemdFile()` to the updater to read the current user from an existing service file during upgrades.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


Dylan-M:
* **#3255** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-collector/3255?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#3256](https://github.com/observIQ/bindplane-otel-collector/pull/3256) <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-collector/3256?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3258](https://github.com/observIQ/bindplane-otel-collector/pull/3258) <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-collector/3258?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-collector/3255?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

scripts/package/postinstall.sh
```diff
@@ -496,7 +502,63 @@ finish_permissions() {
...
+  mkdir -p "$(dirname "$sudoers_file")"
+
+  cat << EOF > "$sudoers_file"
+# Sudoers drop-in for the Bindplane Distribution for OpenTelemetry Collector.
+# Grants the collector's runtime user permission to manage the collector
+# service and install updated service files. Generated at install time for
+# runtime user "${BDOT_USER}".
+
+# Service management via systemctl
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/systemctl start observiq-otel-collector
```
jsirianni: The updater will use service name `observiq-otel-collector.service` as it derives the name from the service file name. 

```suggestion
${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/systemctl start observiq-otel-collector.service
```

scripts/package/postinstall.sh
jsirianni: The updater needs to read this file unprivileged. 

```suggestion
  chown  root:${BDOT_GROUP}  "$config_file"
```

updater/internal/service/service_linux.go
```diff
@@ -102,34 +117,41 @@ func (l linuxSystemdService) Start() error {
...
 
-// installs the service
-func (l linuxSystemdService) install() error {
-	inFile, err := os.Open(l.newServiceFilePath)
+// installServiceFile copies the service file to its installed location.
+// When running as non-root, it uses "sudo install" to place the file
+// with root ownership. Otherwise, it copies the file directly.
+func installServiceFile(src, dst, mode string) error {
+	if needsSudo() {
+		cmd := sudoCommand("install", "-m", mode, "-o", "root", "-g", "root", src, dst)
```
jsirianni: I left a comment in postinstall.sh related to the group ownership of the service file (changing it from root to $BDOT_GROUP). It will also need to change here to avoid clobbering it back to root.

updater/internal/service/service_linux.go
```diff
@@ -102,34 +117,41 @@ func (l linuxSystemdService) Start() error {
...
 	defer func() {
-		err := inFile.Close()
-		if err != nil {
+		if err := inFile.Close(); err != nil {
 			log.Default().Printf("Service Install: Failed to close input file: %s", err)
 		}
 	}()
 
-	outFile, err := os.OpenFile(l.installedServiceFilePath, os.O_CREATE|os.O_WRONLY, 0600)
+	outFile, err := os.OpenFile(dst, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600) // #nosec G304 -- dst is an internal service file path, not user input
```
jsirianni: 0600 was correct before but now we need 0640 so the BDOT_GROUP can read the file.

```suggestion
	outFile, err := os.OpenFile(dst, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0640) // #nosec G304 -- dst is an internal service file path, not user input
```

scripts/package/postinstall.sh
```diff
@@ -496,7 +502,63 @@ finish_permissions() {
...
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/systemctl stop observiq-otel-collector
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/systemctl restart observiq-otel-collector
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/systemctl status observiq-otel-collector
```
jsirianni: These two commands are not used anywhere and should be removed.

scripts/package/postinstall.sh
jsirianni: I suspect this applies to `stop` as well. I would prefer we be consistent in the updater code, using `observiq-otel-collector` or `observiq-otel-collector.service`. One or the other.

jsirianni: `enable` and `disable` as well.

scripts/package/postinstall.sh
```diff
@@ -496,7 +502,63 @@ finish_permissions() {
...
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/chown root\:${BDOT_GROUP} /usr/lib/systemd/system/observiq-otel-collector.service
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/chmod 0640 /usr/lib/systemd/system/observiq-otel-collector.service
+
+# SysV service management
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/sbin/service observiq-otel-collector start
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/sbin/service observiq-otel-collector stop
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/sbin/service observiq-otel-collector restart
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/sbin/service observiq-otel-collector status
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/sbin/chkconfig observiq-otel-collector on
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/sbin/chkconfig observiq-otel-collector off
```
jsirianni: I think we agreed that sysv is out of scope. If it is, lets removes these sudoers entries as well.

updater/internal/service/service_linux.go
```diff
@@ -102,34 +117,41 @@ func (l linuxSystemdService) Start() error {
...
 	return nil
 }
 
-// installs the service
-func (l linuxSystemdService) install() error {
-	inFile, err := os.Open(l.newServiceFilePath)
+// installServiceFile copies the service file to its installed location.
+// When running as non-root, it uses "sudo install" to place the file
+// with root ownership. Otherwise, it copies the file directly.
+func installServiceFile(src, dst, mode string) error {
```
jsirianni: The `mode` param is only used when `needsSudo` is true. Its otherwise ignored.

It would be nice if the sudo and non sudo branches in this function used the same command to deploy the service file. The `install` command seems cleaner than the logic we have today in the non sudo branch.

scripts/package/postinstall.sh
```diff
@@ -496,7 +502,63 @@ finish_permissions() {
...
+        echo "WARNING: sudoers file $sudoers_file failed validation, removing"
+        rm -f "$sudoers_file"
+      fi
+    fi
+  fi
+}
+
 install
 install_service
+install_sudoers
```
jsirianni: We should wrap this function call in a condition so it is only installed when the user has configured an unprivileged installation.

```bash
if [ "${BDOT_UNPRIVILEGED}" = "true" ]; then
  install_sudoers
fi
```

scripts/package/postinstall.sh
```diff
@@ -102,7 +106,8 @@ User=root
 Group=${BDOT_GROUP}
 Environment=PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
 Environment=BINDPLANE_COLLECTOR_HOME=${BDOT_CONFIG_HOME}
-Environment=BINDPLANE_COLLECTOR_STORAGE=${BDOT_CONFIG_HOME}/storage
+Environment=BINDPLANE_COLLECTOR_STORAGE=${BDOT_STORAGE}
+Environment=BINDPLANE_COLLECTOR_LOGS=${BDOT_LOGS}
```
jsirianni: This variable is set but not used by the collectors logger. Lets remove the implementation of `BINDPLANE_COLLECTOR_LOGS` / `BDOT_LOGS` or follow it to completion.

Im in favor of removing it, introducing in a later PR. Will leave it up to you.

scripts/package/postinstall.sh
```diff
@@ -102,7 +106,8 @@ User=root
 Group=${BDOT_GROUP}
 Environment=PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
 Environment=BINDPLANE_COLLECTOR_HOME=${BDOT_CONFIG_HOME}
-Environment=BINDPLANE_COLLECTOR_STORAGE=${BDOT_CONFIG_HOME}/storage
+Environment=BINDPLANE_COLLECTOR_STORAGE=${BDOT_STORAGE}
```
jsirianni, line 108: Same as comment below. This does not seem to be used.

updater/internal/updater/templates.go
```diff
@@ -23,11 +23,12 @@ StartLimitIntervalSec=120
 StartLimitBurst=5
 [Service]
 Type=simple
-User=root
+User={{.User}}
```
jsirianni: This should always be `root` and not dynamic. When a non root user is configured, a systemd drop in is used.

Non root is supported during initial install only. https://docs.bindplane.com/configuration/bindplane-otel-collector/linux-package-configuration

We will explore moving to a non root by default approach in the future.

updater/internal/service/service_linux.go
```diff
@@ -29,6 +29,22 @@ import (
...
+func needsSudo() bool {
+	return os.Getuid() != 0
+}
+
+// sudoCommand creates an exec.Cmd, prepending "sudo" if the process is non-root.
+func sudoCommand(name string, args ...string) *exec.Cmd {
+	if needsSudo() {
+		allArgs := append([]string{name}, args...)
+		//#nosec G204 -- arguments are not user-controlled
+		return exec.Command("sudo", allArgs...)
```
jsirianni, line 43: I think we should use the non interactive flag here `-n`

updater/internal/service/service_linux.go
```diff
@@ -29,6 +29,22 @@ import (
 	"go.uber.org/zap"
 )
 
+// needsSudo returns true if the current process is not running as root.
+func needsSudo() bool {
+	return os.Getuid() != 0
+}
```
jsirianni, line 33: The `needsSudo` function is called several times. If the `sudo` command does not exist, the update could fail after it has started.

The updater should check if `needsSudo` as one of the first execution steps. If it returns true, the updater should check if the command exists. If it does not, it should abort the update before starting (before stopping service, etc).

scripts/package/postinstall.sh
```diff
@@ -496,7 +502,63 @@ finish_permissions() {
...
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/sbin/service observiq-otel-collector status
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/sbin/chkconfig observiq-otel-collector on
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/sbin/chkconfig observiq-otel-collector off
+EOF
+
+  chown root:root "$sudoers_file"
+  chmod 0440 "$sudoers_file"
+}
+
+validate_sudoers() {
```
jsirianni, line 532: If the sudoers file is invalid, the warning is logged but the install continues. Because the sudoers file should only be deployed when `BDOT_UNPRIVILEGED` is set, this function should cause the install to fail.

jsirianni (changes requested):

updater/internal/service/service_linux.go
```diff
@@ -29,6 +29,44 @@ import (
...
+// than failing partway through.
+func CheckSudoAvailable() error {
+	if !needsSudo() {
+		return nil
+	}
+
+	//#nosec G204 -- arguments are constant, not user-controlled
+	if err := exec.Command("sudo", "-n", "true").Run(); err != nil {
+		return fmt.Errorf("updater is running as a non-root user but cannot run sudo non-interactively (a NOPASSWD sudoers entry is required): %w", err)
+	}
```
jsirianni: `sudo -n true` is not covered by any grant in the sudoers config, so it's denied for bdot and every unprivileged update aborts when checking if sudo will be available on the system.

updater/internal/updater/updater.go
```diff
@@ -72,9 +76,28 @@ func NewUpdater(logger *zap.Logger, installDir string) (*Updater, error) {
 		monitor:                  monitor,
 		logger:                   logger,
 		installedSystemdUnitPath: DefaultSystemdUnitFilePath,
+		checkSudo:                service.CheckSudoAvailable,
 	}, nil
 }
 
+// readUserFromSystemdFile reads the systemd unit file and extracts the User value.
+func (u *Updater) readUserFromSystemdFile() (string, error) {
```
jsirianni: I think we won't need this function once `User={.User}` is removed. (See previous comment on templates.go line 26.

scripts/package/postinstall.sh
```diff
@@ -496,7 +499,63 @@ finish_permissions() {
...
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/systemctl restart observiq-otel-collector
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/systemctl status observiq-otel-collector
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/systemctl enable observiq-otel-collector
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/systemctl disable observiq-otel-collector
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/systemctl daemon-reload
+
+# Service file installation and ownership via install(1)/chown/chmod
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/install -m 0640 -o root -g root * /usr/lib/systemd/system/observiq-otel-collector.service
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/chown root\:${BDOT_GROUP} /usr/lib/systemd/system/observiq-otel-collector.service
+${BDOT_USER} ALL=(root) NOPASSWD: /usr/bin/chmod 0640 /usr/lib/systemd/system/observiq-otel-collector.service
```
jsirianni: Nothing invokes these commands. They should removed, re-added if that changes in the future.

jsirianni (changes requested):

scripts/package/postinstall.sh
Dylan-M: We can't set the ownership of the config file to `root:bdot`, because it would break config updates when the collector is running as non-root.

jsirianni: Im suggesting changing the group so the updater **can read** the unit file. 

The file is read unprivileged in the following places:
- readGroupFromSystemdFile — os.ReadFile at updater/internal/updater/updater.go:83, to fill the template's Group= field
- readEnvironmentFromSystemdFile — os.ReadFile at updater.go:103, to preserve BINDPLANE_COLLECTOR_STORAGE
- Backup() — CopyFileNoOverwrite of the installed unit at updater/internal/service/service_linux.go:205

If the ownership is root:root with `0640` permissions, the updater will fail immediately if the collector is running non root because it cannot read this file.

jsirianni (changes requested):

scripts/package/postinstall.sh
Dylan-M: I completely misread `$config_file` as the literal config file. I'm going to change that variable to `$unit_file` instead, because that is too confusing.

PR #3412, opened 2026-04-06
fix(linux): Skip GPG prompt on systems that cannot verify GPG signatures [skip ci]

jsirianni:
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/install/install_unix.sh
+++ b/scripts/install/install_unix.sh
@@ -67,7 +67,13 @@ offline_installation=false
 # rpm --import <revoked_public_key>.asc
 # rpm -q gpg-pubkey, it's one of these
 # rpm -q gpg-pubkey --info, go find the BDOT public key and use the version and release numbers from there.
-RPM_GPG_KEYS_TO_REMOVE=[]
+RPM_GPG_KEYS_TO_REMOVE=""
+
+# GPG_UNSUPPORTED_OS_LIST is a space-separated list of "id:major_version" entries
+# identifying OS versions that cannot perform GPG signature verification due to
+# incompatible RPM/GnuPG versions. Add new entries as "id:major_version".
+# Examples: "centos:6" "rhel:5"
+GPG_UNSUPPORTED_OS_LIST="centos:6 centos:7"
 
 # Colors
 if [ "$non_interactive" = "false" ]; then
@@ -687,7 +693,7 @@ os_arch_check()
 {
   info "Checking for valid operating system architecture..."
   arch=$(uname -m)
-  case "$arch" in 
+  case "$arch" in
     x86_64|aarch64|ppc64|ppc64le|arm64|aarch64_be|armv8b|armv8l|arm|armv6l|armv7l)
       succeeded
       ;;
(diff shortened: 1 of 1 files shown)


scripts/install/install_unix.sh
```diff
@@ -67,7 +67,13 @@ offline_installation=false
...
 # rpm -q gpg-pubkey, it's one of these
 # rpm -q gpg-pubkey --info, go find the BDOT public key and use the version and release numbers from there.
-RPM_GPG_KEYS_TO_REMOVE=[]
+RPM_GPG_KEYS_TO_REMOVE=""
+
+# GPG_UNSUPPORTED_OS_LIST is a space-separated list of "id:major_version" entries
+# identifying OS versions that cannot perform GPG signature verification due to
+# incompatible RPM/GnuPG versions. Add new entries as "id:major_version".
+# Examples: "centos:6" "rhel:5"
+GPG_UNSUPPORTED_OS_LIST="centos:6 centos:7"
```
Dylan-M, line 76: SLES 11 SP4 would have the same GPG verification issues (old RPM/GnuPG versions), but it's not in `GPG_UNSUPPORTED_OS_LIST`. Worse, `get_os_id()` only checks `/etc/os-release` and `/etc/redhat-release` — SLES 11 uses `/etc/SuSE-release` for identification, so it wouldn't be detected at all and GPG verification would proceed and likely fail.

scripts/install/install_unix.sh
```diff
@@ -1015,35 +1027,58 @@ verify_package_deb() {
   return 0
 }
 
+# get_os_id returns a "name:major_version" string for the current OS.
+# Reads /etc/os-release (modern systems) or /etc/redhat-release (legacy RHEL/CentOS).
+get_os_id() {
```
Dylan-M, line 1032: PR #3369 already introduces a more robust `detect_distro_package_type()` with a multi-layer fallback chain that handles `/etc/os-release`, `/etc/redhat-release`, `/etc/SuSE-release`, high-level package managers, and low-level tools. The `get_os_id()` here is a weaker re-implementation that misses several of those cases. This should build on or coordinate with that existing work rather than introducing a separate detection path.

scripts/install/install_unix.sh
Dylan-M, line 76: CentOS 7 ships RPM 4.11 and GnuPG 2.0, which should be capable of basic `rpm --checksig` verification. Since this PR also simplifies the RPM verification by removing the fragile subkey extraction pipeline and relying on `rpm --checksig` directly, CentOS 7 might actually work fine with the simplified path. Is it worth testing that before blocklisting it entirely?

scripts/install/install_unix.sh
```diff
@@ -917,6 +925,10 @@ install_package()
 }
 
 verify_package() {
+  if os_gpg_unsupported; then
```
Dylan-M, line 928: On blocklisted OSes, this silently skips all GPG verification — no download of the GPG keys, no warning logged, no user prompt. The package installs completely unverified with no indication to the operator that verification was skipped. We already have two mechanisms for users to consciously opt out of GPG verification: the interactive prompt when verification fails, and the `--no-gpg-check` CLI flag. Silently bypassing it based on OS detection removes the user's ability to make that security decision, and they won't even know it happened.

Dylan-M (changes requested):

PR #3456, opened 2026-04-23
chore: Undo rename of googlecloudstorage component

mrsillydog:
### Proposed Change

This PR undoes the rename of googlecloudstorage to google_cloud_storage. This change was incorrect, since the upstream exporter was renamed, but the upstream exporter was not actually utilized in our distribution.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


schmikei (approved):
Should we be renaming our component to match?

mrsillydog:
> Should we be renaming our component to match?

There was a bit of a discussion in code quality a few weeks back around this topic - how closely should we be attempting to adhere to upstream standards? I don't think we arrived at a solid answer around the topic, so I think for now, keeping our component name as-is is fine. If we make an effort to rename our components to match upstream standards, it should be done collectively for all components as part of a larger effort to standardize.

schmikei:
Yep the PR solves for right now so I'm good 👍

PR #3475, opened 2026-05-05
deps: OTel v0.151.0/v1.57.0 - BDOT-contrib v1.5.0

dpaasman00:
### Proposed Change

- Update OTel dependencies to [v0.151.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.151.0)
- Update BDOT contrib dependencies to v1.5.0
- Update available components script with new renames upstream

#### Breaking Changes / Deprecations
- A number of `stable` feature gates have been removed. The functionality they gated was previously set to default and there is no change in behavior with these changes. The only change is a runtime failure if still using these gates when running the collector. The list of gates is:
	- `connector.datadogconnector.NativeIngest`
	- `exporter.datadogexporter.UseLogsAgentExporter`
	- `exporter.datadogexporter.metricexportnativeclient` 
	- `pkg.translator.prometheus.NormalizeName`
	- `extension.jaegerremotesampling.replaceThriftWithProto`
	- `receiver.prometheusreceiver.EnableNativeHistograms`
	- `receiver.prometheusreceiver.RemoveStartTimeAdjustment`
	- `receiver.prometheusreceiver.UseCreatedMetric`
- `exporter/signalfx`: Default `api_url` and `ingest_url` values derived from `realm` now use `*.observability.splunkcloud.com` instead of `*.signalfx.com`. Explicit `api_url` and `ingest_url` settings are unchanged. Update network allowlists if they targeted only `*.signalfx.com`.
- `exporter/splunk_hec`: Remove deprecated `batcher` config field. Use `sending_queue::batch` instead.
- `pkg/zipkin`: Promote `pkg.translator.zipkin.DontEmitV0NetworkConventions` and `pkg.translator.zipkin.EmitV1NetworkConventions` feature gates to Beta. This changes the default behavior to emit the new semantic convention attributes instead of the old deprecated ones. The Zipkin translator will now use `network.local.address` (replacing `net.host.ip`), `network.peer.address` (replacing `net.peer.ip`), and `service.peer.name` (replacing `peer.service`) by default when emitting spans. Affects `receiver/zipkin` and `exporter/zipkin`.
- `processor/k8s_attributes`: Disable `otelcol.k8s.pod.association` metric until `pod_identifier` attribute is properly calculated.
- `receiver/jaeger`: Stabilize `DisableRemoteSampling` feature gate which has been in beta for over 2 years. This behavior was already enabled by default. With this stabilization, the behavior cannot be disabled.
- `exporter/prometheusremotewrite`: `add_metric_suffixes` is deprecated. Use `translation_strategy: UnderscoreEscapingWithoutSuffixes` if you are setting this to false. Otherwise, by default, this is set to `UnderscoreEscapingWithSuffixes`.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


go.mod
```diff
@@ -46,162 +46,162 @@ require (
 	github.com/observiq/bindplane-otel-contrib/receiver/windowseventtracereceiver v1.3.0
```
mrsillydog: These are still on 1.3.

go.mod
```diff
@@ -211,42 +211,42 @@ require (
 	github.com/honeycombio/enhance-indexing-s3-exporter/enhanceindexings3exporter v0.0.15
```
mrsillydog: This needs to be bumped to latest

mrsillydog (changes requested):

mrsillydog (approved):

PR #3482, opened 2026-05-06
fix: Windows updater path handling

jsirianni:
### Proposed Change

Replace os.Getwd() with os.Executable() + filepath.Dir() in the updater manager constructors so the updater binary and its staging directory are always resolved relative to the collector's install directory rather than its process working directory. This ensures the correct updater.exe is copied and executed regardless of what working directory the process was launched with, and sets cmd.Dir on the spawned updater process for the same guarantee. The TmpPath in managed mode is updated with the same idiom to keep staging consistent.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Dylan-M (approved):
On Linux/macOS (the change in `updater_manager_others.go`), `os.Executable()` may resolve symlinks — an install accessed via something like `/usr/bin/observiq-otel-collector` symlinked to `/opt/observiq-otel-collector/...` will resolve to the real install dir. Likely intended, but worth flagging.

joe-armstrong (approved):

joe-armstrong (approved):

Dylan-M (approved):
Non-blocking: the `TmpPath` anchoring in `managed.go` isn't tested. `NewManagedCollectorService` is mostly untested already. This should get addressed at some point, but I'm okay with merging without tests.

PR #3526, opened 2026-06-08
fix(v2): prevent Windows install scripts from killing the Powershell window on failure

mrsillydog:
## Summary

Both Windows install scripts could abruptly close the user's PowerShell window on failure, most visibly when running with `-Uninstall`. The BDOT script also failed to find the installed product to uninstall at all.

### Root causes

1. **Host termination.** `Fail()` called `exit` from inside a function. A bare `exit` terminates the entire PowerShell host process, closing the user's window before they can read the `[ERROR]` line. The uninstall path is especially prone to this because it fails fast.

2. **Stale product name (BDOT script only).** `scripts/install/install_windows.ps1` resolved the uninstall ProductCode by an exact match on `"observIQ Distro for OpenTelemetry Collector"`. The MSI now registers as `"BindPlane Distro for OpenTelemetry Collector (BDOT)"` (`windows/wix.json`), so the lookup always returned `$null` → `Fail` → window closed. Since `<Product Id="*">` regenerates the ProductCode every build, matching by name/version is fragile.

### Changes

- `Fail()` now `throw`s; a try/catch entry point prints the error and sets `$LASTEXITCODE` without killing the session. Applied to **both** scripts.
- BDOT script resolves the installed ProductCode via the **stable UpgradeCode** (`windows/wix.json`'s `upgrade-code`) using the Windows Installer COM API. This survives renames/version changes and avoids the slow, side-effecting `Win32_Product` WMI class.

### Notes

- The generic-install script intentionally uninstalls via `msiexec /x <downloaded.msi>` (it can't know an arbitrary distribution's UpgradeCode); that behavior is unchanged. Only the host-termination fix was applied there.
- "Product not installed" remains a hard error (not a silent success) so a misfire on a machine that genuinely has the product surfaces clearly.

## Test plan

- [x] Manually tested on Windows

### DIFF
--- a/scripts/install/install_windows.ps1
+++ b/scripts/install/install_windows.ps1
@@ -118,7 +118,9 @@ $ErrorActionPreference = "Stop"
 $DOWNLOAD_BASE = "https://bdot.bindplane.com"
 $MSI_NAME_AMD64 = "bindplane-otel-collector.msi"
 $MSI_NAME_ARM64 = "bindplane-otel-collector-arm64.msi"
-$PRODUCT_DISPLAY_NAME = "observIQ Distro for OpenTelemetry Collector"
+$PRODUCT_DISPLAY_NAME = "BindPlane Distro for OpenTelemetry Collector (BDOT)"
+# Stable across releases/renames; defined in windows/wix.json ("upgrade-code").
+$UPGRADE_CODE = "{D67CCA1A-6708-4096-8BDE-5069739FB861}"
 
 # ---- Helpers -----------------------------------------------------------------
 
@@ -133,9 +135,12 @@ function Write-Warn {
 }
 
 function Fail {
+    # Throw rather than `exit`. A bare `exit` from inside a function terminates the
+    # whole PowerShell host, which closes the user's window before they can read the
+    # error. The throw is caught at the script's entry point, which prints the error
+    # and sets a non-zero exit code without killing the session.
     param([string]$Message)
-    Write-Host "[ERROR] $Message" -ForegroundColor Red
-    exit 1
+    throw $Message
 }
 
 # ---- Privilege check ---------------------------------------------------------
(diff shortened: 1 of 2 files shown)


scripts/install/install_windows.ps1
```diff
@@ -241,12 +246,22 @@ function Build-MsiexecArgs {
...
+        $installer = New-Object -ComObject WindowsInstaller.Installer
+        $related = $installer.RelatedProducts($UPGRADE_CODE)
+    }
+    catch {
+        # RelatedProducts throws when no products match the upgrade code.
+        return $null
+    }
+
+    foreach ($code in $related) {
+        if ($code) { return $code }
```
mrsillydog, line 264: This is what triggered me to find this - searching for the wrong product on V2 uninstall caused my Powershell window to exit.

scripts/install/install_windows.ps1
```diff
@@ -377,4 +392,19 @@ function Main {
...
-Main
+# ---- Entry point -------------------------------------------------------------
+
+# Run Main inside a try/catch and avoid calling `exit` from script scope. A bare
+# `exit` terminates the host process, closing the user's PowerShell window before
+# they can read the error — especially on the uninstall path, which fails fast
+# when the product can't be found. Setting $LASTEXITCODE instead leaves the window
+# intact while still surfacing a non-zero code to non-interactive callers
+# (e.g. powershell.exe -File install_windows.ps1 -Uninstall).
+try {
```
Dylan-M, line 429: The comment says this surfaces a non-zero code to `powershell.exe -File` callers, but it doesn't. I tested on pwsh 7.6: setting `$LASTEXITCODE = 1` without an `exit` returns process exit code 0 under `-File`, so a scripted uninstall (`-File ... -Uninstall`) sees success even when Main threw. The window-safety fix is right for the inline `& ([scriptblock]::Create(...))` form the install command uses, where `exit` would kill the session and the code is irrelevant. But `-File` callers silently lose the failure.

One way to cover both (a Claude suggestion): guard the exit on how the script was invoked, e.g.
```powershell
if ($PSCommandPath) { exit $LASTEXITCODE }
```
`$PSCommandPath` is set under `-File` but empty for the inline scriptblock, so `-File` callers get the real code and inline callers keep their window. If `-File` exit codes aren't actually needed, the comment should drop that claim instead.

Same block in `generic-install/install_windows.ps1` needs the same treatment.

Dylan-M (changes requested):

Dylan-M (approved):

PR #3527, opened 2026-06-08
fix(v1): prevent Windows install scripts from killing the Powershell window on failure

mrsillydog:
## Summary

Port of #3526 to `main`. Both Windows install scripts could abruptly close the user's PowerShell window on failure, most visibly when running with `-Uninstall`.

### Root cause

`Fail()` called `exit` from inside a function. A bare `exit` terminates the entire PowerShell host process, closing the user's window before they can read the `[ERROR]` line. The uninstall path is especially prone to this because it fails fast.

### Changes

- `Fail()` now `throw`s; a try/catch entry point prints the error and sets `$LASTEXITCODE` without killing the session. Applied to **both** scripts.
- BDOT `install_windows.ps1` resolves the installed ProductCode via the **stable UpgradeCode** (`windows/wix.json`'s `upgrade-code`) using the Windows Installer COM API, instead of an exact match against `Win32_Product` by display name. This is forward-looking: when the distro is rebranded to BindPlane the display name changes, but the UpgradeCode is fixed, so uninstall keeps working across renames/versions. It also avoids the slow, side-effecting `Win32_Product` WMI class.

### main-specific notes

- Unlike the v2.0.0 release branch, the product name in `windows/wix.json` on `main` is still `observIQ Distro for OpenTelemetry Collector`, which currently matches the script — so the name-mismatch bug is not yet live here. The UpgradeCode switch is preventative (and a perf/correctness improvement) rather than a fix for a current break.
- The generic-install script intentionally uninstalls via `msiexec /x <downloaded.msi>` (it can't know an arbitrary distribution's UpgradeCode); that behavior is unchanged. Only the host-termination fix was applied there.

## Test plan

- [x] Manually tested on Windows (against the v2.0.0 release branch variant in #3526)

### DIFF
--- a/scripts/install/install_windows.ps1
+++ b/scripts/install/install_windows.ps1
@@ -119,6 +119,8 @@ $DOWNLOAD_BASE = "https://bdot.bindplane.com"
 $MSI_NAME_AMD64 = "observiq-otel-collector.msi"
 $MSI_NAME_ARM64 = "observiq-otel-collector-arm64.msi"
 $PRODUCT_DISPLAY_NAME = "observIQ Distro for OpenTelemetry Collector"
+# Stable across releases/renames; defined in windows/wix.json ("upgrade-code").
+$UPGRADE_CODE = "{D67CCA1A-6708-4096-8BDE-5069739FB861}"
 
 # ---- Helpers -----------------------------------------------------------------
 
@@ -133,9 +135,12 @@ function Write-Warn {
 }
 
 function Fail {
+    # Throw rather than `exit`. A bare `exit` from inside a function terminates the
+    # whole PowerShell host, which closes the user's window before they can read the
+    # error. The throw is caught at the script's entry point, which prints the error
+    # and sets a non-zero exit code without killing the session.
     param([string]$Message)
-    Write-Host "[ERROR] $Message" -ForegroundColor Red
-    exit 1
+    throw $Message
 }
 
 # ---- Privilege check ---------------------------------------------------------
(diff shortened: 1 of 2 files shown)


scripts/install/install_windows.ps1
```diff
@@ -377,4 +392,19 @@ function Main {
...
-Main
+# ---- Entry point -------------------------------------------------------------
+
+# Run Main inside a try/catch and avoid calling `exit` from script scope. A bare
+# `exit` terminates the host process, closing the user's PowerShell window before
+# they can read the error — especially on the uninstall path, which fails fast
+# when the product can't be found. Setting $LASTEXITCODE instead leaves the window
+# intact while still surfacing a non-zero code to non-interactive callers
+# (e.g. powershell.exe -File install_windows.ps1 -Uninstall).
+try {
```
Dylan-M, line 429: The comment says this surfaces a non-zero code to `powershell.exe -File` callers, but it doesn't. I tested on pwsh 7.6: setting `$LASTEXITCODE = 1` without an `exit` returns process exit code 0 under `-File`, so a scripted uninstall (`-File ... -Uninstall`) sees success even when Main threw. The window-safety fix is right for the inline `& ([scriptblock]::Create(...))` form the install command uses, where `exit` would kill the session and the code is irrelevant. But `-File` callers silently lose the failure.

One way to cover both (a Claude suggestion): guard the exit on how the script was invoked, e.g.
```powershell
if ($PSCommandPath) { exit $LASTEXITCODE }
```
`$PSCommandPath` is set under `-File` but empty for the inline scriptblock, so `-File` callers get the real code and inline callers keep their window. If `-File` exit codes aren't actually needed, the comment should drop that claim instead.

Same block in `generic-install/install_windows.ps1` needs the same treatment.

scripts/install/install_windows.ps1
```diff
@@ -241,12 +246,22 @@ function Build-MsiexecArgs {
...
-    if ($product) {
-        return $product.IdentifyingNumber
+    # Resolve the installed ProductCode via the stable UpgradeCode rather than an
+    # exact display-name match. The display name has changed across releases (e.g.
+    # "observIQ Distro..." -> "BindPlane Distro... (BDOT)"), but the UpgradeCode is
+    # fixed, so this keeps working across renames. The Windows Installer COM API
+    # also avoids the slow, side-effecting Win32_Product WMI class.
+    try {
+        $installer = New-Object -ComObject WindowsInstaller.Installer
+        $related = $installer.RelatedProducts($UPGRADE_CODE)
```
Dylan-M, line 256: The lookup is upgrade-code-based now, but `Write-Info`/`Fail` still print `'observIQ Distro for OpenTelemetry Collector'`. That upgrade code also matches a BDOT-renamed install, so the script can find and act on a product while telling the user it's looking for a different name. That mismatch is exactly the kind of thing that generates confused support tickets, so worth aligning the message with what's actually resolved, or making it name-agnostic.

Dylan-M (changes requested):

Dylan-M (approved):

PR #3532, opened 2026-06-12
chore: Changes for releasing from epic

dpaasman00:
### Proposed Change

Various changes for releasing from epic branch
- Add the `azureauthextension`. This was added to main but wasn't preserved by rebase because components are tracked differently.
- Add `embed_library` build tag. This was dded to main but wasn't preserved by rebase because binaries are built differently.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


manifests/observIQ/manifest.yaml
```diff
@@ -247,6 +248,4 @@ replaces:
   - github.com/mattn/go-ieproxy => github.com/mattn/go-ieproxy v0.0.1
   - github.com/cilium/ebpf => github.com/cilium/ebpf v0.11.0
   - github.com/DataDog/datadog-agent/comp/core/delegatedauth => github.com/DataDog/datadog-agent/comp/core/delegatedauth v0.78.2
-  - github.com/open-telemetry/opentelemetry-collector-contrib/receiver/windowseventlogreceiver => github.com/observiq/opentelemetry-collector-contrib/receiver/windowseventlogreceiver v0.0.0-20260422143330-2a6ebda5c14e
```
Dylan-M, line 250: This drops the WEL and stanza fork pins, so the epic build falls back to upstream v0.153.0. #3533 is taking main to v0.154.0 with these pins updated. Since the epic is meant to replace main, shouldn't it move to v0.154.0 and keep the fork pins, rather than drop them?

manifests/observIQ/manifest.yaml
```diff
@@ -34,7 +35,7 @@ extensions:
   # internal/extension/opampconnectionextension/, kept internal because the
   # whole runtime cluster (collector lifecycle, OpAMP client, packagestate,
   # report manager, etc.) ships in the same module. Local replace below.
-  - gomod: github.com/observiq/bindplane-otel-collector/internal/extension/opampconnectionextension v0.0.0
+  - gomod: github.com/observiq/bindplane-otel-collector/internal/extension/opampconnectionextension v1.101.1
```
Dylan-M, line 38: What's behind the v0.0.0 → v1.101.1 bump here and on snapshotprocessor? The local replace directives still cover them.

Dylan-M (changes requested):

manifests/observIQ/manifest.yaml
mrsillydog: Additionally, if we could take into account Dylan's comment from that PR and sync the stanza and WEL receiver replace syntax (observiq or observIQ for both), that would be great.

manifests/observIQ/manifest.yaml
dpaasman00: This is so the correct version is reported for available components

manifests/observIQ/manifest.yaml
dpaasman00: Yes, I can resolve this drift

Makefile
```diff
@@ -49,6 +51,11 @@ AGENT_LDFLAGS = -s -w \
 	-X github.com/observiq/bindplane-otel-contrib/pkg/version.gitHash=$(GIT_HASH) \
 	-X github.com/observiq/bindplane-otel-contrib/pkg/version.date=$(BUILD_DATE)
 
+# AGENT_BUILD_TAGS are the build tags that should be used when building BDOT
+# 'bindplane' builds with logic used by the v1 OpAMP implementation
+# 'embed_library' used by the telemetry generator receiver to use blitz (PR#3525)
+AGENT_BUILD_TAGS = bindplane embed_library
```
mrsillydog, line 57: Let's move these into the right spots in the manifests

Dylan-M (approved):

PR #3536, opened 2026-06-15
fix(ci): redirect Windows linker temp to D: via SystemTemp to fix disk exhaustion

Dylan-M:
## Problem

The Windows `unit-tests` leg fails deterministically with `link.exe: ... truncate C:\Windows\SystemTemp\go-link-*\go.o: There is not enough space on the disk` (PIPE-1112). Tests never run; every failure is `[build failed]` from the linker.

The previously-merged `TMP`/`TEMP`/`GOTMPDIR=D:\Temp` override (#3450) cannot work here. `-race` forces external linking, so `link.exe` allocates its scratch dir via Go's `os.TempDir()`, which on Server 2022 routes through the `GetTempPath2` Win32 API ([Go src](https://go.googlesource.com/go/+/refs/tags/go1.26.4/src/os/file_windows.go), since Go 1.21). These Namespace runners execute as **SYSTEM**, and [`GetTempPath2` is documented](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w) to ignore `TMP`/`TEMP` for SYSTEM processes and return `C:\Windows\SystemTemp`. The fact the linker resolved to (and wrote into) that SYSTEM-only path confirms the SYSTEM context. `C:` is under disk pressure, so the linker runs out of space.

## Fix

The same `GetTempPath2` docs state the one env var a SYSTEM process *does* honor is `SystemTemp`. So:

- Set `SystemTemp=D:\Temp` to redirect the linker's `go.o` onto the roomy `D:` workspace volume (the actual root-cause fix).
- Relocate `GOCACHE`/`GOMODCACHE` to `D:` as hygiene (read directly by the `go` tool, never via `GetTempPath2`). `GOPATH`/tools stay on `C:` to preserve `setup-go`'s PATH entry.
- Drop the no-op `TMP`/`TEMP`/`GOTMPDIR` env block; add a telemetry step (`go env`, `SystemTemp`, `Get-PSDrive` before/after, largest-`C:`-consumer report) so the run confirms the redirect landed.

Touches only the Windows matrix leg; Ubuntu/macOS unchanged. Fixes PIPE-1112.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -87,6 +87,50 @@ jobs:
           go-version-file: go.mod
           cache: false
 
+      # link.exe (external linking is forced by -race) writes its go-link temp via
+      # Go's os.TempDir(), which on Windows Server 2022 routes through GetTempPath2.
+      # These Namespace runners execute as SYSTEM, and for SYSTEM processes
+      # GetTempPath2 ignores TMP/TEMP and returns C:\Windows\SystemTemp; that volume
+      # is under disk pressure, which is the "not enough space" linker failure.
+      #
+      # Per the GetTempPath2 docs, the one env var a SYSTEM process honors is
+      # SystemTemp, so we set it to redirect the linker's go.o onto the roomy D:
+      # volume (where the workspace already lives). We also relocate the Go build and
+      # module caches to D: as hygiene (GOCACHE/GOMODCACHE are read directly by the go
+      # tool, never via GetTempPath2). GOPATH stays on C: so setup-go's PATH entry for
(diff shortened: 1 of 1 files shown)


jsirianni (approved):

Dylan-M:
### Merge activity

* **Jun 15, 4:33 PM UTC**: @Dylan-M merged this pull request with [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-collector/3536).

PR #3554, opened 2026-06-23
chore: v1.102.1 patch

dpaasman00:
### Proposed Change

- Bump module versions v1.102.1
- Add disable feature gate `exporter.prometheusexporter.DisableAddMetricSuffixes` to prevent breaking change
  - Needed to add parsing logic for `-` and `+` to indicate enabling and disabling. This is the same parsing used for user set gates.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


internal/extension/opampconnectionextension/internal/collector/featuregates.go
mrsillydog, line 1: Can we just collapse setHardcodedFeatureFlags and setUserFeatureFlags into one function? Looks like the only difference is now that one takes a param.

mrsillydog (approved):

PR #3575, opened 2026-07-16
chore: Remove migrated modules, leave the README

mrsillydog:
### Proposed Change

This PR removes all the modules that have been migrated to bindplane-otel-contrib, leaving their READMEs as a link to where the modules now live. It also removes the infrastructure that kept these modules frozen in time.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dpaasman00 (changes requested):
I think we still need the `internal/report/snapshot' component

dpaasman00 (approved):
Misread

PR #3582, opened 2026-07-23
chore(linux): Set systemd unit file ownership to root:bdot

jsirianni:
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/scripts/package/postinstall.sh
+++ b/scripts/package/postinstall.sh
@@ -116,7 +116,9 @@ KillMode=process
 WantedBy=multi-user.target
 EOF
 
-  chown root:root "$config_file"
+  # Own the unit file by root:<runtime group> so the collector's runtime user
+  # (a member of BDOT_GROUP) can read it. Mode stays 0640 (group-readable only).
+  chown "root:${BDOT_GROUP}" "$config_file"
   chmod 0640 "$config_file"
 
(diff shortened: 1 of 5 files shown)


scripts/package/postinstall.sh
```diff
@@ -116,7 +116,9 @@ KillMode=process
 WantedBy=multi-user.target
 EOF
 
-  chown root:root "$config_file"
+  # Own the unit file by root:<runtime group> so the collector's runtime user
+  # (a member of BDOT_GROUP) can read it. Mode stays 0640 (group-readable only).
+  chown "root:${BDOT_GROUP}" "$config_file"
```
jsirianni, line 121: `config_file` is declared a the start of this function and is the systemd unit file.

```bash
config_file="/usr/lib/systemd/system/observiq-otel-collector.service"
```

updater/internal/service/testdata/linux-service.service
```diff
@@ -4,6 +4,7 @@ After=network.target
 [Service]
 Type=simple
 User=root
+Group=root
```
jsirianni, line 7: The test suite determine which user to use for the `chown` command based on this field. The CI runners will not have a `bdot` user, so we can use `root` here.

PR #3587, opened 2026-07-27
fix(bug): Treat blank manager.yaml as absent and recreate it from env vars

eKuG:
A manager.yaml that exists but is empty (or whitespace-only) previously entered managed mode with no endpoint, which crash looped the collector on every start. A blank file carries no user intent, so it now takes the same path as a missing file: bootstrapped from the OPAMP_* environment variables when OPAMP_ENDPOINT is set, standalone mode otherwise. Non-blank files keep the existing behavior, including failing closed on malformed content.

### Proposed Change

A blank file carries no user intent, so bootstrapManagerConfig now treats it exactly like a missing one: recreated from OPAMP_* env vars when OPAMP_ENDPOINT is set, standalone mode otherwise. Non-blank files are untouched, malformed content still fails closed, nothing user-written is ever overwritten.

Tested: new blank-file unit tests; E2E against a live Bindplane server empty manager.yaml + env now connects, takes rollouts, and keeps its agent_id across restarts; unpatched image still crash loops.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


Dylan-M (changes requested):
Two test gaps:

- Only empty + no-env is asserted; could you also cover whitespace-only + no-env?
- The `return readErr` branch is unexercised. Can we get a case where the read fails for something other than `os.ErrNotExist`?

Dylan-M (approved):

PR #3594, opened 2026-07-30
fix: Keep GOMAXPROCS at node core count under container CPU limits

eKuG:
### Proposed Change

Go 1.25 made GOMAXPROCS container-aware: under a cgroup CPU limit it defaults to `max(2, ceil(limit))` instead of the node's logical core count ([release notes](https://go.dev/doc/go1.25), [blog](https://go.dev/blog/container-aware-gomaxprocs), [golang/go#73193](https://github.com/golang/go/issues/73193)). BDOT picked this up when the toolchain moved from Go 1.24 to 1.25 in v1.94.0, an unreviewed runtime-behavior change for every CPU-limited container.

On GKE Autopilot every pod carries a CFS CPU limit, so collectors there silently lost most of their parallelism. Observed impact on a large Pub/Sub ingestion fleet upgrading v1.92.2 → v1.96.0/v1.100.0: throughput regression, sustained expired ack deadlines (300–999/s), redelivery storms, and an unbounded backlog (275M undelivered) — restored on rollback to v1.92.2 (clean A/B/A). Verified on shipped images under `--cpus=2` (via `GODEBUG=schedtrace`): `bindplane-agent:1.92.2` → `gomaxprocs=14` (node cores), `1.100.0` through `1.104.0` → `gomaxprocs=2`.

**Both directions stay available per deployment, no rebuild:**
- `GOMAXPROCS=<n>` env — pin an explicit value (always wins).
- `GODEBUG=containermaxprocs=1` env opt the container-aware clamp back in. `updatemaxprocs` is deliberately left at its default so this opt-in keeps tracking limit changes (Kubernetes in-place pod resize); pinning it would freeze GOMAXPROCS at startup for those users while buying nothing for the default case (with the cgroup consult disabled, the updater only tracks CPU affinity).
- The `cgroup_runtime` extension (already compiled into BDOT, wraps uber-go/automaxprocs) remains the config-driven opt-in for limit-aware GOMAXPROCS + GOMEMLIMIT.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,13 @@ verify-manifest:
 	cd $(BUILD_DIR) && go mod tidy
 	# Compile with the same build tags as the release build ($(AGENT_BUILD_TAGS))
 	# so the gate exercises exactly what `make agent` ships.
-	cd $(BUILD_DIR) && CGO_ENABLED=0 go build -tags "$(AGENT_BUILD_TAGS)" -o /dev/null .
+	cd $(BUILD_DIR) && CGO_ENABLED=0 go build -tags "$(AGENT_BUILD_TAGS)" -o verify-binary .
+	# The GOMAXPROCS runtime default is carried by a //go:debug directive in
+	# $(AGENT_MAIN); it only takes effect if that file survives into the ocb
+	# main package. Assert it made it into the built binary.
+	cd $(BUILD_DIR) && go version -m verify-binary | grep -q 'containermaxprocs=0' || \
+		(echo "built binary is missing DefaultGODEBUG containermaxprocs=0 (lost $(AGENT_MAIN)?)"; exit 1)
+	rm -f $(BUILD_DIR)/verify-binary
 
 # Builds the collector for the current GOOS/GOARCH pair using ocb.
 .PHONY: agent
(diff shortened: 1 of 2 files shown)


jsirianni (changes requested):
I am hesitant to make this change for all users of the collector. Would removing the CPU limit of the deployed containers help? I know GKE autopilot requires limits by default but I believe there is a way around that. https://docs.cloud.google.com/kubernetes-engine/docs/how-to/pod-bursting-gke

If they need limits and want more control over max procs, I would suggest setting `GOMAXPROCS` on the container env.

eKuG:
@jsirianni Scope is narrower than all users: the directive is a no-op without a cgroup CPU limit, VMs, bare metal, and unlimited pods see zero change. For limited pods, this restores the behavior every release ≤ v1.93.x had; the default already flipped for those users in v1.94.0 via the Go 1.25 bump, unreviewed.
Removing limits/bursting works but pushes an infra change onto each affected customer and only covers GKE, ECS, vanilla K8s, docker --cpus are equally affected. GOMAXPROCS env is exactly our mitigation for already-shipped versions; going forward the override cost is symmetric: with this PR, anyone wanting the clamp sets GODEBUG=containermaxprocs=1 on the pod. 
What do you think?

PR #3600, opened 2026-08-05
feat: Container initialization cmd

dpaasman00:
### Proposed Change

Built a new binary that gets bundled into container images. It can be used to initialize the container env for the collector.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


.github/dependabot.yml
```diff
@@ -1,5 +1,13 @@
 version: 2
 updates:
+  - package-ecosystem: "gomod"
+    directory: "/cmd/container-init"
+    schedule:
+      interval: "weekly"
+    ignore:
+      # Opentelemetry updates will be done manually
+      - dependency-name: "github.com/open-telemetry/opentelemetry-collector*"
+      - dependency-name: "go.opentelemetry.io/*"
```
jsirianni: I think we can omit this as the binary does not include third party dependencies.

cmd/container-init/main.go
```diff
@@ -0,0 +1,98 @@
...
+      level: none
+`
+	defaultLoggingConfig = `output: stdout
+level: info
+`
+)
+
+func main() {
+	configPath := flag.String("config", "", "path to write the default collector config (required)")
+	loggingPath := flag.String("logging", "", "path to write the default logging config (required)")
```
jsirianni: I think we should take absolute file paths, and include validation that checks the path with `filepath.IsAbs`. This way we do not need to consider the working directory.

```suggestion
	configPath := flag.String("config", "", "absolute path to write the default collector config (required)")
	loggingPath := flag.String("logging", "", "absolute path to write the default logging config (required)")
```

cmd/container-init/main.go
```diff
@@ -0,0 +1,98 @@
...
+
+	if err := run(*configPath, *loggingPath, *overwrite); err != nil {
+		log.Fatalf("Failed to initialize container: %v", err)
+	}
+}
+
+// run creates the parent directories of configPath and loggingPath and
+// writes default file contents to them. Existing files are left untouched
+// unless overwrite is true.
+func run(configPath, loggingPath string, overwrite bool) error {
```
jsirianni, line 75: I think it would be nice to include simple and brief log lines for each operation (dir create, file create, etc). This way we can debug a deployment by reviewing the initContainer's logs.

jsirianni (changes requested):

cmd/container-init/main.go
```diff
@@ -0,0 +1,108 @@
...
+			return fmt.Errorf("path %s must be absolute", f.path)
+		}
+	}
+
+	for _, f := range files {
+		dir := filepath.Dir(f.path)
+		if err := os.MkdirAll(dir, 0750); err != nil {
+			return fmt.Errorf("create directory %s: %w", dir, err)
+		}
+		log.Printf("created directory %s", dir)
```
jsirianni: `log.Printf("created directory %s", dir)` fires even when MkdirAll found the
directory already present, which is the common case for a mounted volume.

jsirianni (changes requested):

jsirianni (approved):

PR #3602, opened 2026-08-06
feat(opamp): Check for `RetryAfterNanoseconds` in error response

antonblock:
### Proposed Change

Update the OpAMP client to look for `RetryInfo.RetryAfterNanoseconds` in `onErrorHandler`. When it receives an error response with that field, stop sending messages for the duration of that period.

- Add a new `sendGate` struct that can be set to block messages for a duration, and wait until that duration has passed
- Add `gatedOpAMPClient` which wraps the current OpAMP client, and use `sendGate` to wait until sending isn't blocked before sending messages

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/internal/extension/opampconnectionextension/internal/opamp/observiq/gated_opamp_client.go
+++ b/internal/extension/opampconnectionextension/internal/opamp/observiq/gated_opamp_client.go
@@ -1,0 +1,92 @@
+// Copyright observIQ, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package observiq
+
+import (
+	"context"
+
+	"github.com/open-telemetry/opamp-go/client"
+	"github.com/open-telemetry/opamp-go/protobufs"
+)
+
+// gatedOpAMPClient wraps a client.OpAMPClient so that every call which
+// results in a message being sent to the Server first waits on a sendGate.
+// Start, Stop, and AgentDescription are passed straight through since they
+// don't send new outgoing data (or, in the case of Stop, must never be
+// delayed).
+type gatedOpAMPClient struct {
+	client.OpAMPClient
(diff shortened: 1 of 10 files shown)


internal/extension/opampconnectionextension/internal/opamp/observiq/observiq_client.go
```diff
@@ -438,6 +445,12 @@ func (c *Client) onConnectFailedHandler(_ context.Context, err error) {
 
 func (c *Client) onErrorHandler(_ context.Context, errResp *protobufs.ServerErrorResponse) {
 	c.logger.Error("Server returned an error response", zap.String("Error", errResp.GetErrorMessage()))
+
+	if retryInfo := errResp.GetRetryInfo(); retryInfo != nil {
+		wait := time.Duration(retryInfo.GetRetryAfterNanoseconds())
```
jsirianni: This is a uint64 to int64 conversion. I think we should protect against overflow here, even if it is not likely.

internal/extension/opampconnectionextension/internal/opamp/observiq/observiq_client.go
```diff
@@ -438,6 +445,12 @@ func (c *Client) onConnectFailedHandler(_ context.Context, err error) {
...
+		c.logger.Info("Server requested a retry delay, blocking outgoing messages", zap.Stringer("wait", wait))
+		c.sendGate.block(wait)
```
jsirianni: The other call to c.sendGate checks if it is nil but we dont check here. I think its unlikely it would be nil but we should be consistent.

jsirianni (changes requested):

internal/extension/opampconnectionextension/internal/opamp/observiq/gated_opamp_client.go
```diff
@@ -0,0 +1,113 @@
...
+	g.gate.wait(nil)
+	g.OpAMPClient.SetFlags(flags)
+}
+
+// SendCustomMessage calls [client.OpAMPClient.SendCustomMessage] with its OpAMPClient after making sure
+// the sendGate isn't blocking
+func (g *gatedOpAMPClient) SendCustomMessage(message *protobufs.CustomMessage) (chan struct{}, error) {
+	g.gate.wait(nil)
+	return g.OpAMPClient.SendCustomMessage(message)
+}
```
dpaasman00, line 99: Highlighting that with recent changes to the throughput and topology processors that the changes in this PR to `measurements.go` and `topology.go` will be outdated.

However custom messages for those types will still get routed through this behavior because they will invoke this function.

jsirianni (approved):

dpaasman00 (approved):

PR #3648, opened 2026-09-16
fix(snapshotprocessor): hand batches to the snapshot buffer without copying them (BPOP-5844)

eKuG:
### Proposed Change
Depends on observIQ/bindplane-otel-contrib#969 (`pkg/snapshot`). Until that ships, the three go.mod files point at the branch pseudo-version; bump to `pkg/snapshot/v1.15.0` once it is tagged.

The snapshot processor deep-copied every batch before the buffer kept at most 100 records, and Bindplane wires three of these processors into every pipeline. With the new `pkg/snapshot`, which copies only what it retains and admits about 100 items per second once full, the processor and the report manager pass the pipeline payload straight to `Add`; it is never retained or mutated. This also caches the processor's component ID string, which keyed the reporter's buffer map on every batch and allocated each time, and uses the shared pooled gzip writer for snapshot reports instead of two local copies.

Measured on the stock 1.107.0 image against this build with the same load (three snapshot processors, ~13k log records/s, 90 s): allocation overhead of the snapshot processors fell from +25% to +2.6% at 1-record batches and from +20% to +0.1% at 1000-record batches; `processLogs` went from 9.4% to 1.1% of CPU samples, and the copy no longer appears in the CPU or heap profile. `make verify-manifest` passes. Verified end to end against a local Bindplane with this build rolled out: source, destination and agent "View Recent Telemetry" snapshots return records; the agent stays connected.

#### Update: on-demand collection

The contrib branch now also switches a buffer to on-demand mode on pipelines that sustain more than ten buffers' worth of records per second: nothing is buffered between requests and a snapshot request collects its records just in time. No collector code change; this PR only tracks the new contrib head. Docker A/B with three snapshot processors at ~13k log records/s: snapshot path 0.21% of CPU samples at 1-record batches, absent at 1000-record batches; allocation overhead +2.6% / +0.0%. Browser E2E with this build on two pipelines (2,006 and 20 records/s) passed; details in the contrib PR and `pkg/snapshot/PERFORMANCE.md`.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


eKuG:
* **#3648** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-collector/3648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-collector/3648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

PR #3653, opened 2026-09-21
feat(container-init): add -chown to hand a hostPath volume to the collector user

jsirianni:
### Proposed Change

Adds --chown, --uid, --gid flags for setting permissions on a path.

This will allow us to update our Daemonset collector's init container to set permissions, which allows us to reduce the permission level of the final collector container.

### Testing

In testing, I deployed Bindplane collector's from the latest Bindplane release (running as uid 0). I then upgraded to this branch's build and re-configured my Daemonset to run as non root.

The init container logged the following on startup (upgrade):

```
2026/09/22 16:39:47 running as Uid=0,0,0,0 Gid=0,0,0,0 Groups=0,1,2,3,4,6,10,11,20,26,27,1000000000 CapEff=0000000000000005
2026/09/22 16:39:47 skipped /etc/otel/storage/config.yaml: file already exists
2026/09/22 16:39:47 skipped /etc/otel/storage/logging.yaml: file already exists
2026/09/22 16:39:47 chowned /etc/otel/storage to 1000000000:1000000000
```

##### Checklist
- [x] Changes are tested
- [x] CI has passed


cmd/container-init/main.go
```diff
@@ -57,16 +65,36 @@ func main() {
...
 	if err := run(*configPath, *loggingPath, *overwrite); err != nil {
 		log.Fatalf("Failed to initialize container: %v", err)
 	}
+	// After run, so the files it wrote are handed over as well.
+	if target.path != "" {
+		if err := chownTree(target.path, target.uid, target.gid, os.Lchown); err != nil {
+			log.Fatalf("Failed to chown %s: %v", target.path, err)
+		}
+		log.Printf("chowned %s to %d:%d", target.path, target.uid, target.gid)
+	}
```
dpaasman00, line 97: Should there be a check to verify the logging and config files exist under the chownPath?

dpaasman00 (approved):
One thought, otherwise this looks good to me

cmd/container-init/main.go
jsirianni: I don't think so. The config / logging flags will fail if the files don't exist / cant be created.

PR #3655, opened 2026-09-21
feat: redact secrets from support bundles

Dylan-M:
### Proposed Change

Redact secrets in files collected for the bundle. Both the bash and PowerShell scripts stage a copy of each file, redact the copy, and bundle that, so the originals on disk are never touched.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


eKuG (commented):
Tested the redaction against realistic Bindplane-rendered configs (GNU sed 4.9 in a Debian container, since this script is Linux-only). Big win that manager.yaml is now collected and its secret_key redacted, and that originals are genuinely untouched. Four things found, two of them real credentials surviving redaction.

scripts/support/create-support-bundle.sh
```diff
@@ -215,6 +215,24 @@ check_prereqs() {
...
 
+# Redact secrets from a staged bundle file in place (originals untouched).
+# Pass 1: YAML "<sensitive-key>: value" -> [REDACTED]. Pass 2: secret-shaped
+# values anywhere (URL creds, Bearer, AWS key, JWT, PEM), which covers logs.
+# Line-based, so multi-line YAML block scalars are not covered except PEM.
+redact_in_place() {
+  f="$1"
+  [ -f "$f" ] || return 0
+  sed -E -i \
+    -e 's/^([[:space:]]*[-]?[[:space:]]*[A-Za-z0-9_.-]*(password|passwd|secret|token|api[_-]?key|access[_-]?key|private[_-]?key|encryption[_-]?key|credential|passphrase|authorization|bearer|connection[_-]?string|account[_-]?key)[A-Za-z0-9_.-]*[[:space:]]*:[[:space:]]*).*$/\1"[REDACTED]"/I' \
```
eKuG: Two shipped destinations leak their API key through this alternation, verified by running this exact sed against rendered config shapes:

1. **Datadog** — `datadog.yaml:210` renders `key: {{ .api_key }}` (nested under `api:`), and `datadog_agent.yaml` does the same in 3 places. A bare `key:` matches none of `api[_-]?key|access[_-]?key|private[_-]?key|encryption[_-]?key|account[_-]?key`, so `key: <dd api key>` passes through untouched.
2. **Honeycomb** — `honeycomb.yaml` and `honeycomb_refinery_otlp.yaml` (3x) render `x-honeycomb-team: {{ .api_key | quote }}`. No keyword matches, so the key survives.

Also worth adding `creds` (chronicle.yaml uses `creds:`, not `credentials:`).

For a bare `key:` the tradeoff is over-redacting `key_file:` / `keytab_file:`, which only costs a debugging path — safe direction for this tool. Same alternation is in the .ps1, so both need it.

Not a leak, for the record: New Relic's `license_key` renders as the header `api-key:`, which this already matches.

scripts/support/create-support-bundle.sh
```diff
@@ -215,6 +215,24 @@ check_prereqs() {
...
+    -e 's#([A-Za-z][A-Za-z0-9+.-]*://[^:/@[:space:]]+):[^@/[:space:]]+@#\1:[REDACTED]@#g' \
+    -e 's/([Bb]earer[[:space:]]+)[A-Za-z0-9._~+/=-]+/\1[REDACTED]/g' \
+    -e 's/AKIA[0-9A-Z]{16}/[REDACTED-AWS-ACCESS-KEY]/g' \
+    -e 's/eyJ[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+/[REDACTED-JWT]/g' \
+    "$f"
+  sed -E -i '/-----BEGIN[A-Z ]*PRIVATE KEY-----/,/-----END[A-Z ]*PRIVATE KEY-----/c\
+[REDACTED-PRIVATE-KEY]' "$f"
```
eKuG: This range replacement truncates Chronicle configs to EOF.

chronicle.yaml renders `creds: |-` holding the GCP service-account JSON, where `-----BEGIN PRIVATE KEY-----` and `-----END PRIVATE KEY-----` are on the **same line** (escaped `\n`). sed's `/BEGIN/,/END/` looks for the end address on a *later* line, never finds one, and so `c\` replaces everything from that line to end-of-file.

Measured on a chronicle-shaped config: 11 lines became 3, and the entire `service:` block was gone. The secret is removed, but the most useful part of the bundle disappears silently.

The .ps1 already gets this right with a lazy match (`(?s)-----BEGIN[A-Z ]*PRIVATE KEY-----.*?-----END[A-Z ]*PRIVATE KEY-----`). Porting that shape here — a non-greedy substitution rather than a line range — fixes it and makes the two scripts agree.

Minor, same rule: the replacement lands at column 0, so the redacted copy is no longer valid YAML.

scripts/support/create-support-bundle.sh
```diff
@@ -282,19 +312,35 @@ function bundle_files() {
...
         # shellcheck disable=SC2162
         read -p "Do you want to include the collector config (y or n)? " response
         if [ "$response" != "n" ]; then
-            info "Adding collector config $(fg_cyan "$collector_config")$(reset)"
-            tar --append --file="$tar_filename" -C "$collector_dir" config.yaml
+            # Stage a copy, redact it, then bundle the copy so the on-disk
+            # originals are never modified.
+            if [ -f "$collector_config" ]; then
+                info "Adding collector config (redacted) $(fg_cyan "$collector_config")$(reset)"
+                cp "$collector_config" config.yaml
```
eKuG: These stage into the current working directory and then `rm -f` them, unlike the logs above which correctly use an `sb_logs_$$` staging dir.

So running the script from any directory that already contains a `config.yaml` (or `manager.yaml`) silently overwrites that file and then deletes it. Running from `/opt/observiq-otel-collector` itself is saved only by `cp` refusing a same-file copy plus `set -e` — otherwise `redact_in_place` would rewrite the live config and `rm -f` would remove it.

Staging these two into the existing `sb_logs_$$`-style temp dir (or an `mktemp -d`) removes the whole class of problem.

scripts/support/create-support-bundle.sh
```diff
@@ -215,6 +215,24 @@ check_prereqs() {
   decrease_indent
 }
 
+# Redact secrets from a staged bundle file in place (originals untouched).
+# Pass 1: YAML "<sensitive-key>: value" -> [REDACTED]. Pass 2: secret-shaped
+# values anywhere (URL creds, Bearer, AWS key, JWT, PEM), which covers logs.
+# Line-based, so multi-line YAML block scalars are not covered except PEM.
+redact_in_place() {
```
eKuG, line 221: Pass 1 is `^`-anchored, so a `key: value` pair appearing mid-line in log text isn't matched — e.g. `debug config dump: password: plaintext` survives. Worth knowing given the comment says pass 2 "covers logs"; pass 2 only catches secret-*shaped* values (URL creds, Bearer, AKIA, JWT), not arbitrary `password:` inside a message.

Separately: the new dot-source guard is for tests, but sourcing this script in a bare/CI environment exits before defining anything — top-level `num_colors=$(tput colors 2>/dev/null)` fails when `TERM` is unset and `set -e` (line 16) kills the shell. That's how it behaved when I tried to source it in a container. Worth confirming the intended test harness can actually load it.

scripts/support/create-support-bundle.sh
```diff
@@ -215,6 +215,61 @@ check_prereqs() {
...
+            }
+          }
+          if (!done) { line = prefix "[REDACTED-PRIVATE-KEY]"; break }
+        }
+      }
+      print line
+    }
+  ' "$f" > "$f.redact.$$" && mv "$f.redact.$$" "$f"
+  sed -E -i \
+    -e 's/^([[:space:]]*[-]?[[:space:]]*[A-Za-z0-9_.-]*(password|passwd|secret|token|key|creds|honeycomb|api[_-]?key|access[_-]?key|private[_-]?key|encryption[_-]?key|credential|passphrase|authorization|bearer|connection[_-]?string|account[_-]?key)[A-Za-z0-9_.-]*[[:space:]]*:[[:space:]]*).*$/\1"[REDACTED]"/I' \
```
eKuG: Widening the alternation broke mapping keys that merely *contain* a sensitive word but have no value — the block openers. Because the pattern replaces `.*$` (empty) with `"[REDACTED]"`, the key line itself gets a bogus value appended:

```
  - bearertokenauth:"[REDACTED]"
        token: "[REDACTED]"
secrets:"[REDACTED]"
  provider: file
```

`bearertokenauth` is not hypothetical — the Dynatrace v2 and Splunk HEC destinations both render `- bearertokenauth:` as an extension block. Result is invalid YAML (`python3 -c 'yaml.safe_load(...)'` → "mapping values are not allowed here") and it reads as though the component *name* was the secret.

Fix is to require a non-empty value before substituting, e.g. end the match with `[^[:space:]].*$` instead of `.*$`, so `foo:` with nothing after it is left alone. Block scalars are already handled by the awk pass above, so nothing is lost. Same change needed in the `.ps1` (`$keyRe`).

Second, smaller effect of the bare `key` token: `allow_all_keys: false` on the redaction processor becomes `"[REDACTED]"`, and `key_file:` paths go too. Over-redaction is the safe direction and I'd still take it over the Datadog leak, but excluding `*_file`/`*_keys`/`allow_*` would keep some genuinely useful debugging detail.

scripts/support/create-support-bundle.sh
```diff
@@ -215,6 +215,61 @@ check_prereqs() {
...
+          e = b + RSTART + RLENGTH - 1
+          line = substr(line, 1, b - 1) "[REDACTED-PRIVATE-KEY]" substr(line, e + 1)
+        } else {
+          prefix = substr(line, 1, b - 1)
+          done = 0
+          while ((getline nl) > 0) {
+            if (nl ~ /-----END[A-Z ]*PRIVATE KEY-----/) {
+              sub(/^.*-----END[A-Z ]*PRIVATE KEY-----/, "", nl)
+              line = prefix "[REDACTED-PRIVATE-KEY]" nl; done = 1; break
+            }
```
eKuG, line 255: `awk ... > "$f.redact.$$" && mv ...` is fail-open: if awk exits non-zero the `mv` is skipped and the **unredacted** staged copy is what gets tarred. `set -e` does not rescue this — a failing left side of an `&&` list doesn't trigger errexit (verified: `set -e; false && echo hi; echo survived` prints `survived`).

Unlikely to fire now that `awk` is in `PREREQS`, but the failure mode is silent secret leakage, which is exactly what this function exists to prevent. Worth making it explicit — `|| { error "redaction failed for $f"; exit 1; }` — so a redaction failure aborts the bundle instead of shipping the original.

Also leaves `$f.redact.$$` behind on failure.

eKuG (commented):
Re-tested all four items against this head with GNU sed/awk in a container. All fixed and verified: Datadog `key:`, Honeycomb `x-honeycomb-team:`, Chronicle no longer truncates to EOF (structure survives — `service:`, `log_type`, pipelines all intact), block-scalar bodies now covered, mid-line log `password:` caught, config/manager/journalctl staged in `sb_files_$$`, and the `tput || true` fix means the script can actually be sourced. Nice work on the awk rewrite — it covers more than I asked.

One regression from widening the alternation, plus one fail-open nit.

eKuG (approved):

